PR #4, opened 2023-01-16
Add licence to frontend

diversemix:


joel-wright:
I think you meant "frontend"... FTFY

joel-wright (approved):

PR #6, opened 2023-02-08
Map visualisation

PatFawbertMills:
The geo json file is 6mb / 150k lines, which is something I'm going to come back to and review at a later date.


src/hooks/useUpdateCountries.ts
```diff
@@ -1,23 +1,15 @@
...
-    {
-      onError: (err) => {
-        console.log(err);
-      },
+  return useMutation(async (value?: any) => {
+    const { regionName, regions, countries } = value;
+    const region = regions.find((item) => item.slug === regionName);
+    let newList = countries;
+    if (region) {
+      newList = countries.filter((item: any) => item.parent_id === region.id);
```
diversemix, line 11: Notice we're missing the `onError` is that deliberate?

diversemix (approved):
LGTM - just checking about the removal of the `onErrors` :ice_cream:

src/hooks/useUpdateCountries.ts
PatFawbertMills: As these are 'internal' mutations to update the local state I don't see the need to console log the errors. I agree we will need to add some better error handling for API communication though.

PR #9, opened 2023-02-22
Fro 157 search context

PatFawbertMills:
Updating the functionality of the search to use aliased query string values rather than internal app state.

This means a number of features become available:
- Search-specific link sharing
- Remembered search (back button)
- Analytics data is captured


src/pages/family/[id].tsx
```diff
@@ -80,6 +87,12 @@ const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({ pa
     <Layout title={page.title}>
       <section className="mb-8">
         <DocumentHead document={page} onCollectionClick={handleCollectionClick} />
+        <div className="bg-gray-200 border-b border-lineBorder py-4">
+          <div className="container">
+            Family/document from search or local store: <br />{" "}
+            <code className="block whitespace-pre-wrap">{JSON.stringify(familyResult, null, "\t")}</code>
```
Comment (diversemix) [inline src/pages/family/[id].tsx:93]: I'm guessing this is left in deliberately for the moment?

diversemix (approved):
My powers of review frontend are fading :smile: ... but for a first cut this looks good, I expect we'll be round tripping this next week a fair bit :+1:

src/utils/buildSearchQuery.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  query = {
+    ...query,
+    keyword_filters,
+  };
+  return query;
+}
+
+// Example search object
```
diversemix, line 71: Maybe a good idea to remove or put a `TODO` on to remove :smile:

src/pages/family/[id].tsx
Comment (PatFawbertMills) [reply src/pages/family/[id].tsx:93]: Yeh its just the UI for the family page getting it ready to work with the search results

src/utils/buildSearchQuery.ts
PatFawbertMills: Done

PR #12, opened 2023-02-28
Family search api integration

PatFawbertMills:
updating search page to use family rather than document

### DIFF
--- a/src/components/blocks/SearchResult.tsx
+++ b/src/components/blocks/SearchResult.tsx
@@ -7,7 +7,7 @@ interface SearchResultProps {
 }
 
 const SearchResult = ({ family }: SearchResultProps) => {
-  const { family_slug, family_name, family_description, family_geography, family_description_match, family_title_match } = family;
+  const { family_slug, family_name, family_description, family_geography, family_description_match, family_title_match, family_documents } = family;
 
   const showMatches = () => {
     if (family_title_match || family_description_match) {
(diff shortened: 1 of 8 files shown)


src/components/blocks/SearchResult.tsx
```diff
@@ -1,69 +1,52 @@
...
-    document_content_type,
-  } = document;
-
-  const formatDate = () => {
-    const eudate = document_date;
-    const dateArr = eudate.split("/");
-    return `${dateArr[1]}/${dateArr[0]}/${dateArr[2]}`;
-  };
-  const [year] = convertDate(formatDate());
+const SearchResult = ({ family }: SearchResultProps) => {
```
diversemix, line 9: Dunno if you want "Grouped" somewhere in this name - as the search response differs - see [here](https://github.com/climatepolicyradar/navigator-backend/blob/3b82e0eeec74fe8b784aa7511d2441d2f8c98d4d/app/api/api_v1/routers/search.py#L80) and [here](https://github.com/climatepolicyradar/navigator-backend/blob/3b82e0eeec74fe8b784aa7511d2441d2f8c98d4d/app/api/api_v1/routers/search.py#L106)

diversemix (approved):
Looking good as a solid first step :+1:

src/components/blocks/SearchResult.tsx
PatFawbertMills: Good suggestion to catch the difference in API response - but this should be handled in the `useSearch` hook instead of a component.

PR #15, opened 2023-03-07
Search results retrieval and cache

PatFawbertMills:
Needed to overhaul the search function within the app, key highlights:
- useSearch now manages responsibility of caching, the search query and any API calls
- Search is now cached in local storage - see the utility file

I'm reluctant to separate the code out further until there is need for more abstraction required.

### DIFF
--- a/src/api/http-common.ts
+++ b/src/api/http-common.ts
@@ -31,10 +31,10 @@ class ApiClient {
       });
   }
 
-  post(url: string, values: any, config = {}) {
+  post<T>(url: string, values: any, config = {}) {
     // console.log(`POST: ${this.baseUrl}${url}`);
     return this.axiosClient
-      .post(`${this.baseUrl}${url}`, values, config)
+      .post<T>(`${this.baseUrl}${url}`, values, config)
       .then((res) => res)
       .catch((err) => {
         console.log(err);
--- a/src/constants/cache.ts
+++ b/src/constants/cache.ts
@@ -1,0 +1,3 @@
+export const CACHE_NAME = "CPR_search_cache";
+export const CACHE_LIMIT = 100;
+export const CACHE_EXPIRY = 1000 * 60 * 60 * 24 * 7; // 7 days
--- a/src/utils/arrayEquality.ts
+++ b/src/utils/arrayEquality.ts
@@ -1,0 +1,11 @@
+export const arrayOfStringdMatch = (a?: string[], b?: string[]) => {
+  if (a === undefined && b === undefined) return true;
+  if (a === undefined || b === undefined) return false;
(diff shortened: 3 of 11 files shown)


src/api/http-common.ts
```diff
@@ -31,10 +31,10 @@ class ApiClient {
       });
   }
 
-  post(url: string, values: any, config = {}) {
+  post<T>(url: string, values: any, config = {}) {
```
diversemix, line 34: Ooo getting slick now with generics :sunglasses:

src/constants/cache.ts
```diff
@@ -0,0 +1,3 @@
+export const CACHE_NAME = "CPR_search_cache";
+export const CACHE_LIMIT = 100;
```
diversemix, line 2: Do you have an idea of how many bytes each cache entry is? ... if its ~10k then we could make the limit more like 1000 :shrug: ... take-it-or-leave-it :smile:

PatFawbertMills: According to this site: https://www.javainuse.com/bytesize
Each search entry is somewhere between 15 - 25kb, depending on the size of the result e.g. number of matches.

100 is a conservative limit. The maximum size of local storage is 5mb.

src/utils/arrayEquality.ts
```diff
@@ -0,0 +1,11 @@
+export const arrayOfStringdMatch = (a?: string[], b?: string[]) => {
```
diversemix, line 1: Could you convert to a `Set` and use the set equality?

src/utils/daysBetween.ts
```diff
@@ -0,0 +1,10 @@
+export const daysBetween = function (date1: number, date2: number) {
```
diversemix, line 1: Hmm, may be time to use `moment` ... 
```
const span = moment.duration(toDate.diff(fromDate)).asDays();
```

diversemix (approved):
Looking great - no show stoppers just comments :smile:

src/constants/cache.ts
diversemix: Sounds good - let's not be one of those sites that take down your browser :laughing:

src/utils/daysBetween.ts
PatFawbertMills: Moment has been deprecated and is not advised for new projects, I could include something like date-fns but I think for now I'll hold off from adding a date library just for this.

src/utils/arrayEquality.ts
PatFawbertMills: I like this idea, I've not used `Set` much so far will round back to it later.

PR #20, opened 2023-03-21
Additional data variables

PatFawbertMills:
Testing out the technique of adding data vars:
- As attributes where the value is rendered
- As a javascript object property on the window, e.g. `analytics.[attribute] = [value]`

### DIFF
--- a/src/components/PassageMatches.tsx
+++ b/src/components/PassageMatches.tsx
@@ -17,7 +17,7 @@ const PassageMatches = ({ passages, setPassageIndex, activeIndex }: TProps) => {
       ) : (
         <div className="divide-lineBorder divide-y passage-matches-list">
           {passages.map((item, index: number) => (
-            <div key={item.text_block_id}>
+            <div key={item.text_block_id} data-analytics-passage={index + 1}>
               <div
                 className={`p-4 cursor-pointer border-x hover:bg-offwhite ${
                   activeIndex === index ? "border-lineBorder bg-grey-200" : "border-transparent"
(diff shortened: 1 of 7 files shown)


src/components/PassageMatches.tsx
```diff
@@ -17,7 +17,7 @@ const PassageMatches = ({ passages, setPassageIndex, activeIndex }: TProps) => {
       ) : (
         <div className="divide-lineBorder divide-y passage-matches-list">
           {passages.map((item, index: number) => (
-            <div key={item.text_block_id}>
+            <div key={item.text_block_id} data-analytics-passage={index + 1}>
```
diversemix, line 20: Just need to be extra careful here as you are going from zero-based to one-based

PatFawbertMills: This is to make it "friendly" for the analytics guys to handle

src/components/blocks/DocumentInfo.tsx
```diff
@@ -21,6 +21,9 @@ type TDoucmentInfoProps = {
 };
 
 const DocumentInfo = ({ heading, text = "", list = [], id = "", tooltip = "", bulleted = false }: TDoucmentInfoProps) => {
+  const dataProps = {};
+  dataProps[`data-analytics-${heading.toLowerCase().trim().replace(" ", "-")}`] = "";
```
diversemix, line 25: Maybe worth protecting this a little more, eg having a fixed list of allowed keys somewhere to check against?

diversemix (approved):
lgtm

PR #22, opened 2023-03-22
update config api

PatFawbertMills:
Add handling of new API structure. Setting API version across the board.

Strip out a lot of old hooks that are no longer used and rely on legacy code architecture.


diversemix (approved):
Nice little tidy up there!

PR #28, opened 2023-03-28
FRO 195 document matches

PatFawbertMills:
- Fixes to the document matches flags
- Add the "jit_query" flag to search object to fix the pagination bug


diversemix (approved):
lgtm

PR #36, opened 2023-03-31
FRO 181 targets updates

PatFawbertMills:
- Add filtering and sorting to targets

### DIFF
--- a/src/utils/sortFilterTargets.ts
+++ b/src/utils/sortFilterTargets.ts
@@ -1,0 +1,59 @@
+import { TTarget } from "@types";
+
+const sourceMapping = {
+  strategy: "executive",
+  plan: "executive",
+  law: "legislative",
+  framework: "legislative",
+  ndc: "executive",
+  policy: "executive",
+};
+
+export const sortFilterTargets = (targets: TTarget[]) => {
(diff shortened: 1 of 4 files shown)


src/utils/sortFilterTargets.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  /* 
+    Net zero targets should be first, 
+    then GHG targets, 
+    then legislative (legal) targets,
+    then targets in the economy-wide sector
+  */
+  t = t.sort((a, b) => {
+    if (a[netZeroKey] === "TRUE" && b[netZeroKey] === "FALSE") {
+      return -1;
```
diversemix, line 31: Again just style we could ease reading with:
```
const a_after_b = 1;
const a_before_b = -1;
```

diversemix (approved):
lgtm

src/utils/sortFilterTargets.ts
PatFawbertMills: It is fairly standard sort syntax but I'm happy to make the change

diversemix: Its not a request to make a change mearly a suggestion

PR #38, opened 2023-04-03
Cache controls

PatFawbertMills:


joel-wright (approved):

src/pages/document/[id].tsx
```diff
@@ -313,6 +313,8 @@ const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({ pa
 export default FamilyPage;
 
 export const getServerSideProps: GetServerSideProps = async (context) => {
+  context.res.setHeader("Cache-Control", "public, max-age=3600, immutable");
```
Comment (diversemix) [inline src/pages/document/[id].tsx:316]: Would it also be a good idea to set this in `ApiClient` 's `get()` function too?

PR #39, opened 2023-04-04
Fix browse link and conditional rendering

PatFawbertMills:
- Browse link on homepage was incorrect
- Conditionally rendering the search page results in poor SEO


diversemix (approved):
lgtm

PR #40, opened 2023-04-04
Set base URL in axios client

PatFawbertMills:
I also cleaned up the axios client to not repeat the base URL in each request.

### DIFF
--- a/src/api/http-common.ts
+++ b/src/api/http-common.ts
@@ -14,7 +14,7 @@ class ApiClient {
     } else {
       this.baseUrl = process.env.API_URL;
     }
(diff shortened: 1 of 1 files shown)


src/api/http-common.ts
```diff
@@ -14,7 +14,7 @@ class ApiClient {
     } else {
       this.baseUrl = process.env.API_URL;
     }
-    this.axiosClient = axios.create();
+    this.axiosClient = axios.create({ baseURL: this.baseUrl, headers: { "Cache-control": "public, max-age=3600, immutable" } });
```
joel-wright, line 17: @PatFawbertMills let's have a chat about the effect this will have.

joel-wright (approved):

PR #42, opened 2023-04-06
Add logic to handle external document links

PatFawbertMills:
- Link out to external documents

Update copy to make the interaction clear for the user

### DIFF
--- a/src/components/document/FamilyDocument.tsx
+++ b/src/components/document/FamilyDocument.tsx
@@ -16,14 +16,17 @@ export const FamilyDocument = ({ document, matches, status }: TProps) => {
   const hasMatches = typeof matches !== "undefined" && matches > 0;
   // PDFs need to have a cdn location
   // HTMLs need a source url / website
-  const canView =
+  const canPreview =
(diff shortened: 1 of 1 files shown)


src/components/document/FamilyDocument.tsx
```diff
@@ -16,14 +16,17 @@ export const FamilyDocument = ({ document, matches, status }: TProps) => {
...
     document.content_type === "application/pdf" ? !!document.cdn_object : document.content_type === "text/html" ? !!document.source_url : false;
+  const canViewSource = !canPreview && !!document.source_url;
 
   const handleClick = (e: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
     e.preventDefault();
-    // If there is no document to render, don't display the document page
-    if (!canView) return;
-    router.push({ pathname: `/documents/${slug}`, query: router.query });
+    // If there is no document to access, don't direct the user
+    if (!canPreview && !canViewSource) return;
```
diversemix, line 26: Wonder if we should even show the clickable thing that we are handling if so, if we should tell the user why clicking is not doing what they're expecting?

diversemix (approved):
lgtm - just put a comment on - would be good to chat / understand this :smile:

PR #43, opened 2023-04-10
FRO 203 methodology updates

PatFawbertMills:


PatFawbertMills:
If you pull it down to run locally - just bear in mind I have improved the title sizing on the acknowledgements branch as I saw we weren't reducing the size for mobile devices which resulted in BIG titles everywhere

src/pages/methodology/index.tsx
```diff
@@ -11,7 +10,6 @@ const Methodology = () => {
 
   return (
     <Layout title={"Methodology"}>
```
diversemix, line 12: Looks like we have removed the CPR methodology page?

PatFawbertMills: Correct - it was never actually navigable and so I thought best to remove it from the repo. The current methodology is hosted on github here: https://github.com/climatepolicyradar/methodology

PR #46, opened 2023-04-12
Testing redirects

PatFawbertMills:
Note: this also has the csv test code in it too


e2e/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM cypress/included:10.10.0
 
 WORKDIR /opt
-RUN npm install cypress-file-upload cypress-pseudo-localization dotenv
+RUN npm i
```
diversemix, line 4: This assumes the existence of a `package.json` file to get the dependencies from

diversemix (approved):
Looking good

PR #47, opened 2023-04-12
FRO 186 faq page and add accordian

PatFawbertMills:
- New FAQ page


src/sites/cclw/constants/faqs.tsx
```diff
@@ -0,0 +1,353 @@
+import { ExternalLink } from "@components/ExternalLink";
+import { LinkWithQuery } from "@components/LinkWithQuery";
+
+type TFAQ = {
+  title: string;
+  content: JSX.Element;
+};
+
+export const FAQS: TFAQ[] = [
```
diversemix, line 9: Suggestion - we could put the FAQs into a json file or similar - just to separate content from formatting?

PR #50, opened 2023-04-20
FRO 216 homepage faq links

PatFawbertMills:
# is now handled by the link component and on the FAQ page it will open the relevant faq if provided.

### DIFF
--- a/src/sites/cclw/pages/faq.tsx
+++ b/src/sites/cclw/pages/faq.tsx
@@ -1,6 +1,6 @@
+import { Fragment } from "react";
 import { AccordianItem } from "@cclw/components/AccordianItem";
 import { FAQS } from "@cclw/constants/faqs";
-import { Fragment } from "react";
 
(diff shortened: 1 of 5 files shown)


src/sites/cclw/pages/faq.tsx
```diff
@@ -13,7 +13,7 @@ const FAQ = () => {
 
         {FAQS.map((faq, i) => (
           <Fragment key={faq.title}>
-            <AccordianItem title={faq.title} startOpen={i === 0}>
+            <AccordianItem id={faq.id} title={faq.title} startOpen={i === 0}>
```
diversemix, line 16: Hmm ... not all have an `id` what happens in that case - does the attribute still get rendered? Looks like it does in `AccordianItem.tsx` as there are not conditionals around `<div className="mb-4" id={id}>`

diversemix (approved):
lgtm

PR #51, opened 2023-04-20
FRO 217 - adjust UI to cater for long search strings

PatFawbertMills:


diversemix (commented):
When you say "Handle" long strings ... I can't see any code change that does any handling? Maybe I'm misunderstanding something somewhere, could you provide a bit more detail on this - thanks :smile:

PR #61, opened 2023-04-26
Remove html iframe viewer and prevent preview for HTML docs

PatFawbertMills:
HTMLs are no longer previewable in an iframe in the app.


diversemix (approved):
lgtm

PR #69, opened 2023-05-02
Missing meta descriptions

PatFawbertMills:
Add meta descriptions for specific pages and update fallback description.

### DIFF
--- a/src/pages/search/index.tsx
+++ b/src/pages/search/index.tsx
@@ -226,7 +226,10 @@ const Search = () => {
   };
 
   return (
-    <Layout title={t("Law and Policy Search")} heading={t("Law and Policy Search")}>
+    <Layout
+      title={t("Law and Policy Search")}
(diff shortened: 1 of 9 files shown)


src/pages/search/index.tsx
```diff
@@ -226,7 +226,10 @@ const Search = () => {
   };
 
   return (
-    <Layout title={t("Law and Policy Search")} heading={t("Law and Policy Search")}>
+    <Layout
+      title={t("Law and Policy Search")}
+      description="Quickly and easily search through the complete text of thousands of climate change law and policy documents from every country."
```
joel-wright, line 231: This seems a bit off... there aren't thousands of law/policy documents from every country, and we don't have documents for all countries, so whichever way I parse this it's a lie :grin:

joel-wright (approved):
lgtm, just the one comment...

PR #76, opened 2023-05-12
FRO 176 breadcrumbs

PatFawbertMills:
Added a new breadcrumbs component and integrated into pages

### DIFF
--- a/src/components/breadcrumbs/Breadcrumbs.tsx
+++ b/src/components/breadcrumbs/Breadcrumbs.tsx
@@ -33,7 +33,7 @@ const BreadCrumb = ({ last = false, label, href = null }: TBreadcrumbLink) => {
 
 export const BreadCrumbs = ({ geography = null, category = null, family = null, label }: TProps) => {
   return (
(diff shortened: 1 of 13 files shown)


src/components/breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,44 @@
...
+  geography?: TBreadcrumbLink;
+  label: string | React.ReactNode;
+};
+
+const BreadCrumb = ({ last = false, label, href = null }: TBreadcrumbLink) => {
+  return (
+    <>
+      <li>
+        {href ? (
+          <LinkWithQuery className="underline" href={href}>
```
diversemix, line 21: I just want to check that this code in `LinkWithQuery` is not going to cause any issue with breadcrumbs (can't quite work it out myself)
```
  // Remove specific slug parameters from the query string
  delete router.query["docId"];
  delete router.query["id"];
  delete router.query["geographyId"];

```

diversemix (approved):
lgtm - just the one comment

PR #77, opened 2023-05-12
unfccc items

PatFawbertMills:
Adding 2 places that link through to the UNFCCC survey

### DIFF
--- a/src/components/blocks/SearchResultList.tsx
+++ b/src/components/blocks/SearchResultList.tsx
@@ -3,6 +3,7 @@ import SearchResult from "./SearchResult";
 import { TMatchedFamily } from "@types";
 import { DOCUMENT_CATEGORIES } from "@constants/documentCategories";
 import Button from "@components/buttons/Button";
+import { UNFCCC_SURVEY } from "@constants/UNFCCCSurvey";
 
 type TProps = {
   category?: (typeof DOCUMENT_CATEGORIES)[number];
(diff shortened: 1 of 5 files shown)


src/components/blocks/SearchResultList.tsx
```diff
@@ -1,13 +1,19 @@
 import { ExternalLink } from "@components/ExternalLink";
 import SearchResult from "./SearchResult";
 import { TMatchedFamily } from "@types";
+import { DOCUMENT_CATEGORIES } from "@constants/documentCategories";
+import Button from "@components/buttons/Button";
 
 type TProps = {
-  category?: string;
+  category?: (typeof DOCUMENT_CATEGORIES)[number];
```
diversemix, line 8: :grin: I would like this to be its own type ... say `CategoryList` ?

src/sites/cclw/components/Articles.tsx
```diff
@@ -14,6 +14,13 @@ type TArticle = {
 };
 
 const ARTICLES: TArticle[] = [
+  {
+    url: "https://docs.google.com/forms/d/e/1FAIpQLSfpo03ytni3SmEqXfFRW38qz1OGyNgN5HhQF-6fW-QcSLZ67A/viewform",
```
diversemix, line 18: I'd say we need a single reference to this form in the code - currently we have the URL it two places

diversemix (approved):
Two comments - first "kind to the next developer" comment - second would be good to do methinks - but your call :smile:

src/components/blocks/SearchResultList.tsx
PatFawbertMills: I will circle back to this later

PR #82, opened 2023-05-24
Order the regions so that Other is at the bottom

diversemix:
Special gift for @PatFawbertMills

### DIFF
--- a/src/utils/sorting.ts
+++ b/src/utils/sorting.ts
@@ -1,3 +1,5 @@
+export type SortFuncType = (data: any, prop: any) => any;
+
 export const sortData = (data, prop) => {
   var myData = data.sort((a, b) => {
     if (a[prop] < b[prop]) {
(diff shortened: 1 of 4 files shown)


src/utils/sorting.ts
```diff
@@ -11,3 +13,16 @@ export const sortData = (data, prop) => {
 
   return myData;
 };
+
+export const sortGeos = (data, prop) => {
+  var myData = data.sort((a, b) => {
+    if (b[prop] == "other") return -1;
```
PatFawbertMills, line 19: Use `===` please, double equals is generally not recommended anymore

PatFawbertMills (changes requested):
Small change to use `===` over `==` please

src/utils/sorting.ts
diversemix: Damn python indoctrination!

PatFawbertMills (approved):
Thanks, looks good

PR #92, opened 2023-06-20
FRO 168 download csv

PatFawbertMills:
- new hook to contain download logic and state
- button on search page gives feedback for: loading, success and error
- csv is downloaded


src/api/http-common.ts
```diff
@@ -1,5 +1,10 @@
 import axios, { AxiosInstance, AxiosResponse } from "axios";
 
+interface ValidationError {
```
diversemix, line 3: Am I being McThickThick? Where is this used?

src/hooks/useDownloadCsv.ts
```diff
@@ -0,0 +1,68 @@
...
+  headers: {
+    accept: string;
+    "Content-Type": string;
+  };
+};
+
+const downloadFile = ({ data, fileName, fileType }) => {
+  const blob = new Blob([data], { type: fileType });
+
+  const a = document.createElement("a");
```
diversemix, line 16: Do we need to make sure this is hidden? Or if we do is the link not clickable? Have we tried this by enabling a mock of a slow connection in the browser debug? (to see how it behaves)

src/hooks/useDownloadCsv.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  return true;
+}
+
+export function useDownloadCsv() {
+  const [status, setStatus] = useState<TLoadingStatus>("idle");
+
+  const download = async (query: TRouterQuery) => {
+    setStatus("loading");
+    const download = await getDownloadCsv(query);
```
diversemix, line 60: Just to point out we have the var `download` hidden at this point (line 58)

diversemix (approved):
If it works - great ... some questions which I'm happy you resolve if you're good :smile:

src/api/http-common.ts
PatFawbertMills: You certainly aren't - it is a legacy of me attempting to cast the axios error which I learnt is not a good idea

src/hooks/useDownloadCsv.ts
PatFawbertMills: Good spot

diversemix (approved):
Thanks for the education :brain:

PR #93, opened 2023-06-23
Remove www from CCLW url

PatFawbertMills:
Key note: only for the domain `www.climate-laws.org`


diversemix (approved):
LGTM ... FTR we tested this locally by editing `/etc/hosts` and using the http protocol and it worked fine

PR #95, opened 2023-07-03
Clear success for download csv on new search

PatFawbertMills:
Was not clearing the status of the download when performing a new search. I have amended the hook to include a way to reset the status.


diversemix (approved):
LGTM

PR #100, opened 2023-07-17
PDCT 198 - display matches html non english

PatFawbertMills:
Support matching passages for:
- Non-English PDFs
- HTML documents

Also
- Display actual language name rather than 3-letter system value
- Add styling updates to the document viewer page

### DIFF
--- a/src/components/document/FamilyDocument.tsx
+++ b/src/components/document/FamilyDocument.tsx
@@ -1,7 +1,9 @@
 import { useRouter } from "next/router";
 import MatchesButton from "@components/buttons/MatchesButton";
-import { TDocumentContentType, TDocumentPage, TLoadingStatus } from "@types";
 import { Loading } from "@components/svg/Icons";
+import useConfig from "@hooks/useConfig";
+import { getLanguage } from "@helpers/getLanguage";
+import { TDocumentContentType, TDocumentPage, TLoadingStatus } from "@types";
 
 type TProps = {
   document: TDocumentPage;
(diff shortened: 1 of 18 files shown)


src/components/document/FamilyDocument.tsx
```diff
@@ -71,11 +76,11 @@ export const FamilyDocument = ({ document, matches, status }: TProps) => {
           {renderContentType(content_type)}
           {!!language && (
             <span>
-              {language.toUpperCase()}
+              {getLanguage(language, languages)}
```
diversemix, line 79: This is the second place we create a view of `language` + `variant` ... is this worth its own helper? (minor)

src/pages/documents/[id].tsx
```diff
@@ -60,44 +55,61 @@ const DocumentPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({
...
             <div className="container flex-1">
-              <div className="flex flex-col md:flex-row justify-between items-center pb-4 border-b border-lineBorder gap-4">
-                {passageMatches.length > 0 && (
-                  <h3>Document matches for {`'${router.query[QUERY_PARAMS.query_string]}' (${passageMatches.length})`}</h3>
+              <div className="md:flex md:h-[80vh]">
+                {hasPassageMatches && (
+                  <div className={`overflow-y-scroll pr-4 max-h-[30vh] md:block md:max-h-full ${passageClasses(document.content_type)}`}>
+                    <div className="my-4" data-cy="document-matches-description">
+                      <p className="">
+                        {passageMatches.length} matches for "<b>{`${router.query[QUERY_PARAMS.query_string]}`}</b>"
```
Comment (diversemix) [inline src/pages/documents/[id].tsx:77]: What's displayed if we're browsing and `QUERY_PARAMS` is empty?

diversemix (approved):
A couple of minor comments/questions but lgtm :+1:

src/pages/documents/[id].tsx
Comment (PatFawbertMills) [reply src/pages/documents/[id].tsx:77]: It'll get caught at the `{hasPassageMatches && (` above this on line 73, i.e it will not attempt to render any of this section unless there are matching passages.

For future reference - passage matches are only applicable to string searches.

PR #111, opened 2023-07-31
Don't set 'All' as the category as it is invalid

PatFawbertMills:
Bug was:
- Incorrectly setting the category in the query string to 'All' this was then passing null in the API which was returning a 422


diversemix (approved):
lgtm

PR #112, opened 2023-08-15
PDCT 287 matomo set up

PatFawbertMills:
Key notes:
- The `utils/matomo.ts` file is not being used, but I've kept where I got to with it for reference. The matomo team advised us to use MatomoTagManager instead - which you can see initialised in the `CookieConsent.tsx` component
- I rewrote the logic for `LinkWithQuery.tsx` to be exclusive to our designated query params as what was happening was google analytics tracking codes and internal next codes were persisting around the site unnecessarily
- Removing some legacy json test files that weren't in use


src/components/cookies/CookieConsent.tsx
```diff
@@ -34,6 +36,17 @@ export const CookieConsent = () => {
...
       })(window, document, "script", "dataLayer", "GTM-NTNH983");
+      // Matomo
+      (function () {
+        var _mtm = (window._mtm = window._mtm || []);
+        _mtm.push({ "mtm.startTime": new Date().getTime(), event: "mtm.Start" });
+        var d = document,
+          g = d.createElement("script"),
+          s = d.getElementsByTagName("script")[0];
+        g.async = true;
+        g.src = "https://cdn.matomo.cloud/climatelaws.matomo.cloud/container_l8tXApwR.js";
```
diversemix, line 47: Question - do we want to pull out the google id "GTM-NTNH983" and the specific js file for matamo into config?

diversemix (approved):
lgtm :+1:

PR #117, opened 2023-08-22
Fix api urls

PatFawbertMills:
- Remove all traces of `?group_documents=true` from the code
- Update the dev api in the Makefile to `app.dev`


PatFawbertMills:
Incorrectly committed changes to LinkWithQuery component

PR #127, opened 2023-09-27
Add prettier configuration

PatFawbertMills:
First pass based on what I was using on my local vs code set up.

We should look to scrutinise and update this regularly.


katybaulch (approved):
looks good to me 👍

PR #130, opened 2023-09-27
GTM consent order

PatFawbertMills:
Previous attempt did not work as expected within GTM.

Main issues solved here:
- Consent was being set **too late** in the flow
- The previous `gtag` method does not work, I was not declaring and using the gtag module properly (see custom.d.ts)


katybaulch (approved):
Looks good 👍

PR #135, opened 2023-10-03
Prettier config

katybaulch:


PatFawbertMills (approved):
Code looks good.

Only thing for future - I think because the branch was re-created from the old one that was merged the commit history looks a bit janky, and the name isn't quite reflective of the work. So I think it is best to branch from main (if possible) and rename the PR to summarise your changes. E.g. for this one could be "Update ESLint rules"

PatFawbertMills:
Looks like the new rules also brought in some build errors!

PR #137, opened 2023-10-11
Feature/pdct 455 put country page search box at bottom of family and physical document

katybaulch:
Added country page search box to the bottoms of the family and physical document pages.

Linked to [this backend change](https://github.com/climatepolicyradar/navigator-backend/pull/175).

### DIFF
--- a/src/pages/document/[id].tsx
+++ b/src/pages/document/[id].tsx
@@ -27,10 +27,9 @@ import { getCountryName, getCountrySlug } from "@helpers/getCountryFields";
 import { getOrganisationNote } from "@helpers/getOrganisationNote";
 import { sortFilterTargets } from "@utils/sortFilterTargets";
 import { MAX_FAMILY_SUMMARY_LENGTH } from "@constants/document";
-import { TFamilyPage, TMatchedFamily, TTarget } from "@types";
+import { TFamilyPage, TMatchedFamily, TTarget, TGeographySummary } from "@types";
 import Tooltip from "@components/tooltip";
 import SearchForm from "@components/forms/SearchForm";
-import { TGeographySummary } from "@types";
 
 type TProps = {
   page: TFamilyPage;
@@ -99,11 +98,8 @@ const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({ pa
   const sourceLogo = page?.organisation === "CCLW" ? "grantham-logo.png" : null;
   const sourceName = page?.organisation === "CCLW" ? "Grantham Research Institute" : page?.organisation;
 
-  const allDocumentsCount = () => {
-    const allDocumentsCount =
-      geographySummary.family_counts.Legislative + geographySummary.family_counts.Executive + geographySummary.family_counts.UNFCCC;
-    return allDocumentsCount;
-  };
+  const totalDocsInPageGeography =
+    geographySummary.family_counts.Legislative + geographySummary.family_counts.Executive + geographySummary.family_counts.UNFCCC;
 
   const mainDocs = page.documents.filter((doc) => doc.document_role && doc.document_role.toLowerCase().includes("main"));
   const otherDocs = page.documents.filter((doc) => !doc.document_role || !doc.document_role.toLowerCase().includes("main"));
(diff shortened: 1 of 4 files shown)


src/pages/document/[id].tsx
```diff
@@ -283,6 +298,32 @@ const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({ pa
               )}
             </section>
           ))}
+
+          <section className="mt-10" data-cy="top-documents">
+            <h3 className="mb-4">Documents</h3>
```
Comment (PatFawbertMills) [inline src/pages/document/[id].tsx:303]: I think this title needs to be slightly more descriptive, "Search for more", "Search in {countryName}" perhaps.

Might be worth speaking to @alaniswright on this

src/pages/documents/[id].tsx
```diff
@@ -156,6 +175,36 @@ const DocumentPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({
             </div>
           </section>
         )}
+
+        <div className="flex place-content-center">
+          <div className="mt-8 mx-auto max-w-screen-md px-4 lg:px-0">
+            <section className="mt-10" data-cy="top-documents">
```
Comment (PatFawbertMills) [inline src/pages/documents/[id].tsx:181]: I believe this is the same block copied from the geo page, and we now have it in 3 places:
- Geo
- Document (family)
- Document (physical document)

Which leads me to believe we should be refactoring into its own managed component to keep it as DRY as possible.

Also I noticed (not your fault) that the `data-cy` attribute has been incorrectly copy pasta'd on the geo page (by me) and should be more indicative of the content too. Perhaps `data-cy="country-search"`

src/pages/document/[id].tsx
```diff
@@ -307,13 +350,21 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
...
 
-  if (!familyData) {
+  if (geographyCode) {
+    try {
+      const { data: returnedData }: { data: TGeographySummary } = await client.get(`/summaries/geography/${geographyCode}`);
+      geographySummaryData = returnedData;
+    } catch (error) {}
+  }
+
+  if (!familyData || !geographySummaryData) {
```
Comment (PatFawbertMills) [inline src/pages/document/[id].tsx:367]: I personally don't think we should prevent the entire page loading if we don't get the geo summary data loaded.

Similar to how we handle targets, just hide the search box / don't display the country-specific data if we don't have it.

src/pages/documents/[id].tsx
```diff
@@ -180,7 +231,16 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
...
+  if (familyData) geographyCode = familyData.geography;
+
+  if (geographyCode) {
+    try {
+      const { data: returnedData }: { data: TGeographySummary } = await client.get(`/summaries/geography/${geographyCode}`);
+      geographySummaryData = returnedData;
+    } catch (error) {}
+  }
+
+  if (!documentData || !familyData || !geographySummaryData) {
```
Comment (PatFawbertMills) [inline src/pages/documents/[id].tsx:243]: See similar comment on the `/document` page

PatFawbertMills (changes requested):

src/pages/document/[id].tsx
Comment (alaniswright) [reply src/pages/document/[id].tsx:303]: I think this component just needs to include what is in the blue box.

We don't need the title "Documents" that appears on the country page to also appear when we repeat this element at the bottom of document pages

Comment (katybaulch) [reply src/pages/document/[id].tsx]: Sorted.

src/pages/document/[id].tsx
```diff
@@ -283,6 +301,17 @@ const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({ pa
               )}
             </section>
           ))}
+
+          {!!geographySummary && (
+            <section className="mt-8" data-cy="top-documents">
+              <DocumentSearchForm
+                placeholder={`Search the full text of ${totalDocsInPageGeography} documents from ${geographyName}`}
```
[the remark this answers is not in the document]
Comment (PatFawbertMills) [reply src/pages/document/[id].tsx:308]: I realise this would be a logical impossibility apart from on the geo page if the given geo has no documents in it - so safe to leave it as-is for now!

PatFawbertMills (approved):
LGTM

src/pages/document/[id].tsx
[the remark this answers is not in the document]
Comment (katybaulch) [reply src/pages/document/[id].tsx]: Have changed anyway - good to explicitly hide!

PR #140, opened 2023-11-06
PDCT 546 vespa control

PatFawbertMills:
Add 'secret' button to control whether to use the vespa API, it simply adds `?use_vespa=true` to the query string

### DIFF
--- a/src/hooks/useSearch.ts
+++ b/src/hooks/useSearch.ts
@@ -20,12 +20,14 @@ async function getSearch(query = initialSearchCriteria) {
     },
   };
 
+  // TODO: remove this later when BE is updated
+  const url = `/searches${query.use_vespa ? "?use_vespa=true" : ""}`;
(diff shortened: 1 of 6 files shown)


src/hooks/useSearch.ts
```diff
@@ -20,12 +20,14 @@ async function getSearch(query = initialSearchCriteria) {
     },
   };
 
+  // TODO: remove this later when BE is updated
+  const url = `/searches${query.use_vespa ? "?use_vespa=true" : ""}`;
```
THOR300, line 24: Does this need to be `True` with an uppercase T?

THOR300 (commented):
Small question!

src/hooks/useSearch.ts
PatFawbertMills: No, lower case was working fine when I tested this

diversemix (approved):
lgtm

PR #141, opened 2023-11-06
Revert change to adobe page index

PatFawbertMills:
The issue is not with adobe in this case but instead the page indexing difference between vespa and opensearch.

In opensearch the page indexes from 1, whereas in vespa it appears to be 0.

I think rather than set lots of business logic in code to check which search we are using we should update vespa to match opensearch - even if this means we switch it out later.

For clarity:
**In Vespa's search results, the passage matches are all out by 1 page.**

This PR corrects the page number for OpenSearch results whilst we await Vespa being changed to match the logic for page numbers.

### DIFF
--- a/src/hooks/usePDFPreview.ts
+++ b/src/hooks/usePDFPreview.ts
@@ -15,7 +15,7 @@ function generateHighlights(document: TDocumentPage, documentPassageMatches: TPa
         source: document.import_id,
         selector: {
           node: {
-            index: passage.text_block_page,
+            index: passage.text_block_page - 1,
           },
           subtype: "highlight",
           // WE CAN ASSUME BLOCK_COORDS IS ALWAYS LENGTH 4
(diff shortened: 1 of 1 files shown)


src/hooks/usePDFPreview.ts
```diff
@@ -15,7 +15,7 @@ function generateHighlights(document: TDocumentPage, documentPassageMatches: TPa
         source: document.import_id,
         selector: {
           node: {
-            index: passage.text_block_page,
+            index: passage.text_block_page - 1,
```
diversemix, line 18: How does it know this is Vespa and not opensearch?

diversemix: see slack chat https://climate-policy-radar.slack.com/archives/C02ET708QDP/p1699280107665369

diversemix (approved):
lgtm

src/hooks/usePDFPreview.ts
PatFawbertMills: I am purposely not adding any search provider business logic into the PDF previewer as I think we need to fix this upstream from the FE. This means Vespa results will be out by -1.

PR #142, opened 2023-11-21
Initialise GTM correctly

PatFawbertMills:
After doing some reading into GTM again I discovered that I don't think I was initialising the GTM correctly.


katybaulch (approved):
lgtm :+1:

PR #143, opened 2023-11-27
PDCT 534 - cclw header footer

PatFawbertMills:
All changes are for the CCLW theme only:
- New CCLW logo
- Removing logos and old CCLW text from header
- Moving logos to the footer
- New 'powered by' section on landing page
- CCLW plausible code
- Footer text updating
- Favicon updating


src/components/cookies/CookieConsent.tsx
```diff
@@ -123,6 +108,24 @@ export const CookieConsent = () => {
...
+                function gtag(){w[l].push(arguments);}
+                gtag('js', new Date());
+                gtag('config', i);
+              })(window, "dataLayer", "UA-153841121-2");
+            `}
+          </Script>
+        </>
+      )}
+      {theme === "cclw" && (
+        <Script id="cclw-plausible" defer data-domain="climate-laws.org" strategy="afterInteractive" src="https://plausible.io/js/script.js" />
```
THOR300, line 127: What will this line do?

THOR300 (approved):
Can't see any issues with this!

src/components/cookies/CookieConsent.tsx
PatFawbertMills: loads a third party script called plausible if the theme is set to cclw.

only reason I am keeping it in the cookie consent file is just so it is with all the other 3rd party tracking / library stuff

PR #146, opened 2023-12-05
Add use_vespa to the unique identifier for caching

PatFawbertMills:
It was an oversight to not include the vespa control to the caching identifier initially.

The code below amends this.


diversemix (approved):
lgtm

PR #148, opened 2023-12-07
Fix missing maps and update intellisearch for matching part of country name

PatFawbertMills:


src/components/forms/SearchDropdown.tsx
```diff
@@ -52,7 +51,7 @@ export const SearchDropdown = ({ show = false, term, handleSearchClick, largeSpa
 
     // When multiple geography matches are found e.g., Sudan and South Sudan, select the geography with the longest
     // display name.
-    let geography;
+    let geography: TGeography;
```
katybaulch, line 54: Whoops, thanks!!

katybaulch (approved):
Looks good - I haven't spun this up locally to check though!

PR #149, opened 2023-12-20
Accesibility updates

PatFawbertMills:
I've upgraded the rating from 86 and 85 respectively to 96. I am not tackling the remaining 4% as it involves the contrast ratio of the brand blue which is a problem for another day.

Screenshot 2023-12-20 at 15 33 40

Screenshot 2023-12-20 at 15 31 36


katybaulch (approved):
Looks good - great stuff!! 🦖

PR #153, opened 2024-01-31
PDCT-727 Theme factory

PatFawbertMills:
This PR contains the work carried out to separate the theme-specific code from the core application code.

Please see the ticket for additional information and context.


src/components/cookies/CookieConsent.tsx
```diff
@@ -40,7 +40,7 @@ export const CookieConsent = () => {
   };
 
   // gtm controls for CCLW theme only
-  const cclwAnalyticsAllowed = theme === "cclw" && enableAnalytics;
+  // const cclwAnalyticsAllowed = theme === "cclw" && enableAnalytics;
```
diversemix, line 43: Commented out code - should this get removed?

src/components/cookies/CookieConsent.tsx
```diff
@@ -108,7 +108,8 @@ export const CookieConsent = () => {
           </Script>
         </>
       )}
-      {cclwAnalyticsAllowed && (
+      <ThemeAnalytics enableAnalytics={enableAnalytics} />
+      {/* {cclwAnalyticsAllowed && (
```
diversemix, line 112: Same point here - remove?

src/constants/cache.ts
```diff
@@ -1,3 +1,6 @@
-export const CACHE_NAME = "CPR_search_cache";
+// const week = 1000 * 60 * 60 * 24 * 7;
```
diversemix, line 1: Should they stay or should they go? :notes:

src/pages/[page].tsx
```diff
@@ -0,0 +1,106 @@
...
+  page?: TPage & {
+    notFound?: boolean;
+  };
+};
+
+export default function Page({ page }: TProps) {
+  // TODO: fix this properly
+  // Next is throwing a NEXT_REDIRECT error under the hood when attemping to navigate to a missing page at root, e.g. /missing-page
+  if (!page || page.notFound) {
+    return window.location.replace("/not-found");
```
Comment (diversemix) [inline src/pages/[page].tsx:28]: Do we want to console.error something here?

src/redirects/redirects.ts
```diff
@@ -1,42 +1,20 @@
...
-  { source: "/methodology", destination: "/", permanent: true },
-  { source: "/faq", destination: "/", permanent: true },
-];
-
-const theme = process.env.THEME;
+// Please note: this file is no longer being importing into the middleware
 
 function getRedirects(): Array<TRedirect> {
-  const REDIRECT_FILE = process.env.REDIRECT_FILE || "CCLW_redirects.json";
+  const REDIRECT_FILE = process.env.REDIRECT_FILE || `../../themes/${process.env.THEME}/redirects.json`;
```
diversemix, line 7: This default is at odds with the way the code is written in other places ... there the path is absolute, e.g. `/themes/${process.env.THEME}/layouts/main` and not relative. I prefer the absolute as it assumes we are running in a container and IMO is safer :)

diversemix (approved):
Some things to think about - but all looks great with some neat improvements along the way :rocket:

src/redirects/redirects.ts
PatFawbertMills: Yes I agree - I will have another look at this but I feel like I tried this first. I'll also look at using a path alias in the tsconfig

src/pages/[page].tsx
Comment (PatFawbertMills) [reply src/pages/[page].tsx:29]: I've added a `console.warn` for now

src/constants/cache.ts
PatFawbertMills: GO 🔥

src/redirects/redirects.ts
PatFawbertMills: Ah - this file is dead now anyway 😆 it has been superseded by the theme-level configuration in the `next.config.js` file

PatFawbertMills:
Thanks Peter all good catches! 🦅 👁️

PR #157, opened 2024-02-07
PDCT 814 - download csv popup

PatFawbertMills:
When users click on the link to download the search the app now opens a pop-up modal to display a notification regarding the terms and then options to download or cancel.


katybaulch (approved):
Nice stuff 👍🏻

PR #162, opened 2024-03-06
PDCT 783 - world map

PatFawbertMills:
**Important:** As part of this release we have upgraded Next to the latest for `v13`. This was due to an issue cause by SWC as part of an older version of Next.

Adding an interactive world map. The map acts as both a navigational tool to draw users to the geography pages, as well as presenting some basic visualisation of the spread of documents:
image

Screenshot 2024-04-11 at 11 23 59

## Proposed version

- [ ] Skip auto-tagging
- [ ] Patch
- [x] Minor version
- [ ] Major version


src/components/map/WorldMap.tsx
```diff
@@ -0,0 +1,339 @@
...
+type TGeoMarkers = {
+  lawsPolicies: number;
+  unfccc: number;
+};
+
+type TGeographyWithCoords = TGeography & { coords: TPoint; familyCounts: TGeoFamilyCounts; markers: TGeoMarkers };
+
+type TGeographiesWithCoords = { [key: string]: TGeographyWithCoords };
+
+const geoStyle = (isActive: boolean) => {
```
diversemix, line 39: Not a blocker ... Just a note to chat about trying to separate style from behavior into separate files so its clear what is changing.

src/components/map/WorldMap.tsx
```diff
@@ -0,0 +1,339 @@
...
+          <p>
+            <LinkWithQuery href={`/geographies/${geography.slug}`}>View territory profile</LinkWithQuery>
+          </p>
+        </>
+      )}
+    </>
+  );
+};
+
+// TODO:
```
diversemix, line 114: Do we have a ticket for this?

PatFawbertMills: Ah - that was my TODO list for the feature, I have done it but forgot to delete the TODO

src/components/map/WorldMap.tsx
```diff
@@ -0,0 +1,339 @@
...
+
+  const handleGeoHover = (e: React.MouseEvent<SVGPathElement>, hoveredGeo: string) => {
+    setActiveGeography("");
+    openToolTip([e.clientX, e.clientY], hoveredGeo);
+  };
+
+  const handleGeographySelected = (selectedCountry: TGeographyWithCoords) => {
+    setActiveGeography(selectedCountry.display_value);
+    setMapCenter(selectedCountry.coords);
+    setMapZoom(5);
```
diversemix, line 174: Be interesting to know how the map behaves with browser zoom too - guessing "5" is a fixed level?

src/components/tooltip/index.tsx
```diff
@@ -10,11 +10,10 @@ interface TooltipProps {
 
 const Tooltip = ({ id, tooltip, icon = "?", place, interactableContent }: TooltipProps) => {
   return (
-    <div>
+    <div className="">
```
diversemix, line 13: Why the need for a blank class name?

src/styles/main.scss
```diff
@@ -214,7 +214,7 @@ input[type="search"]::-webkit-search-results-decoration {
 
 .subnav-active {
   position: relative;
-  @apply z-20;
+  // @apply -z-10;
```
diversemix, line 217: commented code - remove?

diversemix (approved):
Couple of comment - but looks good to go :hamburger:

src/components/map/GeographySelect.tsx
```diff
@@ -0,0 +1,63 @@
...
+  };
+
+  useEffect(() => {
+    if (!input.length) {
+      setSuggestList([]);
+      return;
+    }
+    let filteredList = [];
+    filteredList = Object.values(list).filter((item) => {
+      /* Make sure item hasn't already been selected and limit list to 20 items */
```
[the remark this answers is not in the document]
PatFawbertMills: No idea why I wrote this, the suggestions are already limited to 10 and we don't have a pre-selected list anyway! I've removed the comment and the code that contained the selected list.

src/components/map/WorldMap.tsx
PatFawbertMills: The map stays relatively consistent in regards to its rendering (I presume due to it being an SVG) whereas the UI around it does obviously increase as you zoom. I hadn't realised this would be the case.

The zoom here refers to the `react-simple-maps` configuration of zoom on a scale of 1-9. 1 being incredibly zoomed in to the point of uselessness (particular given we are not using a hyper-detailed map) and 9 being the entire world.

PR #163, opened 2024-03-07
PDCT 739 handle vespa default

PatFawbertMills:
- Remove hidden link that was triggering a vespa search
- Remove query string to `/searches` API url
- Remove old open search payload property `jit_query`


diversemix (approved):
Great tidy :broom:

PR #165, opened 2024-03-14
Update adobe PDF viewer src file, remove legacy comments

PatFawbertMills:
Clean up the pdf loaders and update the adobe script

### DIFF
--- a/src/api/pdf.js
+++ b/src/api/pdf.js
@@ -9,13 +9,14 @@ then your use, modification, or distribution of it requires the prior
 written permission of Adobe.
 */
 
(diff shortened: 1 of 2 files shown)


diversemix (approved):
Nice clean up - can you add a note why the two "register..." functions were removed?

katybaulch (approved):
lgtm 🦖

PatFawbertMills:
> Nice clean up - can you add a note why the two "register..." functions were removed?

Yes sure - they were added by the previous developer and were not being used / nor possibly even supported anymore as that was > 2 years ago. There is documentation online regarding the set up of event handlers: https://developer.adobe.com/document-services/docs/overview/pdf-embed-api/howtos/

PR #166, opened 2024-03-20
Remove retired search parameters

olaughter:
The backend no longer uses these, removing them here means we can safely remove them from the backend


PatFawbertMills (approved):
Nice

PR #170, opened 2024-04-09
Up limit to 100 on download CSV payload object

PatFawbertMills:
Discovered there was an issue with the download-csv endpoint only returning a maximum of 10 results.

I assume there was a change in the back-end / API logic to use this limit where previously it was ignoring it.


katybaulch (approved):
lgtm 🦖

PR #173, opened 2024-04-09
Feature/pdct 989 add trunk to nav frontend

katybaulch:


diversemix (approved):
Do we have a makefile or similar to install - for @PatFawbertMills 
?

katybaulch:
> Do we have a makefile or similar to install - for @PatFawbertMills ?

Adding now

PR #174, opened 2024-04-09
Feature/pdct 988 add autotag release to nav frontend

katybaulch:
# What's changed

Add auto tag and release to navigator frontend. 

## Proposed version

- [x] Skip auto-tagging
- [ ] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


server.js
```diff
@@ -15,6 +15,7 @@ app.prepare().then(() => {
     handle(req, res, parsedUrl);
   }).listen(3000, (err) => {
     if (err) throw err;
+    /* trunk-ignore(eslint/no-console) */
```
THOR300, line 18: Why is this line needed!?

THOR300 (approved):
LGTM

server.js
katybaulch: Because ESLint has a no-console rule to avoid people leaving in debug statements, and the line below has a debug statement that we want to leave in (have checked this with Patrick).

PR #176, opened 2024-04-11
Bump release CI job version to v6

katybaulch:
# What's changed

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


PatFawbertMills (approved):
Thank you!

PR #189, opened 2024-06-06
CCLW homepage redesign

PatFawbertMills:
# What's changed
PDCT 863

- Redesign of the CCLW homepage

Noteworthy changes:
- Switch the font to Neue Haas Grotesk Display
- Remove acknowledgements and merge into about page
- Ensure World Map loads gracefully and displays only when it has data

## Proposed version

- [ ] Skip auto-tagging
- [ ] Patch
- [x] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


src/components/buttons/SquareButton.tsx
```diff
@@ -16,7 +16,7 @@ interface ButtonProps {
 
 const Button = ({ children, type = "button", disabled = false, onClick = null, color = "default", id, extraClasses = "", ...props }: ButtonProps) => {
   // let classes = "bg-indigo-600 text-white border border-indigo-600 hover:bg-white hover:border-white hover:text-indigo-600";
```
katybaulch, line 18: Do we still need this?

katybaulch (approved):
Woohoo!! Nice work

src/components/buttons/SquareButton.tsx
PatFawbertMills: Old file - deleted it

PR #197, opened 2024-07-09
PDCT 1083 search results rework

PatFawbertMills:
# What's changed
- New pagination architecture using continuation tokens = results up to 500 (20 per page up to 25 pages)
- Increased highlighted passages to 500
- FE search cache has been disabled via a boolean property - so it can be enabled later if we need
- Lots of small UI bug fixes
- New constants to control things like pagination 

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


src/components/buttons/SearchMatchesButton.tsx
```diff
@@ -8,6 +8,10 @@ type TProps = {
   onClick?: () => void;
 };
 
+const formatCount = (count: number) => {
+  return count >= 500 ? "More than 500" : count;
```
diversemix, line 12: Shouldn't this be `MAX_RESULTS` ?

src/components/drawer/FamilyMatchesDrawer.tsx
```diff
@@ -80,8 +72,7 @@ export const FamilyMatchesDrawer = ({ family }: TProps) => {
                   onClick={(e) => handleViewOtherDocsClick(e, document.document_slug)}
                   extraClasses="text-sm text-blue-600"
                 >
-                  View {document.document_passage_matches.length} {pluralise(document.document_passage_matches.length, "match", "matches")}{" "}
-                  highlighted in document
+                  View all matches highlighted in document
```
diversemix, line 75: We've lost the pluralisation here - is that intentional?

src/constants/searchCriteria.ts
```diff
@@ -1,16 +1,16 @@
 import { minYear, currentYear } from "@constants/timedate";
-import { PER_PAGE } from "@constants/paging";
+import { RESULTS_PER_PAGE } from "@constants/paging";
 import { TSearchCriteria } from "@types";
 
 export const initialSearchCriteria: TSearchCriteria = {
   query_string: "",
   exact_match: false,
-  max_passages_per_doc: 100,
+  max_passages_per_doc: 50,
```
diversemix, line 8: Is it worth pulling this out as a const too?

src/components/drawer/FamilyMatchesDrawer.tsx
PatFawbertMills: Yes, unfortunately the data coming from the API is not reliable enough or accurate to display this. So we've defaulted to more general messaging.

src/pages/documents/[id].tsx
```diff
@@ -52,11 +63,11 @@ const scrollToPassage = (index: number) => {
 const DocumentPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({ document, family }: TProps) => {
   const [passageIndex, setPassageIndex] = useState(null);
   const [passageMatches, setPassageMatches] = useState<TPassage[]>([]);
+  const [totalNoOfMatches, setTotalNoOfMatches] = useState(0);
   const router = useRouter();
   const startingPassage = Number(router.query.passage) || 0;
-  const { status, families, searchQuery } = useSearch(router.query, null, document.import_id, !!router.query[QUERY_PARAMS.query_string]);
+  const { status, families, searchQuery } = useSearch(router.query, null, document.import_id, !!router.query[QUERY_PARAMS.query_string], 500);
```
Comment (diversemix) [inline src/pages/documents/[id].tsx:69]: The `500` ... the param is `NoOfPassagesPerDoc` ... but the `max_passages_per_doc` is defined as 50 earlier on?

src/pages/documents/[id].tsx
```diff
@@ -164,41 +173,33 @@ const DocumentPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({
...
                     )}
                   </div>
-                  {hasPassageMatches && (
+                  {totalNoOfMatches > 0 && (
                     <>
-                      <div className="my-4 text-sm" data-cy="document-matches-description">
+                      <div className="my-4 text-sm pr-4 pb-4 border-b" data-cy="document-matches-description">
                         <p>
-                          {passageMatches.length} matches for "<b>{`${router.query[QUERY_PARAMS.query_string]}`}</b>"
+                          Displaying {renderPassageCount(totalNoOfMatches)} matches for "<b>{`${router.query[QUERY_PARAMS.query_string]}`}</b>"
```
Comment (diversemix) [inline src/pages/documents/[id].tsx:180]: Do we need pluralisation here?

src/pages/documents/[id].tsx
Comment (PatFawbertMills) [reply src/pages/documents/[id].tsx:69]: This is so that we can optionally override the default in some given situations. Within the context of a document, we want as many passages returned as possible. I will replace this with the constant to make it clearer!

PR #199, opened 2024-07-17
Pdct 1250/faq update

annaCPR:
# What's changed
- updated text on 3 of the FAQs to reflect the increased search result limit and the actual number of results returned
- updated text on the Download This Search popup to include a note on the actual number of results returned

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


katybaulch (approved):
nice one!!

PR #202, opened 2024-08-12
Feature/pdct 1303 making a skeleton mcf custom app

odrakes-cpr:
# What's changed

Skeleton folder to add a custom app ( file structure is still subject to change)
- adding new folder to the themes directory and adding in the required files to get this page live
- there is no functionality in this directory as of yet as is it very bare bones
- the routes are dynamic so to test locally you need to update the theme env var

Still tbc : 

`tsconfig.json` does not include the new custom app path, left this as is for now
domain name left alone, as this is still a skeleton
have not made any changes to the `robots.txt` file to affect crawling

## Proposed version

- [ ] Skip auto-tagging
- [ ] Patch
- [x] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


katybaulch (approved):
amazing skeleton!! ☠️ nice work!

PatFawbertMills (approved):
Good start

PR #203, opened 2024-08-14
feature/pdct-1336-add-search-component-to-skeleton-mcf-custom-app

odrakes-cpr:
feat: add search component to skeleton app

- this reuses the component in the cpr app and repurposes it for the mcf custom app
- updates to the data, filtering and a methodology page will be handled in a separate PR

# What's changed
Screenshot 2024-08-14 at 17 21 28

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -5,7 +5,6 @@ const config: Config = {
   verbose: true,
 
   moduleNameMapper: {
-    // Map `@mcf/components` to the correct directory
     "^@constants/(.*)$": "<rootDir>/src/constants/$1",
     "^@components/(.*)$": "<rootDir>/src/components/$1",
     "^@hooks/(.*)$": "<rootDir>/src/hooks/$1",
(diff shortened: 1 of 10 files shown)


jest.config.ts
```diff
@@ -0,0 +1,18 @@
+import type { Config } from "jest";
+
+const config: Config = {
+  testEnvironment: "jsdom",
+  verbose: true,
+
+  moduleNameMapper: {
+    // Map `@mcf/components` to the correct directory
```
PatFawbertMills, line 8: Feel like we don't need this comment

PatFawbertMills (changes requested):
Minor point on the comment left in the jest aliases but otherwise all good

PatFawbertMills (approved):
Nice one

PR #204, opened 2024-08-14
feat: update copy across FAQs, methodology and terms of use

odrakes-cpr:
# What's changed

Updating the copy across FAQs, methodology and terms of use as per [notion]( https://www.notion.so/climatepolicyradar/CPD-text-to-update-FAQs-Methodology-Terms-fa02cb8a462e4f0d84221219eddcc93c)

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/themes/cclw/pages/terms-of-use.tsx
+++ b/themes/cclw/pages/terms-of-use.tsx
@@ -185,7 +185,9 @@ const TermsOfUse = () => {
             <p>
               The CPR Database and App contains data that were created by third parties. The data includes document summaries that were written by
               third parties and/or information about the documents (”meta-data”) assigned by third parties, e.g. classifications to sector or legal
-              principle.
+              principle, and Law and Policy document curation and summaries; document metadata from{" "}
+              <ExternalLink url="https://newclimate.org/about-us/legal-notice">Climate Policy Database </ExternalLink> has been adapted by Climate
+              Policy Radar and the Grantham Research Institute LSE.
             </p>
             <p>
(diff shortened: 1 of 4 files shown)


themes/cclw/pages/terms-of-use.tsx
```diff
@@ -185,7 +185,9 @@ const TermsOfUse = () => {
             <p>
               The CPR Database and App contains data that were created by third parties. The data includes document summaries that were written by
               third parties and/or information about the documents (”meta-data”) assigned by third parties, e.g. classifications to sector or legal
-              principle.
+              principle, and Law and Policy document curation and summaries; document metadata from{" "}
```
odrakes-cpr, line 188: might be one for Alan, @PatFawbertMills but not sure if this is how he intended the copy to be added to the terms, it wasn't all too clear in the notion doc.
Screenshot 2024-08-14 at 17 08 09

PatFawbertMills: I think this is in the wrong place. Judging by Alan's notion page the update needs to be made to the table below where we have the list of sources

PatFawbertMills (changes requested):

themes/cclw/pages/terms-of-use.tsx
odrakes-cpr: cool cool, checked with alan+ peter and updated it 
Screenshot 2024-08-19 at 12 00 28

PatFawbertMills (approved):
Good work

PR #206, opened 2024-08-19
Feature/pdct 1359 add worldmap to the mcf home page

odrakes-cpr:
# What's changed

Added world map to the mcf home page

Screenshot 2024-08-19 at 11 50 44

This is just the first iteration of the world map on this landing page, we will update the data and the legend for the mcf docs in a future PR

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


katybaulch (changes requested):
Looks great! Can we split the methodology change and the world map change into two separate PRs though please (as they're for different tickets) for traceability 🙏🏻

katybaulch (approved):
nice one!!

PR #212, opened 2024-09-09
SEO improvements

PatFawbertMills:
# What's changed
4 linear tickets focused on improving the SEO of the application:
- Update title and description for geography and family pages - PDCT-1334
- Ensure family summary is visible to search engines - PDCT-1329
- Set canonical URL for every page without the query string (or hash in future) - PDCT-1330
- Remove `nofollow` from external links - PDCT-1331

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/src/pages/document/[id].tsx
+++ b/src/pages/document/[id].tsx
@@ -35,8 +35,9 @@ import { getMainDocuments } from "@helpers/getMainDocuments";
 
 import { sortFilterTargets } from "@utils/sortFilterTargets";
 import { pluralise } from "@utils/pluralise";
+import { getCanonicalUrl } from "@utils/getCanonicalUrl";
 
-import { TFamilyPage, TMatchedFamily, TTarget, TGeographySummary } from "@types";
+import { TFamilyPage, TMatchedFamily, TTarget, TGeographySummary, TTheme } from "@types";
 
 import { QUERY_PARAMS } from "@constants/queryParams";
 import { EXAMPLE_SEARCHES } from "@constants/exampleSearches";
@@ -154,7 +155,7 @@ const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({ pa
   };
 
   return (
-    <Layout title={`${page.title}`} description={page.summary.substring(0, 164)}>
+    <Layout title={`${page.title}`} description={page.summary.substring(0, 164)} canonical={getCanonicalUrl(router, process.env.THEME as TTheme)}>
       <Script id="analytics">
         analytics.category = "{page.category}"; analytics.type = "{getDocumentCategories().join(",")}"; analytics.geography = "{page.geography}";
       </Script>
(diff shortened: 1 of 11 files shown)


src/pages/document/[id].tsx
```diff
@@ -35,18 +34,21 @@ import { getMainDocuments } from "@helpers/getMainDocuments";
...
 import { EXAMPLE_SEARCHES } from "@constants/exampleSearches";
 import { MAX_FAMILY_SUMMARY_LENGTH } from "@constants/document";
 import { MAX_PASSAGES } from "@constants/paging";
+import { extractNestedData } from "@utils/extractNestedData";
 
 type TProps = {
   page: TFamilyPage;
   targets: TTarget[];
-  geographySummary: TGeographySummary;
+  countries: TGeography[];
```
Comment (katybaulch) [inline src/pages/document/[id].tsx:50]: Do we want to stick with countries or use `geographies`?

src/pages/document/[id].tsx
```diff
@@ -56,7 +58,7 @@ type TProps = {
   - The 'physical document' view is within the folder: src/pages/documents/[id].tsx.
 */
 
-const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({ page, targets = [] }: TProps) => {
+const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({ page, targets = [], countries = [], organisations = null }: TProps) => {
```
Comment (katybaulch) [inline src/pages/document/[id].tsx:61]: See above.

src/pages/document/[id].tsx
```diff
@@ -357,6 +359,8 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
 
   let familyData: TFamilyPage;
   let targetsData: TTarget[] = [];
+  let countriesData: TGeography[] = [];
```
Comment (katybaulch) [inline src/pages/document/[id].tsx:362]: Same as above

src/pages/document/[id].tsx
Comment (PatFawbertMills) [reply src/pages/document/[id].tsx:50]: As its doing the same as the `useConfig` hook I've kept the naming consistent for now. I think the main reason we did that originally was to differentiate between the regions and countries

Comment (katybaulch) [reply src/pages/document/[id].tsx:50]: Okie dokie :)

katybaulch (approved):
Nice work!!

src/utils/getThemeDomain.ts
```diff
@@ -1,5 +1,6 @@
 import { TTheme } from "@types";
-const CPR_DOMAIN = "climatepolicyradar.org";
+
+const CPR_DOMAIN = "app.climatepolicyradar.org";
```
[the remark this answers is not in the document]
katybaulch: Feel free to resolve if we're gonna add the MCF domain here later :)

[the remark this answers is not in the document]
PatFawbertMills: We can add the MCF domain when we get to it, but we don't need to follow the CPR convention of putting the app on a sub domain. CCLW's is on its own domain

PR #215, opened 2024-09-16
Bug - Analytics not consenting

PatFawbertMills:
# What's changed
Analytics were not being tracked using google tag manager. I noticed that where we instantiate our `gtag` method, we did not appear to be updating the 'consent' values.

The confusing thing here is that this previously was working. I have redeclared the gtag as a `Function` and this now appears to be passing the consent update as a method call with arguments.

I've checked this through the google tag manager console and can confirm the consent is being updated correctly.

## Proposed version

- [ ] Skip auto-tagging
- [ ] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


katybaulch (approved):
nice tidy ❇️

PR #223, opened 2024-10-10
Feature/pdct 1563 create header and footer for mcf landing page

odrakes-cpr:
# What's changed

## Proposed version

Add Header and Footer Components to MCF Landing Pages

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

Links: 

Figma designs : 

https://www.figma.com/design/Pejl2RZSC3n5FOTBm2Qt23/MCF-custom-app----initial-wireframes?node-id=915-296&node-type=frame&t=s8m8ndSPTBHvM5vE-0


odrakes-cpr:
Footer

Screenshot 2024-10-10 at 13 08 40


Header
Screenshot 2024-10-10 at 13 08 33

src/components/svg/Icons.tsx
```diff
@@ -661,6 +661,90 @@ export const ManyDocumentsIcon = ({ height = "24", width = "24", color = "curren
...
 
+export const FacebookIcon = ({ height = "20", width = "24", color = "#0171B8" }: IconProps) => {
+  return (
+    <svg
+      aria-label="Facebook Icon"
+      role="img"
+      xmlns="http://www.w3.org/2000/svg"
+      width={width}
+      height={height}
+      viewBox={`0 0 ${height} ${width}`}
```
PatFawbertMills, line 672: don't change the view box as this is the SVG's internal sizing - it isn't the size of the SVG itself, that is controlled by the width and height properties

src/components/svg/Icons.tsx
```diff
@@ -661,6 +661,90 @@ export const ManyDocumentsIcon = ({ height = "24", width = "24", color = "curren
   );
 };
 
+export const FacebookIcon = ({ height = "20", width = "24", color = "#0171B8" }: IconProps) => {
```
PatFawbertMills, line 664: Use `"currentColor"` rather than explicitly using a hex code otherwise if we ever want to change the brand colours we will need to manually go through each icon.

The colour can then be set on the container, which should permeate into the icon

themes/mcf/tailwind.config.js
```diff
@@ -89,6 +92,9 @@ module.exports = {
       },
       letterSpacing: {
         slight: "0.015em",
+        normalHeading: "-0.01em",
+        xtraLargeHeading: "-0.04em",
+        subheading: "0.04em",
```
PatFawbertMills, line 97: Nice!

themes/mcf/components/HeaderComponents/Menu.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const [showMenu, setShowMenu] = useState(false);
+  const menuRef = useRef(null);
+  useOutsideAlerter(menuRef, () => setShowMenu(false));
+
+  const toggleMenu = useCallback((e) => {
+    e.preventDefault();
+    setShowMenu((prev) => !prev);
+  }, []);
+
+  const mainAppColour = "#082C43";
```
PatFawbertMills, line 20: Can we have this somewhere more reusable?

themes/mcf/components/FooterComponents/SocialMediaContent.tsx
```diff
@@ -0,0 +1,60 @@
+import React from "react";
+import { ExternalLink } from "@components/ExternalLink";
+import { FacebookIcon, InstagramIcon, LinkedInIcon, TwitterIcon } from "@components/svg/Icons";
+
+const SocialMediaContent = () => {
+  const multilateralClimateFundSocialMedia = {
```
PatFawbertMills, line 6: define this in a const file and import it, otherwise right here it will define the object on every re-render

themes/mcf/components/Footer.tsx
```diff
@@ -0,0 +1,62 @@
...
+import SocialMediaContent from "./FooterComponents/SocialMediaContent";
+
+const Footer = () => {
+  return (
+    <footer className="flex flex-col bg-grey-400">
+      <div className="py-12">
+        <SiteWidth>
+          <div className="flex flex-col md:flex-row gap-4">
+            <div className="footer__section w-full md:w-1/2 lg:w-2/3" data-cy="footer-mcf">
+              <h2 className="font-greycliff text-2xl font-bold leading-[24px] tracking-[-0.01em] text-left text-headingText text-headingText">
```
PatFawbertMills, line 14: Use new `Heading` component here, also I've not looked into the new font typeface but if the tracking and letter spacing is consistent with whenever we use it we will want to bake it into the heading component / or css rather than define on the fly like this

PatFawbertMills (changes requested):
Looking great, few things I've commented on which would be good to look at

themes/mcf/tailwind.config.js
PatFawbertMills: Combine these new settings with the `Heading` mentioned in the last comment below

themes/mcf/constants/colors.ts
```diff
@@ -0,0 +1,3 @@
+export const mainAppColour = "#082C43";
```
PatFawbertMills, line 1: Let's have a catch up, we should try to keep colours defined within the tailwind config

themes/mcf/styles/styles.scss
```diff
@@ -59,3 +67,31 @@ option {
...
+  }
+  h5 {
+    @apply mb-4;
+  }
+}
+
+.custom-header,
+h1.custom-header,
+h2.custom-header {
+  font-family: "Greycliff", sans-serif;
```
PatFawbertMills, line 83: Let's have a catch up on how we can integrate the tailwind config values into our css here

themes/mcf/pages/homepage.tsx
```diff
@@ -15,15 +19,23 @@ type TProps = {
...
-    <main id="main" className="relative h-full">
-      <div className="bg-cclw-dark text-white">
-        <Hero handleSearchInput={handleSearchInput} searchInput={searchInput} />
-      </div>
-      <div id="world-map" className="container hidden pt-6 md:block">
-        <h2 className="text-center mb-6">Explore by country</h2>
-        <WorldMap />
-      </div>
-    </main>
+    <Layout title="Law and Policy Search">
```
PatFawbertMills, line 22: Probably bets to mention projects rather than Laws and Policies, something like "Multilateral climate fund search"

PatFawbertMills (approved):
Good job mate 👍

PR #228, opened 2024-10-16
Feature/pdct 1592 use new geographies instead of geography for fam document

katybaulch:
# What's changed

In the backend [v1.18.1](https://github.com/climatepolicyradar/navigator-backend/releases/tag/v1.18.1-beta) we have added a new `geographies` key that will supersede `geography`. 

This PR is part 1/2 of implementing multiple geographies in the frontend.  This PR switches the frontend to use the new `geographies` reference (so we can remove `geography` in the backend), and the second part will be to implement the visual multi-geographies for CountryLinks.

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


PatFawbertMills (approved):
Think we should be fine as a temporary fix for multi-geos

PR #230, opened 2024-10-16
PDCT 1460 - filtering form components

PatFawbertMills:
# What's changed
- Rebuild of filters on search
- Componentisation of common form elements such as inputs and checkboxes
- Removal of internationalisation
- Adding framer-motion for UI animations

## Proposed version

- [ ] Skip auto-tagging
- [ ] Patch
- [x] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


PatFawbertMills:
Screen Recording 2024-10-16 at 15 51 12

odrakes-cpr (approved):
tested locally across all three apps, and works well.

PatFawbertMills:
I've noticed the search options are not available when on the mobile view, so I will add that now.

Also picked up an issue with the breadcrumbs on mobile view when on the search page due to an oversight in how the download links sit next to the breadcrumbs.

PR #231, opened 2024-10-17
Feature/pdct 1597 update search result list item familylistitem to include the

odrakes-cpr:
# What's changed

Update Metadata tiles on the search page to include MCF corporas.  

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/src/helpers/getCategoryName.ts
+++ b/src/helpers/getCategoryName.ts
@@ -1,6 +1,6 @@
-import { TCategory } from "@types";
+import { TCategory, TCorpusTypeSubCategory } from "@types";
 
-export const getCategoryName = (category: TCategory) => {
+export const getCategoryName = (category: TCategory, subCategory?: TCorpusTypeSubCategory) => {
   let name = "";
   switch (category) {
     case "Litigation":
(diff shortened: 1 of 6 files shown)


src/helpers/getCategoryName.ts
```diff
@@ -17,6 +17,9 @@ export const getCategoryName = (category: TCategory) => {
     case "UNFCCC":
       name = "UNFCCC";
       break;
+    case "MCF":
+      name = subCategory ? subCategory.toUpperCase() : "MCF";
```
PatFawbertMills, line 21: One to catch up with Conor/Dominyka later but I have a feeling they will want the full name. Can easily circle back later though

PatFawbertMills (approved):
LGTM

src/helpers/getCategoryName.ts
odrakes-cpr: I will confirm in application

src/helpers/getCategoryName.ts
```diff
@@ -1,22 +1,29 @@
...
-    case "Executive":
-    case "Policy":
-      name = "Policy";
-      break;
-    case "UNFCCC":
-      name = "UNFCCC";
-      break;
-  }
-  return name;
+const subCategories: Record<TCorpusTypeSubCategory, string> = {
```
PatFawbertMills, line 3: nice

PR #234, opened 2024-10-23
Feature/pdct 1571 render multiple geographies in frontend

katybaulch:
# What's changed

Render multiple geographies in the frontend.

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/src/components/CountriesLink.tsx
+++ b/src/components/CountriesLink.tsx
@@ -1,0 +1,26 @@
+import { CountryLink } from "@components/CountryLink";
+import { getCountryName } from "@helpers/getCountryFields";
+import { isSystemGeo } from "@utils/isSystemGeo";
+import { TGeography } from "@types";
+
+type TProps = {
+  geographies: string[];
+  countries: TGeography[];
+};
+
+const CountriesLinks = ({ geographies, countries }: TProps) => (
+  <>
+    {geographies?.map(
+      (geography, index) =>
+        !isSystemGeo(geography) && (
+          <span key={index} className="flex gap-1">
+            <CountryLink countryCode={geography} className="text-textDark no-underline">
+              <span>{getCountryName(geography, countries)}</span>
+            </CountryLink>
+          </span>
+        )
+    )}
+  </>
+);
+
+export default CountriesLinks;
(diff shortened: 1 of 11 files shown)


src/components/CountriesLink.tsx
```diff
@@ -0,0 +1,25 @@
...
+  geographies: string[];
+  countries: TGeography[];
+};
+
+export const CountriesLink: FC<TCountriesLink> = ({ geographies, countries }) => (
+  <>
+    {geographies?.map(
+      (geography, index) =>
+        !isSystemGeo(geography) && (
+          <span key={index} className="flex gap-1">
```
PatFawbertMills, line 17: Use the geo as the key, index as key should only be for situations where we cannot be sure the values in the collection will be distinct.

src/components/CountriesLink.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { getCountryName } from "@helpers/getCountryFields";
+import { isSystemGeo } from "@utils/isSystemGeo";
+import { TGeography } from "@types";
+
+type TCountriesLink = {
+  geographies: string[];
+  countries: TGeography[];
+};
+
+export const CountriesLink: FC<TCountriesLink> = ({ geographies, countries }) => (
```
PatFawbertMills, line 12: This component doesn't need to be a FunctionalComponent (and I think this is a legacy feature of react now anyway, reserved for when we need to use `children` as a prop)

PatFawbertMills (changes requested):
Couple of v minor react-y things

src/components/CountriesLink.tsx
katybaulch: Ahh okay - my bad blindly copying from other parts of the codebase! Thanks for letting me know. Think I've updated accordinly - but let me know if I've still got it wrong

katybaulch:
Driveby fix for a bug where XAB geographies are clickable in the breadcrumb
image_720

PatFawbertMills (approved):
LGTM

katybaulch:
> what happens in the breadcrumb when there are multiple geos? I can see in the code here that it looks like we don’t display anything?
> 
> ```typescript
> const breadcrumbGeography = family.geographies && family.geographies.length > 1 ? null : { label: geoName, href: `/geographies/${geoSlug}` };
> ```

it sets the breadcrumbgeo value to null, and it shows this (which is consistent with Conor's designs)

image_720

PR #235, opened 2024-10-23
Feature/pdct 1603 render mcf metadata on family projects page

odrakes-cpr:
# What's changed

Update Projects/Family Page with an addition of a metadata render which renders the metadata items for the project depending on the fund/corpus.

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


odrakes-cpr:
Screenshot 2024-10-23 at 14 58 26

Tested with multi geos

Screenshot 2024-10-23 at 13 38 19

odrakes-cpr:
Updated Order 

Screenshot 2024-10-23 at 15 55 25

src/components/document/FamilyMeta.tsx
```diff
@@ -22,7 +22,7 @@ export const FamilyMeta = ({ category, date, geographies, topics, author, corpus
   return (
     <>
       <CountriesLink geographies={geographies} countries={countries} />
-      {!isNaN(year) && <span data-cy="family-metadata-year">Approval FY: {year}</span>}
+      {!isNaN(year) && <span data-cy="family-metadata-year">{`${category === "MCF" ? "Approval FY: " + year : year}`}</span>}
```
PatFawbertMills, line 25: Can you add a TODO comment that we will need to circle back and look at this later once we get the config more established. Because we won't be able to have conditional checks like this in the near future.

src/components/document/McfFamilyMeta.tsx
```diff
@@ -0,0 +1,79 @@
...
+const MetadataItem = ({ label, icon, value }: MetadataItemProps) => {
+  const isUrl = label === "Source";
+  const isCountry = label === "Geography";
+  const IconComponent = getIcon(icon);
+
+  const getValueContent = () => {
+    if (isUrl && typeof value === "string") {
+      return (
+        <ExternalLink url={value} className="text-blue-600 underline truncate text-sm">
+          {" "}
```
PatFawbertMills, line 43: Random jsx space here

src/helpers/mapFamilyMetadata.ts
```diff
@@ -0,0 +1,42 @@
...
+        });
+      } else if (key.includes("project_value") && values[0] !== "0") {
+        result.push({ label: mapping.label, iconLabel: mapping.iconLabel, value: getSumUSD(values as string[]) });
+      } else if (key === "organisation") {
+        result.push({ label: mapping.label, iconLabel: mapping.iconLabel, value: getSubCategoryName(values as TCorpusTypeSubCategory) });
+      } else {
+        result.push({
+          label: mapping.label,
+          iconLabel: mapping.iconLabel,
+          value: (values as string[]).slice(0, 3).join(", "),
```
PatFawbertMills: Can you make `3` a constant in this file with a descriptive name? it would make this a little more readable and easier to update later if we want to display more

src/components/CountryLink.tsx
```diff
@@ -9,9 +9,18 @@ type TCountryLink = {
...
 };
 
-export const CountryLink: FC<TCountryLink> = ({ countryCode, className = "", emptyContentFallback, children }) => {
+export const CountryLink: FC<TCountryLink> = ({
+  countryCode,
+  className = "",
+  emptyContentFallback,
+  children,
+  showFlag = true,
+  showSlugAsChildren = false,
```
PatFawbertMills, line 22: What instances do we need this?

PatFawbertMills (changes requested):
Few questions and requests

src/components/CountryLink.tsx
odrakes-cpr: currently the country link renders an empty string if a children prop is not provided, as we want to actually show the slug value as well as create a link to the geo page we need to display that slug value aswell

odrakes-cpr:
Handle cases of multi geogrpahies overflowing and coming out of the div.

Screenshot 2024-10-24 at 14 11 14

src/components/CountryLink.tsx
PatFawbertMills: In that case can't we just pass in the slug as part of the children of the given link?

My worry with this is we are making the component be concerned with context

src/components/document/McfFamilyMeta.tsx
```diff
@@ -0,0 +1,78 @@
...
+const getCountryLinks = (countryCodes: string[]): JSX.Element[] => {
+  const countries = countryCodes.map((code, index) => (
+    <div className="flex" key={code}>
+      <CountryLink
+        countryCode={code}
+        showFlag={false}
+        showSlugAsChildren={true}
+        className="text-blue-600 underline truncate text-sm text-transform: capitalize"
+      />
+      {index !== countryCodes.length - 1 && <span>,</span>}
```
PatFawbertMills, line 28: To follow on from the convo above regarding rendering the slug, we can pass the slug into the component here.

Also - we should not include the comma `,` as part of the link

PatFawbertMills (changes requested):
Couple of comments regarding keeping business logic context out of the `CountryLink` component

src/components/document/McfFamilyMeta.tsx
PatFawbertMills: Ignore this, comma isn't inside

PR #237, opened 2024-10-24
feature/pdct-1619-update-geography-page

odrakes-cpr:
# What's changed

- As we want to keep the functionality and look of the geography page at this stage, we are just updating the way we present the data
- filter the data in the server side prop only return MCF related data
- define variable to hold if the app is currently mcf, if so update the document categories that we pass to the nav index, as well as the docs that we are rendering in the render docs file. For the latter we we only parse the current category if there are more than one categories/families, in which case is a non mcf app

## Proposed version

- [ ] Skip auto-tagging
- [ ] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/src/constants/documentCategories.ts
+++ b/src/constants/documentCategories.ts
@@ -1,3 +1,6 @@
 export const DOCUMENT_CATEGORIES = ["All", "Legislation", "Policies", "UNFCCC", "Litigation"];
 
+// For now all individual fund documents are returning as MCF, so we do not need separate categories
+export const MCF_DOCUMENT_CATEGORIES = ["All"];
+
 export type TDocumentCategory = (typeof DOCUMENT_CATEGORIES)[number];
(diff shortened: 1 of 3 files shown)


odrakes-cpr:
Screenshot 2024-10-24 at 19 03 00
Screenshot 2024-10-24 at 19 03 36

odrakes-cpr:
No results

Screenshot 2024-10-28 at 11 16 36

PatFawbertMills (approved):
LGTM. One to come back and sort out post-COP so that we can move away from the `theme` checks in the code.

PatFawbertMills (changes requested):
We need to hide legislative process I've just realised

katybaulch:
> sort out post-COP so that we can move away from the `theme` checks in the code.

https://linear.app/climate-policy-radar/issue/PDCT-1621/frontend-remove-theme-check-business-logic-from-code

katybaulch (approved):
lgtm as Patrick's comment has been addressed

PatFawbertMills (approved):
n1

PR #238, opened 2024-10-25
Feature/pdct 1602 create metadata renderer component for document page

odrakes-cpr:
# What's changed

Add document meta renderer which renders new metadata component to the document page if the document is an MCF document and keeps current styling on all other projects. 

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


odrakes-cpr:
Screenshot 2024-10-28 at 11 54 28

src/components/documents/DocumentHead.tsx
```diff
@@ -70,22 +71,10 @@ export const DocumentHead = ({ document, family, handleViewOtherDocsClick, handl
         />
       </SubNav>
       <SiteWidth>
-        <div className="flex flex-col lg:flex-row">
-          <div className="flex-1 my-4">
+        <div className="flex flex-col justify-between lg:flex-row">
+          <div className="flex-1 my-4 max-w-[1100px]">
```
PatFawbertMills, line 75: We shouldn't be setting containers in this way, that was the whole point of the `sr/components/panels` component set.

Can you use the `SingleCol` component?

src/components/documents/renderers/DocumentMetaRenderer.tsx
```diff
@@ -0,0 +1,23 @@
+import { McfFamilyMeta } from "@components/document/McfFamilyMeta";
+import { DocumentMeta } from "../DocumentMeta";
+import { getApprovedYearFromEvents } from "@helpers/getApprovedYearFromEvents";
+
+const MultilateralClimateFunds = "MCF";
```
PatFawbertMills, line 5: Make constants descriptive and in all uppercase so that they are easier to spot in the code please

PatFawbertMills (changes requested):
Couple of things

odrakes-cpr:
Handle multiple values for implementing agencies and geographies


Screenshot 2024-10-28 at 17 50 17

PatFawbertMills (approved):
Nice work

PR #239, opened 2024-10-28
Search config integration

PatFawbertMills:
# What's changed
### MAJOR UPDATE: v2.0.0-beta: breaking changes
- New theme config file
- Config mostly contains filtering options, find it under `/theme/[theme]/config.json`
- Rebuild of the search filters
- Integration with the new theme config in `search`, `searchFilters`, `useSearch`, `buildSearchQuery`

## Proposed version

- [ ] Skip auto-tagging
- [ ] Patch
- [ ] Minor version
- [x] Major version

Notes:

  brackets and the `x`


src/api/http-common.ts
```diff
@@ -5,6 +5,10 @@ export async function getEnvFromServer() {
   return await axios.get("/api/env").then((res: any) => res);
 }
 
+export async function getFilters() {
+  return await axios.get("/api/filters").then((res: any) => res);
+}
+
```
odrakes-cpr, line 11: All looks good, one thing - I am not sure what this endpoint does ?

odrakes-cpr (approved):

src/api/http-common.ts
PatFawbertMills: It is defined in the `pages` directory, under `/src/pages/api/filters.ts`. In Next you can build API routes: https://nextjs.org/docs/pages/building-your-application/routing/api-routes

katybaulch (approved):
Nice one! Think you might need to merge main into this potentially - in your demo multiple geos weren't rendered and that is on staging atm :)

src/hooks/usePDFPreview.ts
```diff
@@ -74,7 +74,10 @@ export default function usePDFPreview(document: TDocumentPage, documentPassageMa
   const createPDFClient = (startingPassage: number, onLoadCallback?: Function) => {
     viewSDKClient = new ViewSDKClient();
     viewSDKClient.ready().then(() => {
-      const previewFilePromise = viewSDKClient.previewFile(document, adobeKey, "pdf-div", viewerConfig);
+      // Prevent attaching preview to non-existent div
+      const pdfDiv = document.getElementById("pdf-div");
```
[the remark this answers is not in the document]
PatFawbertMills: No this was to prevent the adobe package trying to attach to a div that was not yet rendered.

jamesgorrie (commented):
This comment is more to help me understand the stack we have, and where we commonly do transformations, given I am very unfamiliar with the stack, so excuse any presumptions.

## Transformation

It looks like we transform from `config` => `rendered filter` => `qs params` => `search API params`. 

I am trying to understand why we do these transformations?

When we had the run through (thanks!) - we spoke about how we add new params for new functionality. It made me wonder what it might look like if we remove those transformations, could it remove the need to make these decisions in future?

An example of this might be that the config is (hyper reduced pseudo code):
```JSONC
{
  "categories": {
    "options": {
      "label": "Intl. agreements",
      "query": { /** This would match the API calls */ }
    }
  }
}
```

Then when a filter is clicked, we append the filter to the QS of the app which is then just forwarded to the API.

I know this is walking around some obvious decisions that have been made, so it's a more of a question to try and surface those than a suggestion.


## When to use API calls vs config in the app

We seem to load most data in through API calls. Do we ever load in static config (stuff that never changes) when the app loads on the server rather than via HTTP which comes with a small latency hit?

PatFawbertMills:
Hopefully this answers and explains some of the thinking:

There are multiple challenges that I am trying to solve with this approach:
- Displaying the correct filter values to drive the respective theme (this recent change)
- Build a search payload object
- Save (using the URL params) a recoverable search state. Also enables link sharing, etc
- Working with the way pagination (continuation tokens) works with Vespa

I think it would be really valuable for us to do a deep-dive on this because I have no doubt there are improvements we could make. Hopefully we can simplify the whole thing.

PR #240, opened 2024-10-28
fix: avoid LegacyKeyValueFormat in Docker by using =

jamesgorrie:
# What's changed

I had received a `LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 22)` error when running `make`.

This resolved that.

This is more being done to help me, as a new starter, to verify I have everything that I need to make a change and deploy - rather than being excessively useful change.

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ COPY --chmod=0755  . .
 # Install dependencies.
 RUN npm ci && npm cache clean --force
 
-ENV PATH /home/node/app/node_modules/.bin:$PATH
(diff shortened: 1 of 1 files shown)


jamesgorrie:
@katybaulch - yup, I was trying to work out what change was too small a change, but this is good to go.

katybaulch (approved):
nice spot!

PR #249, opened 2024-10-30
Feature/pdct 1565 add mcf landing page content components

odrakes-cpr:
# What's changed

UI fixes 
Additional Homepage Content - Contextual Search/ About CPE
Adding Methodology Page
Adding Terms of Use Page

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


odrakes-cpr:
Screenshot 2024-10-31 at 08 18 39

Screenshot 2024-10-31 at 08 58 00
Screenshot 2024-10-31 at 08 58 07

themes/mcf/pages/methodology.tsx
```diff
@@ -1,19 +1,194 @@
...
+                  project status list across all the funds (please see the list below):
+                </p>
+                <ul>
+                  <li>Concept Approved</li>
+                  <li>Project Approved</li>
+                  <li>Under Implementation</li>
+                  <li>Project Completed</li>
+                  <li>Cancelled</li>
+                </ul>
+                <p>This table shows how standardised project status relates to each fund project status.</p>
```
annaCPR, line 40: You're gonna have fun with this table when AF is added (eventually) 😆

src/pages/_app.tsx
```diff
@@ -55,6 +55,7 @@ function MyApp({ Component, pageProps, theme, adobeApiKey }: TProps) {
           <ErrorBoundary level="top">
             <Head>
               <link rel="icon" href={favicon} />
+              <link rel="stylesheet" href="https://use.typekit.net/ifw7ubm.css" />
```
PatFawbertMills, line 58: Let's catch up on this - this will load the font into every app. Something I would like to avoid if at all possible

themes/mcf/tailwind.config.js
```diff
@@ -100,7 +100,7 @@ module.exports = {
         maxContent: MAX_CONTENT_WIDTH,
       },
       fontFamily: {
-        greycliff: ["Greycliff", "sans-serif"],
+        greycliff: ["greycliff-cf", "sans-serif"],
```
PatFawbertMills, line 103: Suggest we speak to Conor about the fallback options

themes/mcf/pages/methodology.tsx
```diff
@@ -1,19 +1,194 @@
...
+                What documents are included
+              </Heading>
+              <p className="text-content mb-12">
+                The platform contains publicly available policies and project documents (such as concept notes, full proposals, implementation
+                reports) of the four Multilateral Climate Funds.
+              </p>
+              <Heading level={2} extraClasses="custom-header">
+                Definitions and classifications that are currently in use
+              </Heading>
+              <div className="text-content mb-12">
```
PatFawbertMills, line 28: Rather than putting `text-content` onto every div we can just add it onto line 17 div once

themes/mcf/pages/about.tsx
```diff
@@ -18,24 +18,24 @@ const About = () => {
...
         <SiteWidth>
           <SingleCol>
-            <div className="text-content mb-12">
-              <Heading level={1} extraClasses="custom-header">
+            <div>
+              <Heading level={1} extraClasses="custom-header mb-4">
                 About
               </Heading>
-              <p>
+              <p className="text-content mb-12">
```
PatFawbertMills, line 25: Same comment as below, put `text-content` on wrapping div on line 21 rather than repeating throughout

themes/mcf/components/AboutTheFunds.tsx
```diff
@@ -46,15 +45,13 @@ const AboutTheFundsContent = () => {
...
-            </ExternalLink>
-          </p>
+          <LinkWithQuery className="underline" href="/about">
+            Find out more
+          </LinkWithQuery>
         </div>
         <div className="hidden xl:block xl:h-full">
           {/* eslint-disable-next-line @next/next/no-img-element */}
-          <img src="/images/mcf/mcf-fund-logos.png" alt="Multilateral Climate Fund Logos" className="h-[34rem]" />
+          <img src="/images/mcf/mcf-fund-logos.png" alt="Multilateral Climate Fund Logos" className="h-[28 rem]" />
```
PatFawbertMills, line 54: We are using being in pixels below (see CPR image) and rem here. Personally I am more in favour of pixels as it is a bit more semantically readable to see what's going on. Happy to be persuaded though.

PatFawbertMills (changes requested):
Nice, looking good. Few questions and suggestions.

src/pages/_app.tsx
PatFawbertMills: If we got to this URL there is some CSS we could add into the MCF `.scss` file in the theme directory:
https://use.typekit.net/ifw7ubm.css

Then we can remove this from the app head.

themes/mcf/pages/methodology.tsx
PatFawbertMills: Have had a chat and it makes sense!

themes/mcf/pages/about.tsx
PatFawbertMills: As below discussed and makes sense!

PatFawbertMills (approved):
🚢 lets ship it

PR #250, opened 2024-10-30
Updates from Conor

PatFawbertMills:
# What's changed
- Lots of minor UI updates based on the chat through with Conor

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/src/components/documents/DocumentHead.tsx
+++ b/src/components/documents/DocumentHead.tsx
@@ -67,7 +67,7 @@ export const DocumentHead = ({ document, family, handleViewOtherDocsClick, handl
           geography={breadcrumbGeography}
           category={breadcrumbCategory}
           family={breadcrumbFamily}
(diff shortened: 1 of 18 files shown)


src/components/documents/DocumentHead.tsx
```diff
@@ -67,7 +67,7 @@ export const DocumentHead = ({ document, family, handleViewOtherDocsClick, handl
           geography={breadcrumbGeography}
           category={breadcrumbCategory}
           family={breadcrumbFamily}
-          label={<span className="capitalize">{breadcrumbLabel}</span>}
+          label={breadcrumbLabel ? <span className="capitalize">{breadcrumbLabel}</span> : document.title}
```
odrakes-cpr, line 70: nice

odrakes-cpr (approved):

PR #251, opened 2024-10-31
Feature/pdct 1625 UI adjustments

odrakes-cpr:
# What's changed

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/src/pages/geographies/[id].tsx
+++ b/src/pages/geographies/[id].tsx
@@ -385,24 +385,19 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
     };
   }
 
-  const filterSummaryData = (summaryInformation) => {
+  const filterSummaryData = (summaryInformation: TGeographySummary) => {
     if (isMCFTheme) {
-      const mcfSummaryData = {
+      return {
         family_counts: { MCF: summaryInformation.family_counts.MCF },
         top_families: { MCF: summaryInformation.top_families.MCF },
         targets: [],
       };
-      return mcfSummaryData;
     } else {
-      const { MCF, ...family_counts } = summaryInformation.family_counts;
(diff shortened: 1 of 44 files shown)


src/pages/geographies/[id].tsx
```diff
@@ -393,19 +397,24 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
...
       };
     } else {
+      const { MCF, ...familyCountsWithoutMCF } = summaryInformation.family_counts;
+      const { MCF: mcfTopFamilies, ...topFamiliesWithoutMCF } = summaryInformation.top_families;
       return {
-        family_counts: summaryInformation.family_counts,
-        top_families: summaryInformation.top_families,
+        family_counts: familyCountsWithoutMCF,
+        top_families: topFamiliesWithoutMCF,
         targets: summaryInformation.targets,
```
Comment (odrakes-cpr) [inline src/pages/geographies/[id].tsx:405]: This was throwing off the numbers in the CCLW and CPR app because MCF was still beign passed
see example
Screenshot 2024-10-31 at 12 01 07

src/pages/geographies/[id].tsx
```diff
@@ -393,19 +397,24 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
...
       };
     }
```
Comment (odrakes-cpr) [inline src/pages/geographies/[id].tsx:407]: Fix -
Screenshot 2024-10-31 at 12 03 39

src/pages/geographies/[id].tsx
Comment (odrakes-cpr) [reply src/pages/geographies/[id].tsx:405]: that extra 1 is the one MCF doc

PatFawbertMills (approved):
Good catch

PR #254, opened 2024-10-31
Add Adaptation Fund to MCF config for search

annaCPR:
# What's changed
- added Adaptation Fund to MCF config

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/themes/cpr/config.json
+++ b/themes/cpr/config.json
@@ -66,7 +66,7 @@
       "taxonomyKey": "status",
       "apiMetaDataKey": "family.status",
       "type": "radio",
-      "category": ["MCF.corpus.GCF.n0000", "MCF.corpus.GEF.n0000"]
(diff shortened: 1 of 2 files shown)


themes/cpr/config.json
PatFawbertMills, line 1: @annaCPR we need to also add it to the array `category` in the filters. If we don't when a user selects the fund Adaptation Fund, they won't see the MCF filters

PatFawbertMills: Sorry "it" being the AF corpus ID

PatFawbertMills: The reason we don't need to in the mcf.config is because the category array can be left empty which translates to "always displaying" in the underlying logic for the filters

PatFawbertMills (changes requested):
Regarding AF corpus ID

PatFawbertMills (approved):
LGTM

PR #255, opened 2024-11-04
PDCT-1645 Add MCF favicon

jamesgorrie:
# What's changed
Adds the MCF favicon supplied [here](https://linear.app/climate-policy-radar/issue/PDCT-1645/favicons-for-cpe).

The favicon - I think - is supplied for Climate Policy Explorer. Not sure if this is something we should clarify going forward by changing the theme to CPE. 

**Localhost tab screenshot**
Screenshot 2024-11-04 at 11 38 00

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version


katybaulch (approved):
looks good to me (lgtm)!! I think we want to keep this as MCF for now for consistency, but we should look at potentially updating this after COP

PR #256, opened 2024-11-04
Feature/pdct 1646 fix mcf page description so it isnt a duplicate of cprs

katybaulch:
# What's changed

Fix page description and app name for MCF & refactor to use theme constants.

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


themes/cclw/constants/pageMetadata.ts
jamesgorrie, line 1: Would keeping this in [the theme config](https://github.com/climatepolicyradar/navigator-frontend/tree/main/themes) help with having fewer places to know to look for values that alternate on `theme`?

odrakes-cpr: @jamesgorrie do you mean [here](https://github.com/climatepolicyradar/navigator-frontend/blob/main/themes/cpr/config.json)

odrakes-cpr (approved):
lgtm

themes/cclw/constants/pageMetadata.ts
PatFawbertMills: Yes I think the new config is a good candidate for "CMS-able data". Part of me wonders if we should tackle that in earnest after COP when we have a minute to think about all the places we are currently using references to the "old way" of doing things. This being one example of that.

PatFawbertMills (approved):
LGTM. I vote we circle back to loading theme content from the config post COP

themes/cclw/constants/pageMetadata.ts
katybaulch: I think at the moment, we're only wanting to load theme specific content for search from the config file. I think there were plans to extend this further after COP, but I'd need to check with Patrick

katybaulch: Ah I didn't realise Patrick had reviewed below -- because of his comment I'm going to raise a separate linear ticket [PDCT-1650](https://linear.app/climate-policy-radar/issue/PDCT-1650/load-theme-specific-content-from-configjson-file) for us to revisit post COP).

https://github.com/climatepolicyradar/navigator-frontend/pull/256#pullrequestreview-2413187446

PR #260, opened 2024-11-05
Do not show "no geography" in the meta detail list of geographies

PatFawbertMills:
# What's changed
- Only display "International" in the list of associated geographies
- Strictly not to display "No Geography" at any point

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/src/constants/systemGeos.ts
+++ b/src/constants/systemGeos.ts
@@ -1,9 +1,8 @@
+export const SYSTEM_NO_GEOGRAPHY_CODE = "xaa";
+export const SYSTEM_NO_GEOGRAPHY_NAME = "no-geography";
 export const SYSTEM_INTERNATIONAL_CODE = "xab";
 export const SYSTEM_INTERNATIONAL_NAME = "international";
 
-export const systemGeoCodes = [
-  "xaa", // No Geography
-  SYSTEM_INTERNATIONAL_CODE,
-];
+export const systemGeoCodes = [SYSTEM_NO_GEOGRAPHY_CODE, SYSTEM_INTERNATIONAL_CODE];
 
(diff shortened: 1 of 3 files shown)


src/constants/systemGeos.ts
```diff
@@ -1,6 +1,9 @@
+export const SYSTEM_INTERNATIONAL_CODE = "xab";
+export const SYSTEM_INTERNATIONAL_NAME = "international";
+
 export const systemGeoCodes = [
   "xaa", // No Geography
-  "xab", // International
+  SYSTEM_INTERNATIONAL_CODE,
```
katybaulch, line 6: Are we happy with this change affecting CPR/CCLW as well?

odrakes-cpr (approved):

src/constants/systemGeos.ts
```diff
@@ -1,6 +1,9 @@
+export const SYSTEM_INTERNATIONAL_CODE = "xab";
+export const SYSTEM_INTERNATIONAL_NAME = "international";
+
 export const systemGeoCodes = [
   "xaa", // No Geography
```
katybaulch, line 5: Why isn't this a const too, if we're pulling XAB out into a const?

PatFawbertMills: Should be, I was being lazy

src/constants/systemGeos.ts
PatFawbertMills: That was the requirement as I understood it. Previously we were not displaying either "International" or "No Geography"

katybaulch: Awesome - just wanted to check!

katybaulch (approved):
lgtm

PR #275, opened 2024-11-11
Fix page title in tab

katybaulch:
# What's changed

Fix page title in tab - I tried to be clever by refactoring this out into theme specific constants, but missed how Search doesn't pass these new constants in. I've reverted the refactor in the core src folder whilst I've remembered which bit of code caused the issue -- but we might want to import these constants from the theme specific folder tomorrow to reduce duplication (or leave as this commit, either work).

# Patrick update
I've retained Katy's original comment above but my comment below:

This turned out to be a lot more involved due to changing the logic to get the meta data content out of the theme config file rather than managing the values for each theme in code.

Hopefully we can use this method going forward and continue to strip business/theme logic and content out of the code completely.

- Config files contain text content for meta title and description
- Some basic logic for replacing `{text}` wildcard in text

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


PatFawbertMills (changes requested):
I've just tested this locally and can still see the CPR title and description.

This is happening because of a combination of reasons:
- We aren't passing any props related to the theme, so the site name is not being used
- We are passing in a description on the search page. The issue is it is not accurate for the MCF site
- We are passing in a description on many of the pages

src/pages/api/filters.ts
jamesgorrie, line 1: What's the reason for this being called `filters`?

src/pages/document/[id].tsx
```diff
@@ -167,7 +167,7 @@ const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({
   };
 
   return (
-    <Layout title={`${page.title}`} description={getFamilyMetaDescription(page.summary, geographyNames?.join(", "), page.category)}>
+    <Layout title={`${page.title}`} description={getFamilyMetaDescription(page.summary, geographyNames?.join(", "), page.category)} appName={theme}>
```
Comment (jamesgorrie) [inline src/pages/document/[id].tsx:170]: I know there might be past reasoning for this, but the swapping between `app` and `theme` gets a bit muddled in my head. 

My understanding of this is that we have multiple apps. Each app might have different styling. Is that the `theme`?

jamesgorrie (approved):
Noice

src/pages/api/filters.ts
PatFawbertMills: Me being a numpty when I first created it as I wrongly assumed I would just use it for the filters, I'm going to change it now...

src/pages/document/[id].tsx
Comment (PatFawbertMills) [reply src/pages/document/[id].tsx:170]: Confusion is fair. I think as an org we have some inconsistency referring to them as "custom apps" and "themes".

Comment (PatFawbertMills) [reply src/pages/document/[id].tsx:170]: But yes your comment is correct. My preference for an explanation of the code is:

This repo is a theme _factory_, that can build apps based upon a selected theme.

PatFawbertMills (approved):

PR #288, opened 2024-11-20
add initial playwright test

jamesgorrie:
# What's changed
- adds an initial playwright e2e test

[🍏 Here's a successful run](https://github.com/climatepolicyradar/navigator-frontend/actions/runs/11957984101/job/33336425619?pr=288)

[🔴 Here's broken run](https://github.com/climatepolicyradar/navigator-frontend/actions/runs/11938826616/job/33277946517)

[📜 Here's some nice docs on how to debug the broken run](https://playwright.dev/docs/ci-intro#html-report)

There is a few areas where this significantly differs from [the current implementation of e2e tests](https://github.com/climatepolicyradar/navigator-frontend/tree/main/e2e). Here's the thinking around it

- **The e2e test are now part of the `src` app.**
  - Yup. This is partially because of co-location reasons. We try to keep the tests as close to the implementation as possible to make them less of an after thought.
  - It's easier to just start the app as they're co-located
  - This does make it less modular though, but haven't seen an opportunity where composition would be useful, so this might be a premature abstraction, but could be missing things
- **Playwright?**
  - after talking to @annaCPR and @PatFawbertMills there did not seem to be any reason we went with Cypress bar that it's there. I have done a few migrations recently from Cypress => Playwright, so I am biased, but those were driven by cypress being a little more flaky, and playwright's ecosystem being far more rich due to the engineering effort going into it (thanks M$)
- **A new actions?** This is to keep this PR clear in it's inception, with the desire to bring it into the [ci/cd pipeline](https://github.com/climatepolicyradar/navigator-frontend/blob/main/.github/workflows/ci-cd.yml). The current e2e tests are also located in the `theme-factory` action, which is coupled to this repo, so we're making that coupling more explicit by keeping the code that tests this code here

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,12 +1,15 @@
-API_URL=http://localhost:8000/api/v1
-S3_URL=https://s3.amazonaws.com
+# This is if you are working exclusively on navigator-frontend
+# To test changes to navigator-backend - swap this out for
+# API_URL=http://localhost:8888/api/v1
+API_URL=https://app.dev.climatepolicyradar.org/api/v1
+S3_URL=https://cpr-staging-targets-json-store.s3.eu-west-1.amazonaws.com
 
 THEME=cpr
 
-ADOBE_API_KEY=1234567890
+ADOBE_API_KEY=dca9187b65294374a6367824df902fdf
 
 REDIRECT_FILE="test.json"
 
-HOSTNAME=https://localhost:3000
+HOSTNAME=http://localhost:3000
 
-NEXT_PUBLIC_APP_TOKEN=some_token
+NEXT_PUBLIC_APP_TOKEN=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhbGxvd2VkX2NvcnBvcmFfaWRzIjpbIkNDTFcuY29ycHVzLmkwMDAwMDAwMS5uMDAwMCIsIkNQUi5jb3JwdXMuaTAwMDAwMDAxLm4wMDAwIiwiVU5GQ0NDLmNvcnB1cy5pMDAwMDAwMDEubjAwMDAiXSwiZXhwIjoyMDQyMTEzMzY5LCJpYXQiOjE3MjY1NzY5NjksImlzcyI6IkNsaW1hdGUgUG9saWN5IFJhZGFyIiwic3ViIjoiQ1BSIiwiYXVkIjoiaHR0cHM6Ly9hcHAuZGV2LmNsaW1hdGVwb2xpY3lyYWRhci5vcmcvIn0.mJ2qLJmMyPLGt0rM_tTXhlVv1glxooxmQV0bWrvPwKU
--- a/src/components/search/SearchResultList.tsx
+++ b/src/components/search/SearchResultList.tsx
@@ -58,14 +58,13 @@ const SearchResultList = ({ category, families, activeFamilyIndex, onClick }: TP
     return <div className="h-96 mt-4 md:mt-0">Your search returned no results.</div>;
   }
   return (
-    <section>
-      <h2 className="sr-only">Search results</h2>
+    <>
       {families?.map((family, index: number) => (
         <div key={index} className={`my-10 ${index === 0 && "md:mt-0"}`} data-cy="search-result">
           <SearchResult family={family} onClick={() => onClick(index)} active={activeFamilyIndex === index} />
         </div>
       ))}
-    </section>
+    </>
   );
 };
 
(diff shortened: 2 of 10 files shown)


tests/search.spec.ts
jamesgorrie, line 1: this is based on [a manual test run we have used previously 🔒](https://www.notion.so/climatepolicyradar/Manual-test-plan-69c302285d284598b7571d97134b7b45)

jamesgorrie:
Here's a successful run.

.env.example
```diff
@@ -1,12 +1,15 @@
...
-ADOBE_API_KEY=1234567890
+ADOBE_API_KEY=dca9187b65294374a6367824df902fdf
 
 REDIRECT_FILE="test.json"
 
-HOSTNAME=https://localhost:3000
+HOSTNAME=http://localhost:3000
 
-NEXT_PUBLIC_APP_TOKEN=some_token
+NEXT_PUBLIC_APP_TOKEN=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhbGxvd2VkX2NvcnBvcmFfaWRzIjpbIkNDTFcuY29ycHVzLmkwMDAwMDAwMS5uMDAwMCIsIkNQUi5jb3JwdXMuaTAwMDAwMDAxLm4wMDAwIiwiVU5GQ0NDLmNvcnB1cy5pMDAwMDAwMDEubjAwMDAiXSwiZXhwIjoyMDQyMTEzMzY5LCJpYXQiOjE3MjY1NzY5NjksImlzcyI6IkNsaW1hdGUgUG9saWN5IFJhZGFyIiwic3ViIjoiQ1BSIiwiYXVkIjoiaHR0cHM6Ly9hcHAuZGV2LmNsaW1hdGVwb2xpY3lyYWRhci5vcmcvIn0.mJ2qLJmMyPLGt0rM_tTXhlVv1glxooxmQV0bWrvPwKU
```
jamesgorrie, line 15: [These are already exposed here](https://github.com/climatepolicyradar/navigator-frontend/blob/main/Makefile#L3-L8).

.env.example
```diff
@@ -1,12 +1,15 @@
...
```
annaCPR, line 15: Are we ok committing staging secrets to version control?

tests/search.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { test, expect } from "@playwright/test";
+
+test("search", async ({ page }) => {
+  /** Reject the consent banner */
+  await page.goto("/");
+  await page.getByRole("button", { name: "Reject" }).click();
```
annaCPR, line 6: Can we add an assertion that the heading of the cookie banner is visible? Just to make it clearer as to what we're rejecting.

.env.example
jamesgorrie: Not 100% sure, [but these are exposed here](https://github.com/climatepolicyradar/navigator-frontend/pull/288#discussion_r1852520087).

src/components/search/SearchResultList.tsx
```diff
@@ -58,13 +58,14 @@ const SearchResultList = ({ category, families, activeFamilyIndex, onClick }: TP
     return <div className="h-96 mt-4 md:mt-0">Your search returned no results.</div>;
   }
   return (
-    <>
+    <div>
+      <h2 className="sr-only">Search results</h2>
```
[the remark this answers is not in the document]
jamesgorrie: Mostly, but also accessibility. [Headings are used by screen-readers to navigate the page structure](https://www.a11yproject.com/posts/how-to-accessible-heading-structure/) - this seems like a natural place a person might want to skip to.

This is often a nice byproduct of these tests - in this case, surfacing that we don't sign post what the proceeding piece of content is defined as. We could probably make better use of sections etc.

I thought defining the [accessibility tree](https://developer.mozilla.org/en-US/docs/Glossary/Accessibility_tree) was outside the scope of this PR - but if we think it important to address here, we can.

.env.example
annaCPR: Ok, didn't know that. Just wanted to make sure we're all aware and agreed on this.

.env.example
annaCPR: Ok, didn't know that. Just wanted to make sure we're all aware and agreed on this.

src/components/search/SearchResultList.tsx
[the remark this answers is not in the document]
annaCPR: 👍  I'm definitely pro using more semantic HTML and making our apps more accessible - there's definitely room for improvement there. Might be worth discussing the topic with the wider group.

PR #290, opened 2024-11-28
Feature/pdct 1385 rename doctype in world map to familycategory or similar

katybaulch:
# What's changed

- Rename docType to famCategory to make it clearer about what the dropdown states represent
- Resize MCF markers
- Inlcude MCF projects in the world map legend

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


src/components/map/WorldMap.tsx
```diff
@@ -123,14 +123,15 @@ export default function MapChart() {
   const [mapCenter, setMapCenter] = useState<TPoint>([0, 0]);
   const [mapZoom, setMapZoom] = useState(1);
   const [showUnifiedEU, setShowUnifiedEU] = useState(false);
-  const [selectedDocType, setSelectedDocType] = useState<"lawsPolicies" | "unfccc">("lawsPolicies");
+  const [selectedFamCategory, setSelectedFamCategory] = useState<"lawsPolicies" | "unfccc" | "mcf">("lawsPolicies");
```
jamesgorrie, line 126: For my clarity - these are corpusTypes (with some gnarly MCF hacking in there?

katybaulch: these are not corpus types, but family categories - the latter of which we want to remove later down the line in favour of 'genre'.

katybaulch: see here: https://github.com/climatepolicyradar/navigator-db-client/blob/500b4f81bb3eeee4111a80e2e33b1e6068eb28fe/db_client/models/dfce/family.py#L19

PatFawbertMills (approved):
Looks good

PR #294, opened 2024-12-04
PDCT-1721 geo mcf tab

PatFawbertMills:
# What's changed
- Adding MCF tab on the geography pages
- Moving the document categories constant into the theme config rather than in code
- This only controls the tabs on the geography page at the moment

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/src/pages/search/index.tsx
+++ b/src/pages/search/index.tsx
@@ -481,7 +481,7 @@ const Search: InferGetServerSidePropsType<typeof getServerSideProps> = ({ theme,
                 <section data-cy="search-results">
                   <h2 className="sr-only">Search results</h2>
                   <SearchResultList
-                    category={router.query[QUERY_PARAMS.category]?.toString() as TDocumentCategory}
+                    category={router.query[QUERY_PARAMS.category]?.toString()}
                     families={families}
                     onClick={handleMatchesButtonClick}
                     activeFamilyIndex={drawerFamily}
--- a/src/pages/search/index.tsx
+++ b/src/pages/search/index.tsx
@@ -38,7 +38,7 @@ import { getThemeConfigLink } from "@utils/getThemeConfigLink";
 
 import { QUERY_PARAMS } from "@constants/queryParams";
 
-import { TTheme, TThemeConfig } from "@types";
+import { TDocumentCategory, TTheme, TThemeConfig } from "@types";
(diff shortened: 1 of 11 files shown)


src/pages/search/index.tsx
```diff
@@ -481,7 +481,7 @@ const Search: InferGetServerSidePropsType<typeof getServerSideProps> = ({ theme,
                 <section data-cy="search-results">
                   <h2 className="sr-only">Search results</h2>
                   <SearchResultList
-                    category={router.query[QUERY_PARAMS.category]?.toString()}
+                    category={router.query[QUERY_PARAMS.category]?.toString() as TDocumentCategory}
```
jamesgorrie, line 484: not blocking - what is the intention of this type casting? I ask as it _could_ lead to perceived security that isn't there.

This could have knock on effects of e.g. us writing [non-exhaustive type matching](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#exhaustiveness-checking) as technically this could be anything as we aren't parsing it. 

The example I am thinking of is:
```
type Cat = 'laws' | 'policies'
const cat: Cat = params.cat as Cat


switch(cat) {
  case 'laws':
    renderLaws()
    break

  case 'policies':
    renderPolicies()

  // We don't need a default as we've exhausted Cat
}
```

We could write a simple [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) as a parser and would have to deal with the unparsed state here.


Again not blocking, but it does create a leak in the app.

jamesgorrie (approved):
Noice

src/pages/search/index.tsx
PatFawbertMills: Yeh this is an interesting trade-off. I needed to introduce this casting because I make other parts of the app _more_ type secure. Previously we were passing a string around when we were expecting a `TDocumentCategory`.

I agree with you though, what I should really do is allow a string to be passed into the component, but then validate this against our type at the next stage, because as you say we can't actually guaruntee this type at this point.

PatFawbertMills: I'm removed this force typing and kept the checks within the `SearchResultList` the same.

PR #296, opened 2024-12-16
feat: add new cpr related corpora to the frontend config

odrakes-cpr:
# What's changed

we have new corpora entities that fall under laws and policy ( "CPR.corpus.i00000589.n0000", "CPR.corpus.i00000591.n0000", "CPR.corpus.i00000592.n0000") this new data will be coming through to the frontend app, to allow search to filter on this data we need to add these corpora into the frontend config

Screenshot 2024-12-16 at 13 57 51

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/themes/cpr/config.json
+++ b/themes/cpr/config.json
@@ -15,13 +15,25 @@
       {
         "label": "Policies",
         "slug": "policies",
-        "value": ["CCLW.corpus.i00000001.n0000", "CPR.corpus.i00000001.n0000"],
+        "value": [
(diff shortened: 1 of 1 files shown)


katybaulch:
> It looks like the corpus IDs are different between staging and prod. Let me consolidate these to what is in this PR

I've updated this now and I think we're good to go :)

katybaulch (approved):
lgtm

PR #300, opened 2025-01-02
feat: auto-deploy on merge to main

jamesgorrie:
# What's changed

We'd like to trigger a deploy in `navigator-infra` from this repo `navigator-frontend`.

We don't have enough confidence yet that we can do this on each merge, so this is an interim step to help us figure out where the largest holes in our coverage is by allowing us to opt-in to auto deploying via `workflow_call`, or trigger it manually via `workflow_dispatch`.

To auto deploy, you will need to add the label `deploy` to your PR. This is read on merge, and the auto-deploy step is run if `true`.

The process is a little convoluted. This is because we need to read the labels from the PR. This is only while we're in testing phase. It should be a lot easier once we don't have to read PR labels.

The flow is:
* Create PR
* Label PR with `deploy`
* Merge PR
* If labelled
  * The action reads the PR commit
  * Goes through the hoops to deploy

✨ This should also get us a little closer to using SHAs as versions and also CD.

## A diagram of current vs ideal, proposed workflows
```mermaid
flowchart TD
    nf["navigator-frontend"]
    test["test"]
    tf["theme-factory"]
    tag["tag / semver"]
    pr["get-next-tag-from-pr-body"]
    push["retag-and-push-to-ecr"]

    subgraph current
        subgraph ./
            nf-->tf
        end
        subgraph ./reusable-workflows
            tf-->build-->test-->tag
        end

        subgraph external actions
            tag-->pr-->push
        end

        subgraph navigator-infra
            push-->manualDeploy["Manual deploy"]
        end
    end

    nfLocal["navigator-frontend"]
    subgraph proposed
        subgraph ./2["./"]
            nfLocal-->buildLocal["build"]-->testLocal["test"]
            deployLocal
        end
        
        subgraph externalActions["external actions"]
            testLocal-->pushLocal["retag-and-push-to-ecr"]-->deployLocal
        end

        subgraph navigator-infra2["navigator-infra"]
            deployLocal["deploy"]-->niDeploy["deploy"]
        end
    end


    missing
    
    style missing stroke:red;
    style testLocal stroke:red;
    style proposed stroke:green,stroke-width:5px;
```

## Proposed version

- [x] Patch

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -88,3 +88,71 @@ jobs:
     permissions:
       contents: write
     uses: climatepolicyradar/reusable-workflows/.github/workflows/release.yml@v6
+
+  # everything below here is moving towards a new workflow
+  # and will superscede `release`
+  deploy:
+    needs: release
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        theme: [cpr, cclw, mcf]
+    steps:
+      - name: Build
+        run: |
+          export THEME=${{ matrix.theme }}
+          make build
+      # TODO: We need to run a test phase here - but for now we're relying on the previous deploy steps that run the tests
+      - name: Release
+        uses: climatepolicyradar/retag-and-push-to-ecr@v1
+        env:
+          DOCKER_REGISTRY: ${{ secrets.DOCKER_REGISTRY }}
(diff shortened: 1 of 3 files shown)


jamesgorrie:
@odrakes-cpr - thanks. I'm going to merge as
- it has no side effects in PROD without human intervention
- it's easy to back out of

But allows me to continue refining this.

PR #301, opened 2025-01-06
feat: add e2e to CI workflow

jamesgorrie:
# What's changed

## Proposed version

- [ ] Skip auto-tagging
- [ ] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/tests/search.spec.ts
+++ b/tests/search.spec.ts
@@ -22,12 +22,25 @@ test("search", async ({ page }) => {
   const searchResultsHeading = page.getByRole("heading", { name: "Search results" });
   const searchResults = page.locator("div").filter({ has: searchResultsHeading }).last();
   await expect(searchResults).toBeVisible();
+
+  /** Check the structure of the search result */
+  const firstSearchResult = searchResults.getByRole("listitem").first();
+  /** heading */
+  await expect(firstSearchResult.getByRole("heading")).toBeVisible();
+  /** description */
+  await expect(firstSearchResult.getByRole("paragraph")).toBeVisible();
+
+  /** TODO: Make the markup more semantic */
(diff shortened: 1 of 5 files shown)


tests/search.spec.ts
```diff
@@ -22,12 +22,25 @@ test("search", async ({ page }) => {
...
   await expect(searchResults).toBeVisible();
+
+  /** Check the structure of the search result */
+  const firstSearchResult = searchResults.getByRole("listitem").first();
+  /** heading */
+  await expect(firstSearchResult.getByRole("heading")).toBeVisible();
+  /** description */
+  await expect(firstSearchResult.getByRole("paragraph")).toBeVisible();
+
+  /** TODO: Make the markup more semantic */
```
jamesgorrie, line 33: I've copied this from the [old cypress tests](https://github.com/climatepolicyradar/navigator-frontend/blob/bcb853b8f411ab493b0c69e1cb2cc78db9a2f32f/e2e/cypress/e2e/cclw/journeys/search.cy.js#L43-L51) as it felt like a valuable and I want to delete that test without losing any coverage.

odrakes-cpr (approved):
lgtm

PatFawbertMills:
Looks good. Just so I understand it - because it isn't nested in the `/cclw` folder it will run for all themed versions of the application?

jamesgorrie:
@PatFawbertMills - nope, it is currently running on the default theme, `cpr`. 

Given that this isn't running any theme specific tests - would you expect it too? I wonder how that scales. e.g. we have 12 custom apps, or this is white-labelled, would we expect to spin up a new set of tests for each?

I can imagine this being the case where we might be tenanted, but we currently aren't.

PR #305, opened 2025-01-13
APP-147 review package versions

PatFawbertMills:
# What's changed
- Updating a number of the packages
- Improving our security in regards to flagged vulnerabilities in the packages we were using

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


package.json
```diff
@@ -21,10 +21,9 @@
     "framer-motion": "^11.11.7",
     "jest-environment-jsdom": "^29.7.0",
     "jwt-decode": "^4.0.0",
-    "leaflet": "^1.9.3",
     "next": "^13.5.6",
-    "react": "^18.2.0",
-    "react-dom": "^18.2.0",
+    "react": "18.2.0",
+    "react-dom": "18.2.0",
```
jamesgorrie, line 26: Locked to specific version! ✨

jamesgorrie (approved):

odrakes-cpr (approved):

package.json
PatFawbertMills: Yeh bit of a

PR #309, opened 2025-01-14
refactor: remove CCLW browse e2e test

jamesgorrie:
# What's changed
- removes cclw browse cypress spec

This is already covered via the [search spec](https://github.com/climatepolicyradar/navigator-frontend/blob/main/tests/search.spec.ts).

Part of https://linear.app/climate-policy-radar/issue/APP-156/remove-cypress-tests

## Proposed version

- [x] Patch


odrakes-cpr (approved):
nice

PR #312, opened 2025-01-16
fix: use @use for sass imports

jamesgorrie:
# What's changed

One of the errors when upgrading to next@15 is [not being able to use `sass @import`](https://sass-lang.com/documentation/breaking-changes/import/) - but to use `@use`.

This fixes that. I was half tempted to remove all the sass, but it's quite a large file.

## Proposed version

- [x] Patch


PatFawbertMills (approved):
LGTM

PR #314, opened 2025-01-16
Next upgrade dynamic imports

jamesgorrie:
# What's changed
- Use next/dynamic for dynamic imports to help with the topLevelAwait errors. I am not 100% sure why this doesn't work natively given [we have topLevelAwait set](https://github.com/climatepolicyradar/navigator-frontend/blob/d4b298c271dbdb0bf30acef77f0fa99172aeb45f/next.config.js#L20) - but hey ho, this is the recommended way.
- removes cache headers locally as it causes refresh issues with webpack's HMR
- upgrades next => v15

## Proposed version

- [x] Patch


PatFawbertMills (approved):
LGTM nice one

PR #315, opened 2025-01-20
feat: v0 of feature flags from posthog

jamesgorrie:
# What's changed

- [Uses Posthog feature flags](https://posthog.com/docs/feature-flags)
- [Uses Posthog's early access feature management](https://posthog.com/docs/feature-flags/early-access-feature-management)
- Dumps a redundant script tag on the family page 

As far as I can tell - the general structure from Posthog is
- feature flags manage global flags across your site (they can be turned on, released incrementally etc)
- early access allows users to override the general setting to a personal value

We have chosen posthog as that's where we are storing our other usage data, and feature flags will eventually be part of that, so integration comes for free!

A user will be able to access `/_feature-flags` and turn on any early access features we have made available.

From chats with @odrakes-cpr, and also because we initiate Posthog in Google Tag Manager, we are maintaining use through SSR only for now.

The main reason we are dumping the script tag on the page is to see if the value switches dependant on my users, which I suspect it won't due to caching.

☝️ This will be step 2 in our implementation

---

### Auto generate widget from Posthog

Screenshot 2025-01-21 at 14 58 39

Part of https://linear.app/climate-policy-radar/issue/PLA-405/enable-feature-switches-to-show-the-most-basic-version-of-concept

## Proposed version

- [x] Patch


odrakes-cpr (approved):
lgtm

katybaulch:
Would love a demo of this at some point @jamesgorrie - potentially a show and tell contender?

PR #316, opened 2025-01-20
Add typed getConfig to ApiClient

jamesgorrie:
# What's changed
- adds `getConfig` to the api client
- implements this where we use client.get(`/config`)
- removes some redundant `Geography` types

## Testing
I haven't added e2e tests as tests wouldn't have caught this as this broke when we deployed `navigator-backend` which removed the `organisations` field.

The type system here catches the error we wanted to guard against.

We should still be thinking about how we test `navigator-frontend` when changes are made to `navigator-backend` (had a small thought of adding visual regression to NB).

## Proposed version

- [x] Patch

### DIFF
--- a/src/api/http-common.ts
+++ b/src/api/http-common.ts
@@ -1,4 +1,5 @@
 /* eslint-disable no-console */
+import { TCorpusTypeDictionary, TDataNode, TGeography, TLanguages } from "@types";
 import axios, { AxiosInstance, AxiosResponse } from "axios";
 
 export async function getEnvFromServer() {
@@ -33,10 +34,10 @@ class ApiClient {
   /**
    * Submit a GET request and return the response as a mapped promise.
    */
-  get(url: string, params?: any): Promise<AxiosResponse> {
+  get<T = any>(url: string, params?: any): Promise<AxiosResponse<T>> {
     // console.log(`GET: ${this.baseUrl}${url}`);
     return this.axiosClient
-      .get(url, { params })
+      .get<T>(url, { params })
       .then((res: any) => res)
       .catch((err) => {
         console.log(err);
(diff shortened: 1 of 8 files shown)


odrakes-cpr (approved):
let's get it in

src/api/http-common.ts
```diff
@@ -69,6 +70,10 @@ class ApiClient {
         return Promise.reject(err);
       });
   }
+
+  getConfig() {
+    return this.get<{ geographies: TDataNode<TGeography>[]; corpus_types: TCorpusTypeDictionary; languages: TLanguages }>("/config");
```
annaCPR, line 75: Maybe it's not important but we're missing document_variants in this dictionary. I don't think we use it anywhere in the code but it is being returned by the /config endpoint.

annaCPR (approved):
Can we add a test for the Notes section?

src/api/http-common.ts
jamesgorrie: I'll leave it out as it might be an indicator that we can remove it from the config?

jamesgorrie:
> Can we add a test for the Notes section?

@annaCPR I couldn't quite figure out the right place to test this, I think mainly because I was thinking of testing the regression we had which I have mentioned in the discussion.

Outside of that context - what are we trying to test? "Given the page has corpus description information, we then display that information?"

annaCPR:
Also, there is some logic to extracting the correct corpus image and notes that the type is not guarding against.

jamesgorrie:
>  To keep things simple we could even add a couple of asserts to the search.spec.ts

Yea, I would like to avoid this as the search is the search journey, showing a corpus description on a page isn't really relevant to that.

Agreed that we should have a test, as per the previous comment, I was unsure of what test to help with this. The refactor bug will now be guarded against with typing (a form of testing).

I'll have a little more of a think about this, but in terms of how much effort to put into it, I might keep it minimal as it wasn't an urgent break.

annaCPR:
Cool. Thank you again for looking into this.

PR #319, opened 2025-01-21
APP 164 - reports category

PatFawbertMills:
# What's changed
- Adding "Reports" category
- Adding sub filter for "Author Type"
- Fixing issues with selected state on filters that were not selected but the names were similar

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/src/components/blocks/FilterOptions.tsx
+++ b/src/components/blocks/FilterOptions.tsx
@@ -23,6 +23,17 @@ type TProps = {
   themeConfig: TThemeConfig;
 };
 
+const filterIsSelected = (queryValue: string | string[] | undefined, option: string) => {
+  if (!queryValue) {
+    return false;
+  }
+  if (typeof queryValue === "string") {
+    return queryValue === option;
+  } else {
+    return queryValue.includes(option);
+  }
+};
+
 export const FilterOptions = ({ filter, query, handleFilterChange, corpus_types, themeConfig }: TProps) => {
   const [search, setSearch] = useState("");
 
@@ -35,7 +46,7 @@ export const FilterOptions = ({ filter, query, handleFilterChange, corpus_types,
             <InputRadio
               key={option.slug}
               label={option.label}
-              checked={query && query[QUERY_PARAMS[filter.taxonomyKey]] && query[QUERY_PARAMS[filter.taxonomyKey]].includes(option.slug)}
+              checked={query && filterIsSelected(query[QUERY_PARAMS[filter.taxonomyKey]], option.slug)}
               onChange={() => null} // supress normal radio behaviour to allow to deselection
               onClick={() => {
                 handleFilterChange(QUERY_PARAMS[filter.taxonomyKey], option.slug, true);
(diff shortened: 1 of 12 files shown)


PatFawbertMills:
**⚠️  Note to devs: please do not merge until the corresponding work is completed that adds the OEP corpus to the search token for CPR.**

src/utils/buildSearchQueryMetadata.ts
```diff
@@ -0,0 +1,24 @@
...
+    if (Array.isArray(metadataValue)) {
+      metadataForApi = metadataValue;
+    } else {
+      metadataForApi = [metadataValue];
+    }
+    metadataForApi.map((m) => {
+      query.metadata.push({ name: configMetadata.apiMetaDataKey, value: decodeURI(m) });
+    });
+  }
+  return null;
```
jesskelsall, line 23: Suggestion: This function should return a new `query` object rather than mutating the function parameter. This feels like a side effect rather than functional programming, and is much harder to test.

src/components/blocks/FilterOptions.tsx
```diff
@@ -102,7 +102,7 @@ export const FilterOptions = ({ filter, query, handleFilterChange, corpus_types,
       <InputRadio
         key={option}
         label={option}
-        checked={query && query[QUERY_PARAMS[filter.taxonomyKey]] && query[QUERY_PARAMS[filter.taxonomyKey]].includes(option)}
+        checked={query && query[QUERY_PARAMS[filter.taxonomyKey]] && query[QUERY_PARAMS[filter.taxonomyKey]] === option}
```
jesskelsall, line 105: Question: Why is `filter.taxonomyKey` no longer being treated like an array? I'm not seeing the connection with the rest of the diff. Is this a rolled in bug fix?

jesskelsall (changes requested):

src/utils/buildSearchQueryMetadata.ts
PatFawbertMills: Good point, I'll look at that

src/components/blocks/FilterOptions.tsx
PatFawbertMills: Fair question, the type of the `ParsedUrlQuery` is `string | string[]`. I wrote the first solution assuming I was doing the right thing and supporting multiple filters being selected (when we ever expand the filters to be multi-select), but then I noticed a bug (see third bullet point in description) where names that were similar were being checked.

I'll move this into a separate utility that checks the type first.

katybaulch:
@PatFawbertMills @jesskelsall, @odrakes-cpr & I have added the OEP token to staging (and the next time someone deploys the CPR frontend to prod, the OEP token will be updated there too) - so this PR is no longer blocked

odrakes-cpr (approved):
nice tests

PR #320, opened 2025-01-22
Feature flags powered by cachable cookies

jamesgorrie:
# What's changed

- Uses Posthog's [Early access app](https://posthog.com/docs/feature-flags/early-access-feature-management) to allow a user to opt-into feature flags
- Sets those values on a CPR specific cookie

In combo with https://github.com/climatepolicyradar/navigator-infra/pull/967

Screenshot 2025-01-22 at 20 58 01

## Proposed version

- [x] Patch

Part of https://linear.app/climate-policy-radar/issue/APP-170/enable-feature-switches-to-show-the-most-basic-version-of-concept

### DIFF
--- a/src/pages/_feature-flags.tsx
+++ b/src/pages/_feature-flags.tsx
@@ -1,8 +1,11 @@
-"use client";
+import Button from "@components/buttons/Button";
+import { deleteCookie, setCookie } from "@utils/cookies";
+import getDomain from "@utils/getDomain";
 import { usePostHog } from "posthog-js/react";
 
-function FeatureFlags() {
+export default function FeatureFlags() {
   const posthog = usePostHog();
+
   /**
    * This key is a public key.
    * @see: https://posthog.com/docs/privacy#is-it-ok-for-my-api-key-to-be-exposed-and-public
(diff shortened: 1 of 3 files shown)


src/pages/_feature-flags.tsx
```diff
@@ -13,10 +15,26 @@ function FeatureFlags() {
   });
 
   return (
-    <div>
-      <button id="beta-button">Public Betas</button>
+    <div className="h-screen flex items-center justify-center">
+      <Button id="beta-button">Feature Flags</Button>
+      <Button
+        onClick={() => {
+          posthog.getEarlyAccessFeatures((posthogFeatureFlags) => {
```
PatFawbertMills, line 22: Question: is this so that we can manually control when to get the latest feature flags in a given session?

PatFawbertMills (approved):
LGTM

src/pages/_feature-flags.tsx
jamesgorrie: This
- reads the feature flags for a given user e.g. you, as identified by a [`distinct_id`](https://posthog.com/docs/product-analytics/identify)[^1]
- We then use that response to generate our own `feature_flags` and assign that to a cookie

This is because we need a way to allow these values through the cache which the standard posthog implementation doesn't allow.

[^1]: this is an anonymous ID and doesn't follow you around

PR #322, opened 2025-01-22
Add concepts to family page behind feature flag

jamesgorrie:
# What's changed

- if in concepts-v1 feature flag
- reads the endpoint [created here](https://github.com/climatepolicyradar/navigator-backend/pull/447)
- **[CLUDGE]:** as we are waiting for concepts to be added onto the family response, we just stick a hardcoded value in there for now
- **[CLUDGE]:** fetches full-data from S3 bucket for concepts
- renders a boring list of concepts on the page
- adds new `SearchResponse` type that is a replica of https://github.com/climatepolicyradar/cpr-sdk/blob/ef591a155b63de8680be91bfe4180bdace9c8692/src/cpr_sdk/models/search.py#L609

## Proposed version

- [x] Patch

## 👀 

Screenshot 2025-01-27 at 16 18 30

---

The design was lifted from the only place I could see where we had a label and count
e.g.
Screenshot 2025-01-27 at 16 23 23

Part of https://linear.app/climate-policy-radar/issue/APP-170/enable-feature-switches-to-show-the-most-basic-version-of-concept

### DIFF
--- a/src/pages/document/[id].tsx
+++ b/src/pages/document/[id].tsx
@@ -373,11 +373,13 @@ const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({
               <section className="mt-8">
                 <Heading level={4}>Concepts</Heading>
                 <div className="flex text-sm">
-                  <ul>
+                  <ul className="flex flex-wrap gap-2">
                     {concepts.map((concept, i) => {
                       return (
                         <li key={i}>
-                          <LinkWithQuery href={`/concepts/${concept.wikibase_id}`}>{concept.preferred_label}</LinkWithQuery>
+                          <LinkWithQuery className="capitalize" href={`/concepts/${concept.wikibase_id}`}>
+                            <Button>{concept.preferred_label}</Button>
+                          </LinkWithQuery>
                         </li>
                       );
                     })}
--- a/src/pages/document/[id].tsx
+++ b/src/pages/document/[id].tsx
@@ -184,15 +184,17 @@ const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({
   const [concepts, setConcepts] = useState<TConcept[]>([]);
 
   useEffect(() => {
-    // [conceptId, count]
-    const conceptsData: [string, number][] = vespaFamilyData
+    const conceptsData: { conceptId: string; count: number }[] = vespaFamilyData
       ? vespaFamilyData.families.flatMap((family) => {
           return family.hits.flatMap((hit) => {
-            return Object.entries(hit.concept_counts);
+            return Object.entries(hit.concept_counts).map(([conceptId, count]) => ({
+              conceptId,
+              count,
+            }));
           });
         })
       : [];
-    const conceptsS3Promises = conceptsData.map(([conceptId, count]) => {
+    const conceptsS3Promises = conceptsData.map(({ conceptId, count }) => {
       const url = `https://cdn.dev.climatepolicyradar.org/concepts/${conceptId}.json`;
       return fetch(url).then((response) => response.json());
     });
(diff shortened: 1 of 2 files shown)


src/types/types.ts
```diff
@@ -308,3 +308,30 @@ export interface TDictionary<T> {
 
 export type TOrganisationDictionary = TDictionary<TOrganisation>;
 export type TCorpusTypeDictionary = TDictionary<TCorpusType>;
+
+export type Concept = {
```
PatFawbertMills, line 312: Suggestion: we use the `TConcept` prefix with T for consistency (open to changing this if people feel strongly opinionated on this - just something I've carried over from previous projects)

PatFawbertMills (changes requested):

PatFawbertMills (approved):
LGTM!

jamesgorrie:
Pausing this as we're a little unsure about the data structure from the API.

src/pages/document/[id].tsx
```diff
@@ -169,6 +170,37 @@ const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({
...
+   * - getting a list of those IDs
+   * - creating promises to the S3 file that has more data about the concept
+   * - resolving that promise and treating that as the concept data.
+   *
+   * Not ideal, but we are working on getting more useful data on the family response.
+   */
+  const [concepts, setConcepts] = useState<TConcept[]>([]);
+
+  useEffect(() => {
+    // [conceptId, count]
```
Comment (katybaulch) [inline src/pages/document/[id].tsx:187]: suggestion: can we remove this if it's no longer needed? if it is needed, please can we add a TODO anda. ticket number so it doesn't get lost?

src/pages/document/[id].tsx
```diff
@@ -169,6 +170,37 @@ const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({
...
+    const conceptsData: [string, number][] = vespaFamilyData
+      ? vespaFamilyData.families.flatMap((family) => {
+          return family.hits.flatMap((hit) => {
+            return Object.entries(hit.concept_counts);
+          });
+        })
+      : [];
+    const conceptsS3Promises = conceptsData.map(([conceptId, count]) => {
+      const url = `https://cdn.dev.climatepolicyradar.org/concepts/${conceptId}.json`;
```
Comment (katybaulch) [inline src/pages/document/[id].tsx:196]: suggestion: Can we get the CDN from the env variables instead of hardcoding it here?

Comment (katybaulch) [reply src/pages/document/[id].tsx:196]: We do this with the targets data -- I *think* it's `S3_PATH`

katybaulch (dismissed):
Looking good! Couple of suggestions 👍🏻

PatFawbertMills (approved):
Looks good, WIP but behind the FF so all fine to me.

src/pages/document/[id].tsx
Comment (PatFawbertMills) [reply src/pages/document/[id].tsx:196]: Might be premature to, but I agree in principle

src/pages/document/[id].tsx
Comment (jamesgorrie) [reply src/pages/document/[id].tsx:187]: > can we remove this if it's no longer needed?

What isn't needed?

src/pages/document/[id].tsx
Comment (jamesgorrie) [reply src/pages/document/[id].tsx:196]: More than happy to.
[Could we get the naming of that env value a bit better first](https://github.com/climatepolicyradar/navigator-infra/pull/971)?

Comment (jamesgorrie) [reply src/pages/document/[id].tsx:196]: Actually looking at it - that PR isn't fully featured, and we might need to do some new work. 

Do we want to block on this?

src/pages/document/[id].tsx
Comment (katybaulch) [reply src/pages/document/[id].tsx:187]: Line 187 has some commented out code :)

Comment (jamesgorrie) [reply src/pages/document/[id].tsx]: Apologies - this was to give an indication of what the tuple was. 

I've destructured it and put it into an object with named properties now

PR #323, opened 2025-01-22
APP-74 enable mcf reports filters

PatFawbertMills:
# What's changed
- Adding document 'type' selector for when user selects MCFs category
- Handling building the search query payload based on the selected fund and type
- Adding tests for the `canDisplayFilter` utility, as this now contains a decent chunk of config/business logic

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


katybaulch (approved):
lgtm, nice!!!

PatFawbertMills:
Don't merge this yet please - I need to check that we are fine to deploy this for MCFs before they add any guidance documents.

PR #324, opened 2025-01-28
Update references to family instead of doc

katybaulch:
# What's changed

Our language is inconsistent. In the frontend, we had references to `totalDocuments`, `calculateTotalDocuments, `documentTotals` and `TDocumentTotals` in the `instructions.tsx` file. An inspection of the code shows that this component wants to display the counts of families from the config endpoint by using the counts from corpus.count_by_category. 

To make sure we are using consistent language in our codebase to refer to the same things, I've updated the aforementioned references to refer to families instead of documents so we are clear as developers what the counts refer to.

## Proposed version

- [x] Skip auto-tagging
- [ ] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


odrakes-cpr (commented):
should we also change the name of the helper also

odrakes-cpr (approved):

odrakes-cpr (approved):

PR #330, opened 2025-01-30
concept_filters search

jamesgorrie:
# What's changed

Based on [the documentation of the CPR SDK](https://github.com/climatepolicyradar/cpr-sdk/blob/main/README.md#filtering-for-concept-counts) - we should be able to send over `concept_filters` in the shape of ```{ name: "id", value: `${conceptId}` }```.

As
- the shape expected by `navigator-backend` ([`SearchResponseBody`](https://github.com/climatepolicyradar/navigator-backend/blob/main/app/models/search.py#L54)) - is actually just a subtype of the [SDK's `SearchParameters`](https://github.com/climatepolicyradar/cpr-sdk/blob/bdb3bfdc213923e6a580da1dd91fcc626f636cbb/src/cpr_sdk/models/search.py#L194)
- [and this is parsed at the route level](https://github.com/climatepolicyradar/navigator-backend/blob/main/app/api/api_v1/routers/search.py#L47)
- **We don't need to do any work on the `navigator-backend`**

### Document page
Screenshot 2025-02-03 at 16 20 40

## Semver
- [x] Patch

### DIFF
--- a/src/constants/queryParams.ts
+++ b/src/constants/queryParams.ts
@@ -23,4 +23,7 @@ export const QUERY_PARAMS: TQueryStrings = {
   sector: "sc",
   // Reports
   author_type: "at",
+  // Passthrough
+  "concept_filters.id": "concept_filters.id",
+  "concept_filters.name": "concept_filters.name",
 };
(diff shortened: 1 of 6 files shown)


katybaulch (approved):
Great work - super excited to see this!!

PR #332, opened 2025-02-03
Feature/app 197 group concepts at family level by root concepts

katybaulch:
# What's changed

There are 8 root level concepts:
1. Targets (Q1651)
2. Sectors (Q709)
3. Climate risk (Q975)
4. Fossil fuels (Q638)
5. Impacted groups (Q672)
6. Finance (Q1337)
7. Instruments (Q1171)
8. Greenhouse gases (Q218)

The corresponding Q number refers to its ID on this website https://climatepolicyradar.wikibase.cloud/wiki/Main_Page. A concept's page on this website can be found using the link https://climatepolicyradar.wikibase.cloud/wiki/Item:Q_NUM where Q_NUM is the `wikibase_id` associated with the concept.

Instead of showing only the concepts directly associated with a family and its documents, we want to show:
- where a concept is either a root level concept or is directly a subconcept_of a root level concept
    - for each root level concept, show the root level concept name and the total sum of the concepts associated with that root level in the family/documents
    - if a root level concept has no associated concepts in that family/its documents, do not coalesce or show the root level concepts aren't associated with the family/its documents
    - some concepts may be subconcepts_of one or more root level concepts. where this is true, the concept count must be incremented for root level concepts directly associated with the concept
- where a concept is not a root level concept OR directly a subconcept_of one or more of the root level concepts, show `Other` and count the total sum of each of the concept hits where the parent concept is not directly one of the root level concepts

When displaying these root level concepts and their counts:
- We must only show unique concepts
- Concepts should be stack ranked most-least
- root level concepts (with the exception of the `Other` category) are clickable and will currently take you to the wikibase page associated with that root level concept

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/src/pages/document/[id].tsx
+++ b/src/pages/document/[id].tsx
@@ -226,11 +226,10 @@ const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({
     // Process concepts
     const processedConceptCounts = processConcepts(concepts);
 
-    // Sort concepts by count, descending (with 'Other' always at the end if present)
-    const sortedConcepts = Object.entries(processedConceptCounts).sort(([a], [b]) => {
-      if (a === "Other") return 1;
-      if (b === "Other") return -1;
-      return processedConceptCounts[b] - processedConceptCounts[a];
+    // Sort concepts by count, descending, and then alphabetically if counts are the same
+    const sortedConcepts = Object.entries(processedConceptCounts).sort(([conceptA, countA], [conceptB, countB]) => {
+      if (countB !== countA) return countB - countA;
+      return conceptA.localeCompare(conceptB);
     });
 
     setRootLevelConcepts(sortedConcepts);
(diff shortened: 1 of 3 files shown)


src/pages/document/[id].tsx
```diff
@@ -194,16 +195,46 @@ const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({
...
+    const processedConceptCounts = processConcepts(concepts);
+
+    // Sort concepts by count, descending, and then alphabetically if counts are the same
+    const sortedConcepts = Object.entries(processedConceptCounts).sort(([conceptA, countA], [conceptB, countB]) => {
+      if (countB !== countA) return countB - countA;
+      return conceptA.localeCompare(conceptB);
+    });
+
+    setRootLevelConcepts(sortedConcepts);
+  }, [concepts]);
```
Comment (jamesgorrie) [inline src/pages/document/[id].tsx:236]: suggestion: could we wrap this into 1 effect? It would reduce the need for 2 states (`concepts` and `rooLevelConcepts`) to 1 state (`rootLevelConcepts`), which feels a little simpler.

jamesgorrie (approved):
Just double checking the logic here:
- On family pages we only show `rootLevelConcepts`
- On document pages we show all concepts

So we don't show sub-concepts (grouped) on the family page?

src/pages/document/[id].tsx
Comment (katybaulch) [reply src/pages/document/[id].tsx]: Done.

PR #333, opened 2025-02-03
fix: automatically save feature flags on save

jamesgorrie:
# What's changed
- automatically saves features flags on save
- adds a notice as to why it might not be working for folks

## 👀 
Screenshot 2025-02-03 at 16 37 59

## Proposed version

- [x] Patch


katybaulch (approved):
Nice tidy 🧹

PR #335, opened 2025-02-04
Concept filtering from document page

katybaulch:
# What's changed

- Click a concept on the document page
- That concept should be applied to the search query terms & router
- The clicked concept(s) should show up as pills
- The user can reclick a concept(s) to remove the concept filters

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


src/components/documents/DocumentHead.tsx
```diff
@@ -100,6 +102,42 @@ export const DocumentHead = ({ document, family, handleViewOtherDocsClick, handl
             </div>
           </div>
         </div>
+
+        <div className="flex my-4">
```
PatFawbertMills, line 106: Suggestion: move the div with the margin inside the conditional render, otherwise we will have a white space when there are no concepts loaded.

src/constants/queryParams.ts
```diff
@@ -23,7 +23,7 @@ export const QUERY_PARAMS: TQueryStrings = {
   sector: "sc",
   // Reports
   author_type: "at",
-  // Passthrough
+  // Pass through
   "concept_filters.id": "concept_filters.id",
```
PatFawbertMills, line 27: Suggestion: use a shortened term here. The idea of the aliases was to twofold:
- Create a more readable (and shorter) URL
- Obscure some of the business logic from the user, e.g. users often don't think in terms of "IDs"

PatFawbertMills: Could use `cfi` and `cfn`

PatFawbertMills (changes requested):
Couple of suggestions to consider.

src/components/documents/DocumentHead.tsx
katybaulch: done

src/constants/queryParams.ts
katybaulch: have done this!

katybaulch: from Patrick on slack:

> for the QUERY_PARAMS constant, the idea is we have the key - which is our reference for in the code, so we can make this something that makes it easy for us to develop with, and then the value - this is what is displayed to the user in the URL. When you are on the search page and you interact with the filters you’ll see the URL update with the value side of our QUERY_PARAMS, whilst under the hood in the code we are always referring to the key
> 
> the blast radius of changing the value side will mainly affect users who have bookmarked certain searches or try to type straight into the URL to apply filters, as they will discover the alias in their URL will no longer work
> 
> i don’t see that being a worry right now though as concepts are new

PatFawbertMills (approved):
LGTM nice one

katybaulch:
image

katybaulch:
image

PR #336, opened 2025-02-04
OEP fixes

PatFawbertMills:
# What's changed
- Image was referring to lower case image = was breaking when not on local
- Category comparison was case sensitive which was breaking the links from the OEP page through to search on "Reports" != "reports"

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


src/components/blocks/SearchFilters.tsx
```diff
@@ -28,7 +28,7 @@ const MethodologyLink = dynamic(() => import(`/themes/${process.env.THEME}/compo
 
 const isCategoryChecked = (selectedCatgeory: string | undefined, themeConfigCategory: TThemeConfigOption) => {
   if (selectedCatgeory) {
-    if (selectedCatgeory === themeConfigCategory.slug) {
+    if (selectedCatgeory.toLowerCase() === themeConfigCategory.slug.toLowerCase()) {
```
katybaulch, line 31: praise: beautiful!

katybaulch (approved):
lgtm!!

PR #337, opened 2025-02-04
fix: remove hardcoded data from concepts lookup

jamesgorrie:
# What's changed
- removes hardcoded data
- make `concept_counts` optional as it is from the API
- splits `concept_counts` key by `:` as the key is now the `{id}:{name}`

https://github.com/user-attachments/assets/9d03a496-90ff-4089-84d4-e09fa3725137

## Proposed version

- [x] Patch

### DIFF
--- a/src/pages/document/[id].tsx
+++ b/src/pages/document/[id].tsx
@@ -171,39 +171,30 @@ const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({
     }
   };
 
-  /**
-   * Beta concepts work - this work is all behind the concepts-v1 feature flag.
-   *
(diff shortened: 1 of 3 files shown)


odrakes-cpr (approved):

katybaulch (approved):
LOVE IT!!!

Just wanted to check that we are currently happy with all concepts showing on the doc page, rather than grouping by root level concepts?

jamesgorrie:
> Just wanted to check that we are currently happy with all concepts showing on the doc page, rather than grouping by root level concepts?

@katybaulch yup! 

I think the grouped concepts _without_ showing the sub concepts on the family page might need to show them - but let's get in front of some 👀 !

PR #338, opened 2025-02-04
fix: remove queryString types in favour of const

jamesgorrie:
# What's changed
- simplifies our type system
- removes the amount of code changes needed to make a querystring change

Just noticed this while we were working on the concepts filtering (#332 #334) work.

## Proposed version

- [x] Patch


katybaulch (approved):
very nice *borat gif would go here*

PatFawbertMills (approved):
Looks good

PR #341, opened 2025-02-06
Show root level concept & sub concepts and their counts on family page

katybaulch:
# What's changed

image

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


src/pages/document/[id].tsx
```diff
@@ -384,31 +381,42 @@ const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({
...
+
+                    <div className="flex text-sm">
+                      <ul className="flex flex-wrap gap-2">
+                        {Object.entries(subconcepts).map(([subconcept, { name, count, wikibaseId }]) => (
+                          <li key={subconcept}>
+                            <ExternalLink
+                              className="capitalize"
+                              url={`https://climatepolicyradar.wikibase.cloud/wiki/Item:${wikibaseId}#${encodeURIComponent(name)}`}
+                            >
+                              <Button color="clear" data-cy="view-family-concept" extraClasses="flex items-center text-sm">
```
Comment (jamesgorrie) [inline src/pages/document/[id].tsx:410]:
```suggestion
                              <Button color="clear" extraClasses="flex items-center text-sm">
```

as I don't think we use it anywhere and we're moving away from Cypress.

src/utils/processConcepts.ts
```diff
@@ -12,15 +12,11 @@ const ROOT_LEVEL_CONCEPTS = {
...
-}
-
-export const processConcepts = (concepts: (TConcept & { count: number })[]): { [key: string]: number } => {
-  const conceptMap: { [key: string]: number } = {};
-  let otherCount = 0;
+export const processConcepts = (
+  concepts: (TConcept & { count: number })[]
+): { [rootConcept: string]: { [subconcept: string]: { name: string; count: number; wikibaseId: string } } } => {
+  const conceptMap: { [rootConcept: string]: { [subconcept: string]: { name: string; count: number; wikibaseId: string } } } = {};
+  let otherConcepts: { [subconcept: string]: { name: string; count: number; wikibaseId: string } } = {};
```
jamesgorrie, line 19: readability: not sure if it's just because it's not in an IDE - but I am finding this quite hard to parse...

jamesgorrie (approved):
looking grand!

src/utils/processConcepts.ts
katybaulch: fixed

src/utils/processConcepts.ts
```diff
@@ -12,15 +12,23 @@ const ROOT_LEVEL_CONCEPTS = {
...
+interface Concept {
+  name: string;
   count: number;
   wikibaseId: string;
 }
 
-export const processConcepts = (concepts: (TConcept & { count: number })[]): { [key: string]: number } => {
-  const conceptMap: { [key: string]: number } = {};
-  let otherCount = 0;
+interface ConceptMap {
```
jamesgorrie, line 21: In TypeScript - I think I would just define these as `types` i.e. `type ConceptMap`. 

There's no official take on this - but a (if not the only) functional difference between a `type` and an `interface` is that you can `extend` an `interface`.

TO that end I normally follow the logic
- if it's never extended e.g. a shape of some data - use `type`
- if you do need to extend something (which is rarely) use `interface`

@PatFawbertMills @jesskelsall any thoughts?

katybaulch: Noted for next time -- will leave this for now just in the interest of time

PR #345, opened 2025-02-10
feature/app-155-preventionweb-is-missing-a-logo-on-appcpr-and-cclw-under

odrakes-cpr:
fix: no explicit setting of height and width means image not rendered
- the images for the prevent web logo was being rendered at 0 x 0 pixels

# What's changed

Before
Screenshot 2025-02-10 at 14 27 49

After

Screenshot 2025-02-10 at 14 45 13

N.B - have tested to ensure that the other logos for other corpora is rendered properly
## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


katybaulch (approved):
lgtm

PR #346, opened 2025-02-10
feature/app-158-counts-on-country-pages-and-search-do-not-add-u

odrakes-cpr:
# What's changed

fix: add prevention web and cpd to the filter query we were sending to the vespa db in the cclw theme app.

- we were missing these corpora ids when searching the vespa laws and policies so it would only return the cclw docs, and would create a mismatch between the geo pages and search pages

As per the master sheet, cclw should contain, cpd,cclw,unfcc &. prevention web
Screenshot 2025-02-10 at 15 51 20

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/themes/cclw/config.json
+++ b/themes/cclw/config.json
@@ -15,13 +15,13 @@
       {
         "label": "Policies",
         "slug": "policies",
-        "value": ["CCLW.corpus.i00000001.n0000"],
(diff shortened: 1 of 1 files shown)


odrakes-cpr:
Please see the same results for the brazil geo page and search page 

Screenshot 2025-02-10 at 15 41 01
Screenshot 2025-02-10 at 15 41 13

annaCPR (approved):
great catch! well done!

PR #352, opened 2025-02-12
feat: add right panel to display concepts on family page

jamesgorrie:
# What's changed
- implementing the designs on [this Figma](https://www.figma.com/design/FHu4NDZiDiCBnpbtD5JxXm/Knowledge-Graph-1st-release-Feb-25?node-id=2002-1018&t=SaqtbiXDFT0fSVjl-0)
- ...

## Proposed version

- [x] Patch


odrakes-cpr (approved):

src/pages/document/[id].tsx
```diff
@@ -451,7 +424,44 @@ const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({
               </section>
             ))}
           </SingleCol>
-        </SiteWidth>
+          {/* TODO: use a panel for this */}
+          {concepts.length > 0 && (
+            <div className="grow-0 shrink-0 px-5 border-l pt-5 w-[460px] text-sm">
```
Comment (PatFawbertMills) [inline src/pages/document/[id].tsx:429]: Question: What happens here on mobile? I don't see anything related to responsiveness.

PatFawbertMills (changes requested):
Check: Just wanted to check we have covered at least some minimum mobile support for the new concepts side bar.

src/pages/document/[id].tsx
Comment (jamesgorrie) [reply src/pages/document/[id].tsx:429]: From @conordelahunty's (and my) thoughts: "that’s beyond our remit right now".
The point of this is to get it to internal people as soon as.

As a minimum - would hiding it be OK?

Comment (PatFawbertMills) [reply src/pages/document/[id].tsx:429]: I had forgotten we have featured flags for this kind of quick iteration, so ignore me. I think its fine to circle back later.

PatFawbertMills (approved):
Exciting update 🎉

PR #361, opened 2025-02-17
fix: don't re-fetch concept data on re-render

jamesgorrie:
# What's changed
- uses useEffectOnce as we have multiple re-renders that kick of multiple fetches for concepts, which is non-performant

## Proposed version

- [x] Patch


katybaulch (approved):
lgtm

PR #362, opened 2025-02-17
feat: split ci-cd into merge_to_main and pull_request

jamesgorrie:
# What's changed
- pull_request
  - adds testing jobs
- merge_to_main
  - adds testing jobs
  - adds deploy steps
    - get `tag`
    - release `tag` to ECR
    - release `tag` to GitHub
- removes ci-cd

This is in the effort of simplifying our pipeline to help move towards deploying on merge-to-main.

Instead of having 1 `ci-cd` script with a set of complex `if` statements, we define our two well-understood workflows.

There is some duplication, but this feels like a good trade off for understandability.

This should also allow us to sunset theme-factory.

## Proposed version
- [x] Patch


.github/workflows/merge_to_main.yml
```diff
@@ -23,9 +25,82 @@ jobs:
...
-          npx wait-on http://localhost:3000 # Update the URL if your server runs on a different port
+          npx wait-on http://localhost:3000
       - run: npx percy snapshot snapshots.js
         env:
           PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
       - if: always()
         run: kill $(lsof -t -i:3000) || true
+
+  # test
+  code-quality:
```
katybaulch, line 36: question (blocking): What has happened to the typescript compiler check we usually do as part of the Node code quality job?

.github/workflows/merge_to_main.yml
```diff
@@ -23,9 +25,82 @@ jobs:
...
+        env:
+          DOCKER_REGISTRY: ${{ secrets.DOCKER_REGISTRY }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        with:
+          repo-name: navigator-frontend-${{ matrix.theme }}
+          semver-tag: ${{ steps.determine_next_tag.outputs.new_tag }}
+      - name: Release to GitHub
+        uses: climatepolicyradar/reusable-workflows/.github/workflows/git.yml@main
+      # TODO: deploy
```
katybaulch, line 106: quibble (non-blocking): I feel as though we are losing the separation of concerns somewhat here by combining the tag & release job with the deploy job. Are we able to separate them, and use an output from the tag & release job above in a new deploy job when it's written?

.github/workflows/pull_request.yml
```diff
@@ -28,9 +31,49 @@ jobs:
...
           PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
       - if: always()
         run: kill $(lsof -t -i:3000) || true
+
+  code-quality:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: trunk-io/trunk-action@v1
+        with:
+          arguments: --ci
```
katybaulch, line 46: question: Same as above.

.github/workflows/merge_to_main.yml
jamesgorrie: I _could_ - but not sure I see what benefit we get from that given the only thing to use it is deploy?

I've generally found being hyper explicit in CI is useful, as reusability has limited benefits - but definitely happy to have my mind changed!

.github/workflows/pull_request.yml
jamesgorrie: Question re: TypeScript? If so, same as above.

katybaulch (approved):
lgtm

PR #364, opened 2025-02-18
APP-223 Bump tailwind major version

PatFawbertMills:
# What's changed
- Tailwind major breaking change upgrade from v3 -> v4
- Lots of small UI fixes along the way
- Removal of all old theme configuration
- Migration over to `.css` files
- Removal of all scss
- Housekeeping of some smaller components as drive-bys

## Proposed version

- [ ] Skip auto-tagging
- [ ] Patch
- [x] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


public/css/cclw/theme.css
```diff
@@ -0,0 +1,83 @@
...
+      font-size: var(--text-2xl);
+    }
+    h2 {
+      font-size: var(--text-xl);
+    }
+  } */
+}
+
+/*
+  TODO: remove specific css styles and overrides by moving into their components
```
jesskelsall, line 22: Happy with this approach for now.

src/styles/cclw.css
```diff
@@ -0,0 +1,69 @@
...
+  --text-h2: 36px;
+  --text-h3: 20px;
+  --text-h4: 21px;
+  --text-h5: 18px;
+  --text-h6: 16px;
+
+  --font-weight-medium: 540;
+}
+
+#cclw .gradient-container {
```
jesskelsall, line 12: What's the difference between this file and `public/css/cclw/theme.css`? Do we need both?

src/styles/main.scss
```diff
@@ -1,6 +1,10 @@
-@tailwind base;
-@tailwind components;
-@tailwind utilities;
+// ------------
+// IMPORTANT:
+// PLEASE NOTE -- THIS FILE IS NOT IN USE ANYMORE
```
jesskelsall, line 3: If it's not in use, let's just delete it entirely.

src/styles/styles.css
```diff
@@ -0,0 +1,631 @@
...
+body,
+html,
+#__next {
+  height: 100%;
+}
+
+body {
+  @apply text-textNormal;
+  font-family: "InterVariable", "Inter", ui-sans-serif, system-ui, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol",
+    "Noto Color Emoji" !important;
```
jesskelsall, line 105: I'm not thrilled by how many `!important` declarations there are in the CSS overall, but I can accept it as a "lesser evil" for now, particularly if it was out of scope of this PR to resolve this. Let's avoid anything like this at all costs when building the component library components.

jesskelsall (changes requested):
On the whole looks solid. I've spot checked some of the places where Tailwind is used and am happy. I have a few queries about some of the specifics of the stylesheets which I'd like addressing before I approve.

public/css/cclw/theme.css
PatFawbertMills: Please see the `/src/styles/cclw` for the correct implementation.

This approach did not work because in the public directory I do not have access to tailwind.

src/styles/main.scss
PatFawbertMills: Done.

src/styles/styles.css
PatFawbertMills: Dunno why I am doing this here, going to review now. Also made me realise I should move the font declaration into a theme var as per the standard 🤦

jesskelsall (approved):
Approved after my comments were addressed and with pragmatic considerations for getting this working for now. We'll continue to improve our Tailwind usage as new components are built.

PR #365, opened 2025-02-18
Display "Guidance" for MCF reports

PatFawbertMills:
# What's changed
- Update the category render code to support checking for MCF as a source and displaying the label "Guidance"

🚨 This is a debatable solution because I am creating a workaround to use the source and casting as a sub category type in order to check against the fund acronym. Really keen to hear people's thoughts on if this is acceptable or not.

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/src/helpers/getCategoryName.ts
+++ b/src/helpers/getCategoryName.ts
@@ -1,14 +1,12 @@
 import { TCategory, TCorpusTypeSubCategory } from "@types";
 
-const reportSubCategories: Record<TCorpusTypeSubCategory | "OEP", string> = {
+const reportSubCategories = {
   AF: "Guidance",
   CIF: "Guidance",
   GCF: "Guidance",
   GEF: "Guidance",
   Reports: "Reports",
   OEP: "Reports",
-  "Intl. agreements": "Intl. agreements",
-  "Laws and Policies": "Laws and Policies",
 };
 
 const subCategories: Record<TCorpusTypeSubCategory, string> = {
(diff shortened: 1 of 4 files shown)


henry-cpr:
I can see why it feels a bit distasteful! 
on the one hand -- it feels like a workaround with a small blast radius, and it's a tiny change, so I'm inclined to say yea fuck it whack it in. 

On the other hand, it does feel like the wrong place to do this. Am I right in interpreting this as a presentation-layer only concern? i.e. backend shouldn't care, a report is a report, but this specific fund calls it different? Or is it that our data model is slightly misaligned with business needs? 

It feels like a can of worms tbh. Personally I'd wait for the emergence of new requirements that require similar workarounds and then work out the right abstraction to hoist this into at that point, since it's such a small requirement at this stage.

src/helpers/getCategoryName.ts
```diff
@@ -1,13 +1,24 @@
...
+const reportSubCategories: Record<TCorpusTypeSubCategory | "OEP", string> = {
+  AF: "Guidance",
+  CIF: "Guidance",
+  GCF: "Guidance",
+  GEF: "Guidance",
+  Reports: "Reports",
+  OEP: "Reports",
+  "Intl. agreements": "Intl. agreements",
+  "Laws and Policies": "Laws and Policies",
+};
```
odrakes-cpr, line 12: Intl Agreements and Laws & Policies, would not fall under reports as they are a different corpus type & category altogether

odrakes-cpr (approved):

jesskelsall (approved):
Happy with this approach.

PR #376, opened 2025-02-24
fix: use main or other documents link on concepts link

jamesgorrie:
# What's changed
- use either `mainDocument` || `otherDocument` slug for link through to document page filtered by concept

This is because there is not always a main document.
From what I can tell if there isn't, there will always be other documents.

We could probably do better typing here, but it fixes the bug for now.

> [!CAUTION]
> This is a really annoying area where our [`strict: false` in our tsconfig](https://github.com/climatepolicyradar/navigator-frontend/blob/main/tsconfig.json#L8) isn't catching things it should...

## To replicate
- turn on concepts-v1 feature flag
- visit https://app.climatepolicyradar.org/document/regulation-eu-2021-82-adaptation-strategy_ba42?q=Adaptation+strategy 

## Proposed version

- [x] Patch

### DIFF
--- a/src/pages/document/[id].tsx
+++ b/src/pages/document/[id].tsx
@@ -491,10 +491,7 @@ const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({
                           return (
                             <li key={concept.wikibase_id}>
                               {conceptDocumentLink ? (
-                                <Link
-                                  className="capitalize hover:no-underline"
-                                  href={`/documents/${conceptDocumentLink}?cfn=${concept.preferred_label}`}
-                                >
+                                <Link className="capitalize hover:no-underline" href={`${conceptDocumentLink}?cfn=${concept.preferred_label}`}>
                                   <Button
                                     color="clear"
                                     data-cy="view-document-viewer-concept"
--- a/src/pages/document/[id].tsx
+++ b/src/pages/document/[id].tsx
@@ -248,6 +248,13 @@ const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({
     });
   });
 
+  let conceptDocumentLink: string | undefined;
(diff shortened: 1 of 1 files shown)


src/pages/document/[id].tsx
```diff
@@ -483,7 +490,7 @@ const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({
                             <li key={concept.wikibase_id}>
                               <Link
                                 className="capitalize hover:no-underline"
-                                href={`/documents/${mainDocuments[0].slug}?cfn=${concept.preferred_label}`}
+                                href={`/documents/${conceptDocumentLink}?cfn=${concept.preferred_label}`}
                               >
```
Comment (odrakes-cpr) [inline src/pages/document/[id].tsx:494]: will the href here not be `/documents/${'/documents/${otherDocuments[0].slug}'}?cfn=${concept.preferred_label}` ? should it be `documents/documents.....`

src/pages/document/[id].tsx
```diff
@@ -483,7 +490,7 @@ const FamilyPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({
                             <li key={concept.wikibase_id}>
                               <Link
                                 className="capitalize hover:no-underline"
-                                href={`/documents/${mainDocuments[0].slug}?cfn=${concept.preferred_label}`}
+                                href={`/documents/${conceptDocumentLink}?cfn=${concept.preferred_label}`}
```
Comment (PatFawbertMills) [inline src/pages/document/[id].tsx:493]: Suggestion: should we also validate that `conceptDocumentLink` is not undefined? As far as I understand the new code on line 251 there is a chance a family could not have either a main or an other document.

Comment (katybaulch) [reply src/pages/document/[id].tsx]: Done

katybaulch (approved):
lgtm (shush)

PatFawbertMills (approved):
Nice one

PR #377, opened 2025-02-25
wip: add strict: true to tsconfig

jamesgorrie:
# What's changed
- uses tsconfig `strict: true`. This is mainly to not ignore nullish values through the codebase which has allowed us to release bugs and is generally considered not safe.

This is more of an RFC to see if we like this approach as it does mucky the codebase a lot, but means we can confidently move forward with it on, and remove the `expect-errors` as we go.

It would also come with a commit hash in https://github.com/climatepolicyradar/navigator-frontend/blob/main/.git-blame-ignore-revs

- [x] Patch

### DIFF
--- a/src/components/blocks/FilterOptions.tsx
+++ b/src/components/blocks/FilterOptions.tsx
@@ -46,9 +46,11 @@ export const FilterOptions = ({ filter, query, handleFilterChange, corpus_types,
             <InputRadio
               key={option.slug}
               label={option.label}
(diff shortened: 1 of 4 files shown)


PatFawbertMills:
In fairness the only reason I can legitimately think of is developer effort to go through and update - and I predict there could be situations that require more than a simple rewrite! Otherwise I do believe it will be a net positive.

PR #379, opened 2025-02-25
fix: use permissions: write-all on merge_to_main

jamesgorrie:
# What's changed
Adds write-all on merge to main.

Currently our build is failing and I would like to rule out permissions.

## Proposed version

- [x] Patch


katybaulch (approved):
lgtm

PR #381, opened 2025-02-25
Node_Env=production at build

jamesgorrie:
# What's changed
We're currently see the build break as it is trying to use tools included in `devDependencies`. 

This ensures we
* install everything
* get the benefits of being in `NODE_ENV=production` that NextJS offers

It should also fix the build.

An alternative, which I think might be clearer, is not to differentiate between `devDependencies` and `dependencies`.

- [x] Patch

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -1,0 +1,177 @@
+name: CI/CD
+
+on:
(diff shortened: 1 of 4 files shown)


PatFawbertMills (approved):
nice catch

PR #382, opened 2025-02-25
Feature/app 299 update family page concepts styling

katybaulch:
# What's changed

- Make family page concepts look like document viewer concepts
- Add new concepts panel component

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/src/utils/processConcepts.ts
+++ b/src/utils/processConcepts.ts
@@ -14,6 +14,26 @@ export const ROOT_LEVEL_CONCEPTS = {
 };
 export const rootLevelConceptsIds = Object.keys(ROOT_LEVEL_CONCEPTS);
 
+export const fetchAndProcessConcepts = async (conceptIds: string[], fetchConcept: (id: string) => Promise<TConcept>) => {
+  const rootConceptsS3Promises = rootLevelConceptsIds.map((conceptId) => {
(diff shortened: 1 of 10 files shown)


src/utils/processConcepts.ts
```diff
@@ -14,6 +14,26 @@ export const ROOT_LEVEL_CONCEPTS = {
 };
 export const rootLevelConceptsIds = Object.keys(ROOT_LEVEL_CONCEPTS);
 
+export const fetchAndProcessConcepts = async (conceptIds: string[], fetchConcept: (id: string) => Promise<TConcept>) => {
```
PatFawbertMills, line 17: Suggestion: have a look into creating custom hooks, ideally we want manage the loading of concepts in a similar manner to how we load other data asynchronously using hooks. See inside the `/hooks` folder for examples.

https://react.dev/learn/reusing-logic-with-custom-hooks

I don't think worth holding up this work though.

PatFawbertMills (approved):
Comment regarding custom hooks but looks good.

src/utils/processConcepts.ts
katybaulch: It will be done now!

PR #384, opened 2025-02-25
feat: use github sha for ecr and github releases

jamesgorrie:
# What's changed
- uses `github.sha` for ecr tag
- ~uses `github.sha` for github release~ removed as per @katybaulch's comment below

This means we will have some doubling up of the GitHub releases - one with the `v` naming convention and one with the hashing naming convention.

The plan is to sunset semver and use hashes only.

## Proposed version
- [x] Patch


.github/workflows/merge_to_main.yml
```diff
@@ -105,9 +103,21 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         with:
           repo-name: navigator-frontend-${{ matrix.theme }}
-          semver-tag: ${{ steps.determine_next_tag.outputs.new_tag }}
+          semver-tag: ${{ github.sha }}
   create-git-release:
+    runs-on: ubuntu-latest
```
katybaulch, line 108: thought: perhaps we skip running this git release job for now, and tag with the SHA as planned? 

Just thinking that this allows us to deploy commit hashes whilst maintaining releases separately from tags (which I thought was the behaviour we wanted to aim for in the long run)?

jamesgorrie: My only concern is visibility - technically we should just be able to grab any hash from any commit to main though, so not that concerned.

jamesgorrie: > git release job **_for now_**

jamesgorrie: I've removed it with that intention.

PR #385, opened 2025-02-25
Feature/app 296 single concept and text search in global search

katybaulch:
# What's changed

Add concept filtering under feature flag in global search.

- Concept filtering by name only enabled at the moment
- Concepts input is a type ahead
- Beta label where the concepts filter accordion appears

NOTE: We don't have all classfiers in yet. It is very likely that the user will select a concept we don't have data for or they will get 0 results based on other options selected. This is okay for now. Multiple concept search is not yet enabled in the API so if the user types more than one concept, the previous concept they had will be removed and the new one applied.

## Proposed version

- [ ] Skip auto-tagging
- [ ] Patch
- [x] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/src/components/accordian/Accordian.tsx
+++ b/src/components/accordian/Accordian.tsx
@@ -4,7 +4,6 @@ import { motion, AnimatePresence } from "framer-motion";
 import { Heading } from "./Heading";
 
 import { AiOutlineMinus, AiOutlinePlus } from "react-icons/ai";
-import { Label } from "@components/labels/Label";
 
 type TProps = {
   title: string;
@@ -14,7 +13,7 @@ type TProps = {
   className?: string;
   showFade?: "true" | "false";
   fixedHeight?: string;
-  isBeta?: boolean;
+  headContent?: React.ReactNode;
 };
 
 export const Accordian = ({
@@ -24,7 +23,7 @@ export const Accordian = ({
   fixedHeight = "300px",
   children,
   showFade = "false",
-  isBeta = false,
+  headContent,
   ...props
 }: TProps) => {
   const [isOpen, setIsOpen] = useState(startOpen);
@@ -34,7 +33,7 @@ export const Accordian = ({
       <div className={`flex justify-between cursor-pointer group`} onClick={() => setIsOpen(!isOpen)} data-cy="accordian-control">
         <div className="flex items-center gap-2">
           <Heading>{title}</Heading>
-          {isBeta && <Label>Beta</Label>}
+          {headContent}
         </div>
         <span className={`text-textDark opacity-40 group-hover:opacity-100 ${isOpen ? "" : ""}`}>
           {isOpen ? <AiOutlineMinus /> : <AiOutlinePlus />}
--- a/src/components/accordian/Accordian.tsx
+++ b/src/components/accordian/Accordian.tsx
@@ -4,6 +4,7 @@ import { motion, AnimatePresence } from "framer-motion";
 import { Heading } from "./Heading";
 
 import { AiOutlineMinus, AiOutlinePlus } from "react-icons/ai";
+import { Label } from "@components/labels/Label";
 
 type TProps = {
   title: string;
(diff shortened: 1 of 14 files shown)


src/components/accordian/Accordian.tsx
```diff
@@ -13,15 +14,28 @@ type TProps = {
...
 
-export const Accordian = ({ title, startOpen = false, overflowOverride, fixedHeight = "300px", children, showFade = "false", ...props }: TProps) => {
+export const Accordian = ({
+  title,
+  startOpen = false,
+  overflowOverride,
+  fixedHeight = "300px",
+  children,
+  showFade = "false",
+  isBeta = false,
```
PatFawbertMills, line 27: Suggestion: I am conscious of Accordion growing into a bit of a beast if we add modifiers using this approach. Personally I would prefer a prop like `headContent` of type React.Node and then we can pass in any label we want, whether that be one that says "Beta" or "New" or whatever we want in future.

src/components/blocks/SearchFilters.tsx
```diff
@@ -1,4 +1,4 @@
-import { useEffect, useState, useMemo } from "react";
+import { useEffect, useState, useMemo, useCallback } from "react";
```
PatFawbertMills, line 1: We don't appear to be using `useCallback` below unless I missed it?

src/pages/search/index.tsx
```diff
@@ -2,7 +2,7 @@ import { AnimatePresence, motion } from "framer-motion";
 import { GetServerSideProps, InferGetServerSidePropsType } from "next";
 import { useRouter } from "next/router";
 import { ParsedUrlQueryInput } from "querystring";
-import { useEffect, useRef, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
```
PatFawbertMills, line 5: We don't appear to be using `useCallback` below unless I missed it?

src/components/blocks/SearchFilters.tsx
katybaulch: Ugh you're right, apologies! In Python my IDE gets rid of unused imports for me - need to find an equivalent for TS. Good spot!

katybaulch: Done

jamesgorrie:
@katybaulch you got any screenshots?

src/pages/search/index.tsx
```diff
@@ -524,6 +540,8 @@ const Search: InferGetServerSidePropsType<typeof getServerSideProps> = ({ theme,
         onCancelClick={() => setShowCSVDownloadPopup(false)}
         onConfirmClick={() => handleDownloadCsvClick()}
       />
+      {/* This is here in the short term for us to test features flags with our cache settings */}
```
jamesgorrie, line 543: suggestion: remove this

src/pages/search/index.tsx
```diff
@@ -532,14 +550,27 @@ export default Search;
...
   try {
     themeConfig = await readConfigFile(theme);
   } catch (error) {}
 
+  let conceptsData: TConcept[];
+  try {
+    const client = new ApiClient(process.env.CONCEPTS_API_URL);
+    const conceptsV1 = featureFlags["concepts-v1"];
+    if (conceptsV1) {
+      const { data: returnedData } = await client.get(`/concepts/search?limit=10000&q=`);
```
jamesgorrie, line 566: 🚤 10000 for the 🏅

src/components/accordian/Accordian.tsx
katybaulch: DOne!

jamesgorrie (commented):
Quick Q around how we're using the query string params

src/utils/buildSearchQuery.ts
```diff
@@ -52,23 +53,14 @@ export default function buildSearchQuery(
...
-        }))
-      : [{ name: "name", value: conceptFilters }];
+  if (routerQuery[QUERY_PARAMS.concept_id]) {
+    const conceptIds = routerQuery[QUERY_PARAMS.concept_id];
+    concept_filters.ids = Array.isArray(conceptIds) ? conceptIds : [conceptIds];
+  }
+
+  if (routerQuery[QUERY_PARAMS.concept_name]) {
+    const conceptNames = routerQuery[QUERY_PARAMS.concept_name];
+    concept_filters.names = Array.isArray(conceptNames) ? conceptNames : [conceptNames];
```
jamesgorrie, line 63: How does this work? This is the shape required by the search API to query for docs with concepts.

Are we conflating the shape of that API and the new new concepts API? Unless I am missing where we transform this out into that shape?

src/pages/search/index.tsx
katybaulch: Done

src/constants/queryParams.ts
```diff
@@ -23,6 +23,6 @@ export const QUERY_PARAMS = {
   // Reports
   author_type: "at",
   // Pass through
-  "concept_filters.id": "cfi",
-  "concept_filters.name": "cfn",
+  concept_id: "cfi",
+  concept_name: "cfn",
```
jamesgorrie, line 27: This might need renaming back too

src/utils/buildSearchQuery.ts
katybaulch: This was a mistake, accidental commit!

src/constants/queryParams.ts
jamesgorrie: Ah, got it - we don't need to.

Every layer of transformation we add, we expect our future selves to remember we made, and what is often harder to work out in a year, why we made it.

src/types/types.ts
```diff
@@ -6,6 +6,11 @@ export type TSearchKeywordFilters = {
   countries?: string[];
 };
 
+export type TSearchConceptFilters = {
```
jamesgorrie, line 9: q: Where is this used?

PR #386, opened 2025-02-26
APP-226: Add Icon component

jesskelsall:
# What's changed

Adds an `<Icon />` component which encapsulates all SVG icon usage in the app. This replaces the existing icon exports (e.g. `<CloseIcon />`) but keeps the majority of its source code.

- Added `Icon` component.
- Overhauled `svg/Icons.tsx`:
  - `IconName` string union type of all possible icon names.
  - `allIcons` object with each icon React component function.
- Replaced all icon usage with `Icon`.
- Added Storybook stories for `Icon`, including a custom grid showing all icons at once.

Screenshot 2025-02-26 at 11 18 43

It is possible in the future that we will want to replace all of these icons with a third party library. I still think this work is valuable right now, and doubly so if it progresses that conversation & decision.

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


src/components/Pill.tsx
```diff
@@ -1,4 +1,4 @@
-import { CloseIcon } from "./svg/Icons";
+import { Icon } from "./icon/Icon";
```
jamesgorrie, line 1: tangent: where do we use relative vs aliased import paths?

src/components/icon/Icon.tsx
```diff
@@ -0,0 +1,10 @@
+import { allIcons, IconName, IconProps } from "@components/svg/Icons";
+
+interface IconComponentProps extends IconProps {
+  name: IconName;
+}
+
+/**
+ * Renders an SVG icon from our internal icons library. All icons are SVG markup as JSX, allowing us to set currentColor to colour them.
+ */
+export const Icon = ({ name, ...iconProps }: IconComponentProps) => allIcons[name](iconProps);
```
jamesgorrie, line 10: thought: Would we want to deprecate the use of `svg/Icon` - as a thought experiment, as a new developer to the codebase, I am not sure I would know which to use?

src/components/Pill.tsx
jesskelsall: Aliased paths are used outside of the aliased directory. In this case you're seeing a relative path because it's one component importing another, rather than an unrelated file/page importing a component.

src/components/icon/Icon.tsx
jesskelsall: Good point. I liked the idea of keeping this file because it separates content from execution, but it might be more wholesale to keep everything together in one place, particularly because the `Icon` implementation is so [elegantly] bare.

katybaulch:
Adds an `<Icon />` component which encapsulates all SVG icon usage in the app. This replaces the existing icon exports (e.g. `<CloseIcon />`) but keeps the majority of its source code.

- Added `Icon` component.
- Overhauled `svg/Icons.tsx`:
  - `IconName` string union type of all possible icon names.
  - `allIcons` object with each icon React component function.
- Replaced all icon usage with `Icon`.
- Added Storybook stories for `Icon`, including a custom grid showing all icons at once.

Screenshot 2025-02-26 at 11 18 43

It is possible in the future that we will want to replace all of these icons with a third party library. I still think this work is valuable right now, and doubly so if it progresses that conversation & decision.

PR #387, opened 2025-02-26
Feature/app 291 single concept and text search on document page

katybaulch:
# What's changed

Enable search & searching within a single concept in the document viewer. There are no designs for this yet, so I've gone with the path of least resistance to get it working & we can expand after that.

image
image
image

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


PatFawbertMills:
Question: how is this different to the existing search on document view? Is it in addition to a concept?

katybaulch:
> Question: how is this different to the existing search on document view? Is it in addition to a concept?

Correct. If the user types in a term, they'll get the regular results like they would have before. No change there.

Now, if the user clicks a single concept, they are also presented with a text box where they can search for text within the concept hits.

PatFawbertMills (approved):
Note for future self: lets review how we use the `useCallback` and `useMemo` hooks to achieve the functionality we need.

PatFawbertMills:
Had a catch up with Katy to discuss a rough plan for reviewing the code again when we are ready to get it out of feature flag mode.

PR #389, opened 2025-02-27
Posthog provider

jamesgorrie:
# What's changed
- adds the posthog provider and captures pageview

This is in an effort to remove this from Google Tag Manager and get all the benefits of keeping it in GitHub provides.

## Proposed version

- [x] Patch

Part of https://linear.app/climate-policy-radar/issue/APP-301/move-posthog-from-tag-manager-=-navigator-frontend


src/context/PostHogProvider.tsx
```diff
@@ -0,0 +1,84 @@
...
+    }
+  }, [pathname, searchParams, posthog]);
+
+  return null;
+}
+
+// Wrap this in Suspense to avoid the `useSearchParams` usage above
+// from de-opting the whole app into client-side rendering
+// See: https://nextjs.org/docs/messages/deopted-into-client-rendering
+export function SuspendedPostHogPageView() {
```
PatFawbertMills, line 35: Smart.

PatFawbertMills (approved):
Pumped for this.

Just to check: do we also need to update the GTM script code?

PR #391, opened 2025-02-27
fix: add push to ecr to merge_to_main

jamesgorrie:
# What's changed

## Proposed version

- [ ] Skip auto-tagging
- [ ] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


odrakes-cpr (approved):

.github/workflows/merge_to_main.yml
```diff
@@ -85,6 +85,7 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     needs: [code-quality, test, test-e2e]
+    environment: production
```
katybaulch, line 88: thought: perhaps we make this staging for now, so we have deploys straight to staging and can maintain more granularity on prod deploys? I'm a little out of the loop on whether we're confident enough pushing straight to prod, so forgive me if we're already there!

katybaulch: my gosh of course, sorry 🤦🏻 I didn't read the PR properly. All good my side!

katybaulch (approved):
lgtm, sorry for not reading your PR properly! I saw deploy and production and panicked!

PR #395, opened 2025-03-03
APP-288 document viewer

PatFawbertMills:
# What's changed
- Rewrite of the logic in the pdf reader to remove some of the silently failing promises.

The previous iteration was encountering race-conditions due to how the promises were being structured.

## Proposed version

- [ ] Skip auto-tagging
- [ ] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/src/components/EmbeddedPDF.tsx
+++ b/src/components/EmbeddedPDF.tsx
@@ -1,9 +1,13 @@
-import { useRef, useMemo, useEffect, useContext } from "react";
+import { useRef, useState, useMemo, useEffect, useContext } from "react";
 import Script from "next/script";
-import { TDocumentPage, TPassage } from "@types";
-import usePDFPreview from "@hooks/usePDFPreview";
-import Loader from "./Loader";
+
 import { AdobeContext } from "@context/AdobeContext";
+
+import usePDFPreview from "@hooks/usePDFPreview";
+
+import Loader from "./Loader";
+
+import { TDocumentPage, TPassage } from "@types";
 
 type TProps = {
   document: TDocumentPage;
@@ -13,25 +17,32 @@ type TProps = {
 };
 
 const EmbeddedPDF = ({ document, documentPassageMatches = [], passageIndex = null, startingPassageIndex = 0 }: TProps) => {
+  const isLoading = useState(true);
   const containerRef = useRef(null);
   const adobeKey = useContext(AdobeContext);
 
-  const pdfPreview = usePDFPreview(document, documentPassageMatches, adobeKey);
+  const pdfPreview = usePDFPreview(document, adobeKey);
 
-  // Ensure the instance of the PDF client is not reset on render
+  // Ensure the instance of the PDF client is not reset on re-render
   // otherwise we lose the ability to interact with the pdf
   // eslint-disable-next-line react-hooks/exhaustive-deps
-  const { createPDFClient, passageIndexChangeHandler } = useMemo(() => pdfPreview, [document, documentPassageMatches, adobeKey]);
+  const { createPDFClient, passageIndexChangeHandler, documentMatchesChangeHandler } = useMemo(() => pdfPreview, [document, adobeKey]);
 
   useEffect(() => {
     if (containerRef?.current !== null) {
-      createPDFClient(startingPassageIndex);
+      createPDFClient();
     }
-  }, [containerRef, document, createPDFClient, startingPassageIndex]);
+  }, [containerRef, document, createPDFClient]);
 
   useEffect(() => {
-    passageIndexChangeHandler(passageIndex);
-  }, [passageIndexChangeHandler, passageIndex]);
+    passageIndexChangeHandler(passageIndex, documentPassageMatches);
+    /* trunk-ignore(eslint/react-hooks/exhaustive-deps) */
+  }, [passageIndexChangeHandler, passageIndex, JSON.stringify(documentPassageMatches)]);
+
+  useEffect(() => {
+    documentMatchesChangeHandler(documentPassageMatches, startingPassageIndex);
+    /* trunk-ignore(eslint/react-hooks/exhaustive-deps) */
+  }, [documentMatchesChangeHandler, JSON.stringify(documentPassageMatches)]);
 
   return (
     <>
@@ -48,4 +59,5 @@ const EmbeddedPDF = ({ document, documentPassageMatches = [], passageIndex = nul
     </>
   );
 };
+
 export default EmbeddedPDF;
--- a/src/hooks/usePDFPreview.ts
+++ b/src/hooks/usePDFPreview.ts
@@ -1,6 +1,5 @@
 import ViewSDKClient from "@api/pdf";
 import { TPassage, TDocumentPage } from "@types";
-import { PDF_SCROLL_DELAY } from "@constants/document";
 
 function generateHighlights(document: TDocumentPage, documentPassageMatches: TPassage[]) {
   const date = new Date();
@@ -54,7 +53,7 @@ function generateHighlights(document: TDocumentPage, documentPassageMatches: TPa
   });
 }
 
-export default function usePDFPreview(physicalDocument: TDocumentPage, documentPassageMatches: TPassage[], adobeKey: string) {
+export default function usePDFPreview(physicalDocument: TDocumentPage, adobeKey: string) {
   const viewerConfig = {
     showDownloadPDF: false,
     showPrintPDF: false,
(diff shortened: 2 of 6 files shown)


src/components/EmbeddedPDF.tsx
```diff
@@ -13,25 +17,32 @@ type TProps = {
...
+      createPDFClient();
     }
-  }, [containerRef, document, createPDFClient, startingPassageIndex]);
+  }, [containerRef, document, createPDFClient]);
 
   useEffect(() => {
-    passageIndexChangeHandler(passageIndex);
-  }, [passageIndexChangeHandler, passageIndex]);
+    passageIndexChangeHandler(passageIndex, documentPassageMatches);
+    /* trunk-ignore(eslint/react-hooks/exhaustive-deps) */
```
jamesgorrie, line 39: suggestion: Could we leave a description as to why we do this? I assume it's because we pass a string of the object, rather than the object? 

Why is that?

src/components/EmbeddedPDF.tsx
```diff
@@ -13,25 +17,32 @@ type TProps = {
...
+  }, [passageIndexChangeHandler, passageIndex, JSON.stringify(documentPassageMatches)]);
+
+  useEffect(() => {
+    documentMatchesChangeHandler(documentPassageMatches, startingPassageIndex);
+    /* trunk-ignore(eslint/react-hooks/exhaustive-deps) */
+  }, [documentMatchesChangeHandler, JSON.stringify(documentPassageMatches)]);
```
jamesgorrie, line 45: question: assuming we should actually have `startingPassageIndex` as a dep?

src/hooks/usePDFPreview.ts
```diff
@@ -67,59 +66,58 @@ export default function usePDFPreview(physicalDocument: TDocumentPage, documentP
 
   const annotationConfig = { showToolbar: false, showCommentsPanel: false, downloadWithAnnotations: true, printWithAnnotations: true };
 
-  let viewSDKClient = null;
-  let embedApi = null;
-  let annotationManagerApi = null;
+  let viewSDKClient_CACHE = null;
```
jamesgorrie, line 69: thought: A more JS/React idiomatic way of describing this would be memoised.

src/hooks/usePDFPreview.ts
```diff
@@ -67,59 +66,58 @@ export default function usePDFPreview(physicalDocument: TDocumentPage, documentP
...
+      viewSDKClient = viewSDKClient_CACHE;
+    } else {
+      viewSDKClient = new ViewSDKClient();
+      viewSDKClient_CACHE = viewSDKClient;
+    }
+
+    await viewSDKClient.ready();
+    const adobeViewer = await viewSDKClient.previewFile(physicalDocument, adobeKey, "pdf-div", viewerConfig);
+    const embedApi = await adobeViewer.getAPIs();
+    embedApi_CACHE = embedApi;
```
jamesgorrie, line 84: There's no check on this memo - so is it needed?

suggestion: add check or remove the memo.

src/hooks/usePDFPreview.ts
```diff
@@ -67,59 +66,58 @@ export default function usePDFPreview(physicalDocument: TDocumentPage, documentP
...
-        });
-        if (onLoadCallback) {
-          onLoadCallback();
-        }
-      });
-    });
+  const createPDFClient = async () => {
+    let viewSDKClient = null;
+    if (viewSDKClient_CACHE) {
+      viewSDKClient = viewSDKClient_CACHE;
```
jamesgorrie, line 75: suggestion: given the return type of this method - could we just memoise that? e.g.

```ts
let apisMemo: { embedApi: EmbedApi, annotationManagerApi: AnnotationManagerApi };

if (apisMemo) {
  return apisMemo
} else {
  const embedApi = await adobeViewer.getAPIs();
  const annotationManagerApi = await adobeViewer.getAnnotationManager();

  return { embedApi, annotationManagerApi }
}
```

src/hooks/usePDFPreview.ts
```diff
@@ -67,59 +66,58 @@ export default function usePDFPreview(physicalDocument: TDocumentPage, documentP
...
-            // annotationManager.addAnnotations(generateHighlights(document, documentPassageMatches));
-            documentMatchesChangeHandler(documentPassageMatches);
-          }
-        });
-        if (onLoadCallback) {
-          onLoadCallback();
-        }
-      });
-    });
+  const createPDFClient = async () => {
```
jamesgorrie, line 72: suggestion: would a name like `getAdobeApis` be more true to the return type?

src/hooks/usePDFPreview.ts
```diff
@@ -67,59 +66,58 @@ export default function usePDFPreview(physicalDocument: TDocumentPage, documentP
...
+    annotationManagerApi.setConfig(annotationConfig);
+    return {
+      embedApi,
+      annotationManagerApi,
+    };
   };
 
-  const passageIndexChangeHandler = (passageIndex: number) => {
+  const passageIndexChangeHandler = async (passageIndex: number, documentPassageMatches: TPassage[]) => {
+    let embedApi = embedApi_CACHE;
```
jamesgorrie, line 94: thought: I think leaving the caching of the APIs to the above method would simplify this a bit by letting that method deal with the memo exclusively. Anything calling that method would just need to `await geAdobeApis`

src/pages/search/index.tsx
jamesgorrie, line 1: suggestion: I'm not sure these changes are related to this PR - could we add them as a separate? I'm not 100% sure what they're doing, and it's muddying the water a bit.

src/utils/getMatchedPassagesFromFamiy.ts
```diff
@@ -0,0 +1,15 @@
+import { TDocumentPage, TMatchedFamily, TPassage } from "@types";
+
+export const getMatchedPassagesFromSearch = (families: TMatchedFamily[], document: TDocumentPage): [TPassage[], number] => {
+  let passageMatches: TPassage[] = [];
+  let totalNoOfMatches = 0;
+  families.forEach((family) => {
+    family.family_documents.forEach((cacheDoc) => {
+      if (document.slug === cacheDoc.document_slug) {
+        passageMatches.push(...cacheDoc.document_passage_matches);
+        totalNoOfMatches = family.total_passage_hits;
```
jamesgorrie, line 10: question: should this just be `=` and not an additive operator?

src/utils/getMatchedPassagesFromFamiy.ts
```diff
@@ -0,0 +1,15 @@
+import { TDocumentPage, TMatchedFamily, TPassage } from "@types";
+
+export const getMatchedPassagesFromSearch = (families: TMatchedFamily[], document: TDocumentPage): [TPassage[], number] => {
+  let passageMatches: TPassage[] = [];
+  let totalNoOfMatches = 0;
+  families.forEach((family) => {
+    family.family_documents.forEach((cacheDoc) => {
```
jamesgorrie, line 7: question: why is this called `cacheDoc`?

jamesgorrie (commented):
Looking very tidy! 

Couple questions on the return type and naming of `createPDFClient` as well as seeking a little clarity on how we're managing the deps for the `useEffects`

src/hooks/usePDFPreview.ts
PatFawbertMills: Critically, it also calls the `previewFile` promise method, which initialises the PDF into the HTML container on the page. The Adobe Apis can only then be accessed from the return of that promise.

In general I agree some of the names in this file could be more descriptive though, I'll review.

src/pages/search/index.tsx
PatFawbertMills: For some context, after I realised we can easily do query string updates without causing next to reload the page I had a look at the search page. The UX feels a bit nicer using this method of only reloading the search, moving the user to the top vs. reloading the page. Happy to move to another PR though.

katybaulch (approved):
lgtm, amazing work! 🍪

src/pages/search/index.tsx
PatFawbertMills: For clarity, this line `router.push({ query: router.query }, undefined, { shallow: true });`

Is to do with shallow routing, which is how I think this functionality should be done.

Source: https://nextjs.org/docs/pages/building-your-application/routing/linking-and-navigating#shallow-routing

src/utils/getMatchedPassagesFromFamiy.ts
PatFawbertMills: Not sure, renamed.

src/utils/getMatchedPassagesFromFamiy.ts
PatFawbertMills: Yes, we only want to use the value provided for the given doc we are currently viewing.

src/pages/search/index.tsx
jamesgorrie: I still think these should be part of separate PRs.

- It makes things easier to track and describe
- makes this PR clearer and easier to review
- Makes it easier to revert changes

I'm basically paraphrasing [this document](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/getting-started/helping-others-review-your-changes).

src/utils/getMatchedPassagesFromFamiy.ts
```diff
@@ -0,0 +1,15 @@
+import { TDocumentPage, TMatchedFamily, TPassage } from "@types";
+
+export const getMatchedPassagesFromSearch = (families: TMatchedFamily[], document: TDocumentPage): [TPassage[], number] => {
+  let passageMatches: TPassage[] = [];
+  let totalNoOfMatches = 0;
+  families.forEach((family) => {
+    family.family_documents.forEach((doc) => {
```
jamesgorrie, line 7:
```suggestion
    family.family_documents.forEach((doc) => {
      totalNoOfMatches = family.total_passage_hits;
```

src/utils/getMatchedPassagesFromFamiy.ts
jamesgorrie: Ah - then would it make more sense [here](https://github.com/climatepolicyradar/navigator-frontend/pull/395/files#r1979044770)?

jamesgorrie: Actually scrap my whole thought - I get what this is doing. It's finding the document 
in the family and then hoisting `family.total_passage_hits` on that match.

src/hooks/usePDFPreview.ts
jamesgorrie: > it also calls the previewFile promise method, which initialises the PDF into the HTML container on the page. The Adobe Apis can only then be accessed from the return of that promise.

This feels like exactly what a method called `getAdobeApis` should encapsulate and obfuscate from anyone using this API.

PatFawbertMills: Where this gets interesting and a bit trickier, is that I want to have access to the APIs but without re-running the `previewFile` method, because in the browser this is what triggers the document loader to re-load from scratch.

I feel like reloading is acceptable UX for the highlighting (mainly because of the out-of-memory issue), but for the page change (which relies upon the `embedApi` this is not acceptable).

jamesgorrie (approved):
🚤

src/hooks/usePDFPreview.ts
PatFawbertMills: The issue I've run into here is that if I want to access either API, I have to reinitialise the adobe viewer from scratch. Which would mean a reload of the PDF reader in the browser.

So for now I've decided, I'll memo (save in memory) the viewerApi, as we use this to change page - **within the context of a annotated session**. That important distinction is that if I memoize it forever, the page change doesn't work after reloading the PDF reader the first time, because the method is trying to access a previously instantiated pdf reader. 

I could well be getting something mixed up here! But I think that makes sense?

PatFawbertMills: embedApi === viewerApi just as a heads up. I've decided to try to match the adobe naming convention and improve readability.

src/hooks/usePDFPreview.ts
PatFawbertMills: This hasn't worked because when we get the out of memory error the annotation refuses to work at all. So (similar to the comment below), I have decided to memoize the viewerApi **only in the context of a given annotated session** / or no annotation technically counts as a session.

src/hooks/usePDFPreview.ts
PatFawbertMills: Don't disagree - ideally we wouldn't have this memory issue on the highlighter and then we could memoize the APIs

src/components/EmbeddedPDF.tsx
PatFawbertMills: Yes and done

PR #396, opened 2025-03-04
fix: group concepts by 'other' root concept

jamesgorrie:
# What's changed
- adds a "Other" root concept
- adds any concepts that aren't a subconcept of a root concept to that concept

This is because of the way we get he data - currently `parent_concept` is the immediate parent. We would need to traverse up until we hit a "root concept" to be able to get this, but this would be very inefficient given our current seupt.

as per the comment, an example
- indigenous people (Q684)
- is subconcept of marginalised ethnicity (Q676)
- is subconcept of marginalised people (Q1170)
- is subconcept of impacted group (Q672)

The current data structure doesn't allow for traversal of this in an easy manner. 

This is just an "easy" way to get the data on the page for now.

Maybe this could be part of the [concepts API](https://github.com/climatepolicyradar/concepts)?

## Proposed version

- [x] Patch

# Screenshots

Screenshot 2025-03-04 at 08 44 32

### DIFF
--- a/src/components/concepts/ConceptsPanel.tsx
+++ b/src/components/concepts/ConceptsPanel.tsx
@@ -23,15 +23,47 @@ export const ConceptsPanel = ({ rootConcepts, concepts, conceptCountsById, onCon
     [onConceptClick]
   );
 
+  /** This is a little hackery to deal with the fact that
+   * - we only groupBy a set of root concepts which are at the top of the concept chain
+   * - some concepts might be a couple steps from a root concept
+   *
+   * e.g.
+   * - indigenous people (Q684)
+   * - is subconcept of marginalised ethnicity (Q676)
+   * - is subconcept of marginalised people (Q1170)
+   * - is subconcept of impacted group (Q672)
+   *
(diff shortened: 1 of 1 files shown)


src/components/concepts/ConceptsPanel.tsx
```diff
@@ -23,15 +23,47 @@ export const ConceptsPanel = ({ rootConcepts, concepts, conceptCountsById, onCon
...
+   *
+   * e.g.
+   * - indigenous people (Q684)
+   * - is subconcept of marginalised ethnicity (Q676)
+   * - is subconcept of marginalised people (Q1170)
+   * - is subconcept of impacted group (Q672)
+   *
+   * We would like to show "indigenous people" as a subconcept of "impacted group" but the data doesn't allow it for now
+   */
+  const conceptsWithOtherRootConcept = concepts.map((concept) => {
```
katybaulch, line 38: Suggestion: there is a helper function 'processConceptas' we made at the beginning of this spike that we might be able to use here

PR #397, opened 2025-03-04
Revert "Revert "Posthog provider""

jamesgorrie:
Reverts climatepolicyradar/navigator-frontend#393

Essentially a re-implementation of: https://github.com/climatepolicyradar/navigator-frontend/pull/389

But we will only have 1 account where data goes, and we can then differentiate between data in that account. As recommended by [the docs](https://posthog.com/tutorials/nextjs-cookie-banner).

# Testing

Previously the feature-flags didn't work. This has been rectified by leaving the init script on that page with `opt_in_site_apps` set.

## Proposed version

- [x] Patch

Towards https://linear.app/climate-policy-radar/issue/APP-301/move-posthog-from-tag-manager-=-navigator-frontend


src/components/cookies/CookieConsent.tsx
```diff
@@ -13,10 +13,12 @@ import dynamic from "next/dynamic";
...
 declare let gtag: Function;
-
-export const CookieConsent = () => {
+type Props = {
+  onConsentChange: (consent: boolean) => void;
+};
+export const CookieConsent = ({ onConsentChange }: Props) => {
   const [hide, setHide] = useState(true);
-  const [enableAnalytics, setEnableAnalytics] = useState(false);
+  const [enableAnalytics, setEnableAnalytics] = useState<boolean>();
```
PatFawbertMills, line 21: Question: why not default to `false`? Curious to know if you ran into an issue with this approach.

jamesgorrie: Nice catch - the reason was for the hoisted `consent` value - I was using
- `undefined` = unknown
- `true | false` = known value

But that was conflating the two implementations. I've updated it so we just stick to `false` by default and `true` when we know.

PatFawbertMills (approved):
lgtm

jamesgorrie:
Thanks @PatFawbertMills.

Dropping into draft as I need to validate this approach with @alaniswright

PR #399, opened 2025-03-05
Add loading indication to pdf reader

PatFawbertMills:
# What's changed
Overall goal
- Adds a loading indicator to the PDF reader, which is listening to both the search status and the completion of the highlighting.

Additional work:
- Rewrote the annotation/highlighting approach to the PDF reader
- Annotations are not added en mass anymore, we 'register' the new passages with the adobe viewer, and when the page changes we add the annotations only for the new page in view
- This supports both clicking on a passage to navigate to a page, as well as the internal adobe page change.

This is a huge improvement in performance, plus we are no longer limited by the out-of-memory exception that Adobe was throwing when we were attempting to add 500 at a time.

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


src/components/EmbeddedPDF.tsx
```diff
@@ -26,15 +28,25 @@ const EmbeddedPDF = ({ document, documentPassageMatches = [], passageIndex = nul
...
-  }, [addAnnotations, documentPassageMatches, startingPassageIndex]);
+    setIsLoading(true);
+    registerPassages(documentPassageMatches, startingPassageIndex).finally(() => {
+      setIsLoading(false);
+    });
+  }, [registerPassages, documentPassageMatches, startingPassageIndex]);
+
+  useEffect(() => {
+    if (searchStatus === "loading" || isLoading) {
+      return setIsLoading(true);
```
jesskelsall, line 46: Question: Does setting `isLoading` to true when it is already true cause a re-render?

src/hooks/usePDFPreview.ts
```diff
@@ -108,20 +129,53 @@ export default function usePDFPreview(physicalDocument: TDocumentPage, adobeKey:
...
+      adobeViewer = newAdobeViewer;
+    }
+    if (!adobeViewer) {
+      return;
+    }
+    await adobeViewer.registerCallback(
+      window.AdobeDC.View.Enum.CallbackType.EVENT_LISTENER,
+      async (event: any) => {
+        if (event.type === "CURRENT_ACTIVE_PAGE") {
+          // console.log("Current active page has changed. Adding annotations for page: ", event.data.pageNumber);
```
jesskelsall, line 166: Minor ask: Remove unused debugging.

PatFawbertMills:
Closing for now, whilst I get back into it

PR #401, opened 2025-03-05
fix: use `recursive_subconcept_of` for grouping concepts

jamesgorrie:
# What's changed
- use `recursive_subconcept_of` for grouping concepts. This was introduced by @harrisonpim here https://github.com/climatepolicyradar/knowledge-graph/pull/281

### Drive-bys
- moves the whole abstraction of `fetchConcept` into `processConcepts`
- uses [`Promise.allSettled`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) in case any of the S3 calls fail

## screenshots

| before | after |
|---|---|
| Screenshot 2025-03-05 at 14 25 00 | Screenshot 2025-03-05 at 14 24 51 |

## Proposed version

- [x] Patch

### DIFF
--- a/src/components/concepts/ConceptsPanel.tsx
+++ b/src/components/concepts/ConceptsPanel.tsx
@@ -23,31 +23,28 @@ export const ConceptsPanel = ({ rootConcepts, concepts, conceptCountsById, onCon
     [onConceptClick]
   );
 
-  /** This is a little hackery to deal with the fact that
-   * - we only groupBy a set of root concepts which are at the top of the concept chain
-   * - some concepts might be a couple steps from a root concept
-   *
-   * e.g.
-   * - indigenous people (Q684)
-   * - is subconcept of marginalised ethnicity (Q676)
-   * - is subconcept of marginalised people (Q1170)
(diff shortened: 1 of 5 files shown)


odrakes-cpr (approved):

harrisonpim:
Not sure whether this is already on your radar but for what it's worth i think you're still missing a few concepts in those screenshots! eg [target (Q1651)](https://climatepolicyradar.wikibase.cloud/wiki/Item:Q1651) should be its own concept as well as a top-level group thing

katybaulch (approved):
lgtm, nice one!!

jamesgorrie:
> Not sure whether this is already on your radar but for what it's worth i think you're still missing a few concepts in those screenshots! eg [target (Q1651)](https://climatepolicyradar.wikibase.cloud/wiki/Item:Q1651) should be its own concept as well as a top-level group thing

@harrisonpim - as per a chat with @conordelahunty - we're just going to dupe them as the first label in that group. PR coming soon.

PR #402, opened 2025-03-05
Feature/app 310 getting concepts work production ready paired programming (part 1)

katybaulch:
# What's changed

Part 1 of getting concepts document viewer production ready.

Includes:
- Fix document viewer page rerendering
- Move concepts logic into concepts doc viewer
- Show loader for concept search so we have user feedback for the lag

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/src/components/documents/ConceptsDocumentViewer.tsx
+++ b/src/components/documents/ConceptsDocumentViewer.tsx
@@ -200,15 +200,14 @@ export const ConceptsDocumentViewer = ({
                 <div id="document-search" className="flex flex-col gap-2 md:pl-4">
                   {(selectedConcepts.length > 0 || initialQueryTerm) && (
                     <div className="flex gap-2">
-                      <Link className="capitalize hover:no-underline" href={`/documents/${document.slug}`} onClick={handleClearSearch}>
-                        <Button
-                          color="dark-dark"
-                          data-cy="view-document-viewer-concept"
-                          extraClasses="flex items-center text-[14px] font-normal pt-1 pb-1 bg-black text-white border-none"
-                        >
-                          ← Back
-                        </Button>
-                      </Link>
+                      <Button
+                        color="dark-dark"
+                        data-cy="view-document-viewer-concept"
+                        extraClasses="flex items-center text-[14px] font-normal pt-1 pb-1 bg-black text-white border-none"
+                        onClick={handleClearSearch}
+                      >
+                        ← Back
+                      </Button>
                     </div>
                   )}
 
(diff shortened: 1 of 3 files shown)


PatFawbertMills:
Tune in tomorrow folks, for part II! 📺

src/components/documents/ConceptsDocumentViewer.tsx
```diff
@@ -72,6 +68,28 @@ export const ConceptsDocumentViewer = ({
     totalNoOfMatches: 0,
   });
 
+  /** Concepts: WIP */
```
jamesgorrie, line 71: Why WIP?

src/pages/documents/[id].tsx
```diff
@@ -81,6 +79,7 @@ const DocumentPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({
   const exactMatchQuery = !!router.query[QUERY_PARAMS.exact_match];
   const startingPassage = Number(router.query.passage) || 0;
 
+  // TODO: Remove this once we have hard launched concepts in product.
```
Comment (jamesgorrie) [inline src/pages/documents/[id].tsx:82]: question: What exactly should we be removing here?

src/pages/documents/[id].tsx
```diff
@@ -281,9 +254,10 @@ const DocumentPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({
           handleViewSourceClick={handleViewSourceClick}
         />
 
-        <section className="flex-1 flex" id="document-viewer">
-          <FullWidth extraClasses="flex-1">
-            {concepts.length === 0 && (
+        {/* TODO: Remove this once we have hard launched concepts in product. */}
```
Comment (jamesgorrie) [inline src/pages/documents/[id].tsx:257]: answering own question: assuming this is the same as above - so basically remove the un-componented code. 🧹

jamesgorrie (approved):
This is great - nice follow up PR.

I'm going to a pedant here - the simplification of this component is great! And I'd argue the code was "production ready" and this PR

* does some simplification refactoring (always good)
* adds a loader

So the code was production ready, and an improvement.

I'm being a pedant for that reason - the initial code was an improvement! This was an improvement on that! And I am sure there will be more to come.

src/pages/documents/[id].tsx
Comment (katybaulch) [reply src/pages/documents/[id].tsx:257]: you got it!

src/pages/documents/[id].tsx
Comment (katybaulch) [reply src/pages/documents/[id].tsx:82]: uncomponented code - which we are able to do when we launched!

src/components/documents/ConceptsDocumentViewer.tsx
katybaulch: removed!

PatFawbertMills:
Happy to jump on a call to discuss, but this decision was partly as a reaction to feedback regarding the UX of the new concept search, as well as what I considered to be some over complication of the code which would affect us in the short term with maintainability and the ability to make incremental updates.

PR #403, opened 2025-03-05
Add root concept to self

jamesgorrie:
# What's changed
- adds the root concept to the list of concepts if it is available
- refactored this out to create a `groupBy` as the code was getting a bit hard to read, and it felt more fit-for-purpose

Follow up from #401 

- **Note the greenhouse gas**
- **Target is known to be weird as it is a sub_concept of Policy instrument, and a root concept itself**. We have spoken with @AnneIsARealProgrammerNow about this

| before | after |
|---|---|
| Screenshot 2025-03-05 at 17 07 08 | Screenshot 2025-03-05 at 17 06 53 |

## Proposed version

- [x] Patch

### DIFF
--- a/src/components/concepts/ConceptsPanel.tsx
+++ b/src/components/concepts/ConceptsPanel.tsx
@@ -35,13 +35,13 @@ export const ConceptsPanel = ({ rootConcepts, concepts, conceptCountsById, onCon
     has_subconcept: [],
   };
 
-  const rootConceptIdToConceptsMap: { [rootConceptId: string]: TConcept[] } = Object.groupBy(concepts, (concept) => {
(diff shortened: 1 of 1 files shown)


src/components/concepts/ConceptsPanel.tsx
```diff
@@ -52,17 +35,29 @@ export const ConceptsPanel = ({ rootConcepts, concepts, conceptCountsById, onCon
     has_subconcept: [],
   };
 
+  const conceptsGroupedByRootConcept: { [rootConceptId: string]: TConcept[] } = Object.groupBy(concepts, (concept) => {
```
PatFawbertMills, line 38: til: `groupBy`

PatFawbertMills (approved):
Reads cleaner 🧹

PR #404, opened 2025-03-05
APP-228: Apply Button component side-wide

jesskelsall:
# What's changed

This PR applies the new `Button` component everywhere it can reasonably be used.

- Replaced usage of the following in favour of using `Button` directly.
  - Components: `Close`, `FilterToggle`, `MatchesButton`, `SearchMatchesButton`. Several were only used in one instance (too early componentisation).
  - Ad-hoc `button` HTML tag usage.
  - Some button icon instances.
- Does not replace the following:
  - Any `button` HTML tags that looks too much like text.
  - Button icons that will likely get refactored later as part of their parent components (e.g. search bar).
- Implemented an equivalent `getButtonClasses` function for use in CCLW custom pages where a link needs to look like a button.
- Added support for icon buttons - round buttons that include only an icon.
- Simplified the markup for text + icon inside a button via the `content="both"` prop.
- Added hover and focus-within states.
- Tidied up the Button Storybook stories.
- Standardised how to display a close icon using `Button`.
- Moved `Button` and `Icon` into `src/components/atoms` (atomic design classification), which will also help separating old components from new component library components.

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


jesskelsall:
New Storybook stories for `Button`:

Screenshot 2025-03-05 at 17 31 28

src/components/atoms/button/Button.tsx
```diff
@@ -0,0 +1,102 @@
+import { useMemo } from "react";
```
PatFawbertMills, line 1: Shouldn't be as excited about this folder structure as I am 😮‍💨

src/components/atoms/button/Button.tsx
```diff
@@ -0,0 +1,102 @@
...
+  content?: "text" | "icon" | "both";
+  disabled?: boolean;
+  size?: "small" | "medium" | "large";
+  rounded?: boolean;
+  variant?: "solid" | "faded" | "outlined" | "ghost";
+}
+
+type ButtonProps = ButtonClassArgs & React.ButtonHTMLAttributes<HTMLButtonElement>;
+
+export const getButtonClasses = ({
```
PatFawbertMills, line 15: Nice.

PatFawbertMills (approved):
Top tier work.

PR #407, opened 2025-03-06
Fix: Button component type and usage

jesskelsall:
# What's changed

Very minor PR to address a few runtime errors due to the `Button` component's type and how string props are computed.

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/src/components/atoms/button/Button.tsx
+++ b/src/components/atoms/button/Button.tsx
@@ -10,7 +10,7 @@ interface ButtonClassArgs {
   variant?: "solid" | "faded" | "outlined" | "ghost";
 }
 
-type ButtonProps = ButtonClassArgs & React.ButtonHTMLAttributes<HTMLButtonElement>;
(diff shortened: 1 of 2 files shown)


src/components/atoms/button/Button.tsx
```diff
@@ -10,7 +10,7 @@ interface ButtonClassArgs {
   variant?: "solid" | "faded" | "outlined" | "ghost";
 }
 
-type ButtonProps = ButtonClassArgs & React.ButtonHTMLAttributes<HTMLButtonElement>;
+type ButtonProps = ButtonClassArgs & React.ComponentProps<"button">;
```
PatFawbertMills, line 13: TIL

PatFawbertMills (approved):

PR #413, opened 2025-03-10
Feature/app 342 Give user feedback when concept selected that isn't in current document

katybaulch:
# What's changed

When a user clicks on a concept on the family page, they are shown matches in the main document only rather than in any other supporting documents or the family itself. This PR provides the user with some feedback if they try to search for a concept in the main document that doesn't exist in that document.

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


src/components/documents/ConceptsDocumentViewer.tsx
```diff
@@ -68,61 +73,87 @@ export const ConceptsDocumentViewer = ({
...
     });
 
-    return Array.from(uniqueConceptMap.entries())
-      .map(([conceptKey, count]) => ({ conceptKey, count }))
-      .sort((a, b) => b.count - a.count);
-  }, [vespaFamilyData]);
+    return Array.from(uniqueConceptMap.values())
+      .map(({ concept, count }) => ({
+        ...concept,
+        count,
```
jamesgorrie, line 124: observation: As far as I can tell this piece of code is just recreating the same element?

src/components/documents/UnavailableConcepts.tsx
```diff
@@ -0,0 +1,33 @@
...
+    </div>
+    <p className="text-xl font-medium">Concept not in Document</p>
+    <p>Concept matching is currently in beta mode.</p>
+    <p>
+      Some concepts from your search are not be present in this document but might be present in{" "}
+      <Link
+        href={`/document/${familySlug}?${unavailableConcepts.map((concept) => `cfn=${concept}`).join("&")}`}
+        className="text-blue-600  hover:underline"
+      >
+        other documents in this entry
```
jamesgorrie, line 24: suggestion: `entry`? I know this is family, and I know we don't expose this - but is "result" potentially more in line with our breadcrumb?

Screenshot 2025-03-11 at 08 53 55

jamesgorrie (commented):
This looks good, but I wonder if it's going to be superseded by linking concepts to documents on the family page? CC @conordelahunty 

Just to clarify my reading of it though, a document page has:
* `documentConcepts`
* `unavailableConcepts` which `=` the diff of `familyConcepts` and `documentConcepts`
* And we only show this when there are zero results for a concept?

src/components/documents/UnavailableConcepts.tsx
```diff
@@ -0,0 +1,33 @@
...
+      <div className="rounded-full bg-blue-50 p-6 mb-2">
+        <Icon name="findInDoc" width="48" height="48" />
+      </div>
+    </div>
+    <p className="text-xl font-medium">Concept not in Document</p>
+    <p>Concept matching is currently in beta mode.</p>
+    <p>
+      Some concepts from your search are not be present in this document but might be present in{" "}
+      <Link
+        href={`/document/${familySlug}?${unavailableConcepts.map((concept) => `cfn=${concept}`).join("&")}`}
```
PatFawbertMills, line 21: For this use the `LinkWithQuery` component, which has the ability to pass the query object

src/components/documents/UnavailableConcepts.tsx
katybaulch: This text was confirmed by Alan, so will leave this alone :)

katybaulch:
> This looks good, but I wonder if it's going to be superseded by linking concepts to documents on the family page? CC @conordelahunty
> 
> Just to clarify my reading of it though, a document page has:
> 
> * `documentConcepts`
> * `unavailableConcepts` which `=` the diff of `familyConcepts` and `documentConcepts`
> * And we only show this when there are zero results for a concept?

Yeah this will be superceded by Conor's designs, but we need a way of feeding back to the user until then :)

src/components/documents/ConceptsDocumentViewer.tsx
katybaulch: is this not needed for the sort?

PR #415, opened 2025-03-11
Add import grouping and sorting via flat config

katybaulch:
# What's changed

- Migrated ESLint to v8+
- Migrated ESLint config file to flat config
- Fixed ESLint config errors
- Updated tsconfig.json `baseUrl` to `.`
- Updated tsconfig.json custom path aliases to be `@/*` so we can import sort our own files separately from third party libraries
- Fixed relative imports

## Proposed version

- [ ] Skip auto-tagging
- [ ] Patch
- [x] Minor version
- [ ] Major version

Notes:

  brackets and the `x`


package.json
```diff
@@ -25,6 +27,7 @@
     "framer-motion": "^11.11.7",
     "jwt-decode": "^4.0.0",
     "next": "15.1.4",
+    "next-auth": "^4.24.11",
```
jamesgorrie, line 30: question: What's this for?

eslint.config.mjs
```diff
@@ -0,0 +1,131 @@
...
+      "**/node_modules",
+      "**/*.config.js",
+      "**/*.config.mjs",
+      "**/eslint.config.mjs",
+      "**/vite.config.ts",
+      "tailwind.config.js",
+      "postcss.config.js",
+      "**/*.js",
+      ".storybook/preview.ts",
+      ".storybook/main.ts",
```
jamesgorrie, line 23: question: why not include these? I understand the build files like `dist`, but why files that we control?

eslint.config.mjs
```diff
@@ -0,0 +1,131 @@
...
+      "@typescript-eslint": fixupPluginRules(typescriptEslint),
+      "@next/next": nextPlugin,
+      import: fixupPluginRules(importPlugin),
+      "react-hooks": reactHooks,
+    },
+
+    languageOptions: {
+      globals: {
+        ...globals.browser,
+        ...globals.es2021,
```
jamesgorrie, line 38: suggestion: As we're using `esnext` in `tsconfig`, could we use the most modern version here i.e. `es20205`?

eslint.config.mjs
```diff
@@ -0,0 +1,131 @@
...
+        AudioWorkletGlobalScope: "readonly",
```
jamesgorrie, line 39: question: Where do we use `AudioWorklet`?

eslint.config.mjs
```diff
@@ -0,0 +1,131 @@
...
+    },
+
+    rules: {
+      // Extend Next.js core web vitals rules
+      ...nextPlugin.configs["core-web-vitals"].rules,
+
+      // TypeScript specific rules
+      "@typescript-eslint/no-explicit-any": "warn",
+      "@typescript-eslint/explicit-function-return-type": "off",
+      "@typescript-eslint/no-non-null-assertion": "off",
```
jamesgorrie, line 73: Do we have to leave this here given that our `tsconfig` allows for this?

I wouldn't want to be re-enforcing this as a choice, as i've seen it let through loads of bugs.

[See this PR spike for more info](https://github.com/climatepolicyradar/navigator-frontend/pull/377)

jamesgorrie (commented):
Just a couple questions on the main bits of the config.

Assuming our e2e / visual regression test will catch anything else from the larger refactor.

package.json
katybaulch: A missing import in the `next-auth.d.ts` file - the library is imported bu not in our package.json file

PR #420, opened 2025-03-11
APP-277: Add Input component

jesskelsall:
# What's changed

Adds a new `Input` component to Storybook, which will replace most `<input />` uses in my next PR.

- Add Base UI as a dependency.
- Add `components/molecules/input`
  - This is a molecule because it relies on `Icon` and eventually `Tooltip` (Base UI).
  - Built upon Base UI Input.
  - Effectively a div that wraps an input plus potentially an icon and/or clear button.
  - Some styling flexibility in terms of sizes and colours.
  - Props for overriding container and input styles.
- Add `Input` stories.
- Add `joinTailwindClasses` util function.
  - Handles repetitive string joins which are then used in `className` prop.
  - Tests.

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "build-storybook": "storybook build"
   },
   "dependencies": {
+    "@base-ui-components/react": "^1.0.0-alpha.6",
     "@hookform/resolvers": "^2.8.8",
     "axios": "^0.24.0",
     "babel-loader": "^9.1.0",
--- a/src/components/molecules/input/Input.tsx
+++ b/src/components/molecules/input/Input.tsx
@@ -31,8 +31,6 @@ export const Input = ({
   const classes = useMemo(() => {
     /* Colour */
 
-    // TODO may want to only support one colour
-
     const outlineColor = color === "brand" ? "focus-within:outline-border-brand" : "focus-within:outline-border-mono";
 
     /* Size */
--- a/src/components/molecules/input/Input.tsx
+++ b/src/components/molecules/input/Input.tsx
@@ -1,0 +1,99 @@
+import { Input as BaseInput } from "@base-ui-components/react";
+import { Icon, IconName, iconNames } from "@components/atoms/icon/Icon";
+import { joinTailwindClasses } from "@utils/tailwind";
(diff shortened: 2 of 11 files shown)


jesskelsall:
Variations of `Input`:

Screenshot 2025-03-11 at 10 03 34

src/styles/styles.css
```diff
@@ -88,6 +88,11 @@
   }
 }
 
+/* Necessary for Base UI overlays (tooltip, popover etc.) */
+.root {
+  isolation: isolate;
```
PatFawbertMills, line 93: TIL: new css property

src/components/molecules/input/Input.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+interface InputProps extends Omit<React.InputHTMLAttributes<HTMLInputElement>, "size"> {
+  clearable?: boolean;
+  containerClasses?: string;
+  color?: "brand" | "mono";
+  icon?: IconName | React.ReactNode;
+  iconOnLeft?: boolean;
+  inputClasses?: string;
+  size?: "small" | "medium" | "large";
+  valueSetter?: Dispatch<SetStateAction<string>>;
```
PatFawbertMills, line 14: Question: is this used only for the clear event?

package.json
```diff
@@ -17,6 +17,7 @@
     "build-storybook": "storybook build"
   },
   "dependencies": {
+    "@base-ui-components/react": "^1.0.0-alpha.6",
```
PatFawbertMills, line 20: Exciting!

src/components/molecules/input/Input.tsx
jesskelsall: Yes - I played with a few other ways of dealing with the clear event, but settled on this. If you want the clear button to work, you need to pass your `useState` setter function into this prop.

Having a less opinionated input that you can totally customise or use for real edge cases was preferable to a locked-down component with its own internal state.

PR #424, opened 2025-03-11
Update custom path for theme folders

katybaulch:
# What's changed

As part of import sorting PR, we are going to update our custom path aliases so that they have a preceding / e.g., @cclw becomes @/cclw. This is so we can set an import rule later on in ESLint that groups our custom files separately from 3rd party libraries.

## Proposed version

- [x] Skip auto-tagging
- [ ] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

### DIFF
--- a/scripts/build-custom-app.sh
+++ b/scripts/build-custom-app.sh
@@ -99,7 +99,7 @@ import React from "react";
 // generic layer component
 import Layout from "@/components/layouts/LandingPage";
 
-import Hero from "@${theme}/components/Hero";
+import Hero from "@/${theme}/components/Hero";
 
 type TProps = {
   handleSearchInput: (term: string, filter?: string, filterValue?: string) => void;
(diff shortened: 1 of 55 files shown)


jamesgorrie (approved):
Nice work.
I do sometimes find it useful to add work like this to the https://github.com/climatepolicyradar/navigator-frontend/blob/main/.git-blame-ignore-revs

[Here's a bit of docs on it](https://github.com/orgs/community/discussions/5033)

katybaulch:
> Nice work. I do sometimes find it useful to add work like this to the https://github.com/climatepolicyradar/navigator-frontend/blob/main/.git-blame-ignore-revs
> 
> [Here's a bit of docs on it](https://github.com/orgs/community/discussions/5033)

Yep - we've done this in the https://github.com/climatepolicyradar/navigator-frontend/pull/415 PR :)

PR #425, opened 2025-03-17
fix: add the type of buttons on CPR searchform

jamesgorrie:
# What's changed
As we don't specify the type of `button` on the search `form` - they are inferred as `type="submit"`.

This explicitly sets the `clear = button` and `search = submit`.

Fixes https://linear.app/climate-policy-radar/issue/APP-378/using-enter-to-search-in-the-homepage-is-not-working

## Proposed version

- [x] Patch


PatFawbertMills (approved):
LGTM

PR #427, opened 2025-03-18
fix: reinstate 'Passage matches click' event in posthog

jamesgorrie:
# What's changed
- adds `Passage matches click` event when someone does just that
- adds `index` to the event

This was previously done in tag manager via CSS selectors and the like - so wee're just reinstating it.

## Proposed version

- [x] Patch

### DIFF
--- a/src/components/PassageMatches.tsx
+++ b/src/components/PassageMatches.tsx
@@ -2,6 +2,7 @@ import { useEffect, useState } from "react";
 import Loader from "./Loader";
 import { TPassage } from "@/types";
 import { Icon } from "./atoms/icon/Icon";
+import { usePostHog } from "posthog-js/react";
 
 type TProps = {
   passages: TPassage[];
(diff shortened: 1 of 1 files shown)


src/components/PassageMatches.tsx
```diff
@@ -38,7 +41,13 @@ const PassageMatches = ({ passages, onClick, activeIndex, pageColour = "textDark
       ) : (
         <div className="my-5" id="passage-matches">
           {passages.map((item, index: number) => (
-            <div key={item.text_block_id} data-analytics="document-passage-result" id={`passage-${index}`} className="mb-2">
+            <div
+              key={item.text_block_id}
+              data-analytics="document-passage-result"
+              id={`passage-${index}`}
+              className="mb-2"
+              onClick={() => posthog.capture("Passage matches click", { index })}
```
PatFawbertMills, line 49: To confirm for my benefit: the click event on the div below bubbles up to this event too?

PatFawbertMills (approved):
LGTM - with comment just to confirm something

src/components/PassageMatches.tsx
jamesgorrie: @PatFawbertMills Yup - any child (assume this is what you mean "below") of this will trigger this event.

PR #429, opened 2025-03-18
fix: use prod env vars as default

jamesgorrie:
# What's changed
- uses production env variables as the default

I thought this had been done before, this makes for a more stable e2e testing environment and is low risk as the app is read only.

As per a discussion with @PatFawbertMills 

## Proposed version

- [x] Patch

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,22 +3,23 @@
 # API_URL=http://localhost:8888/api/v1
 
 # Production
-API_URL=https://app.dev.climatepolicyradar.org/api/v1
-TARGETS_URL=https://cpr-staging-targets-json-store.s3.eu-west-1.amazonaws.com
-CDN_URL=https://cdn.dev.climatepolicyradar.org
+API_URL=https://app.climatepolicyradar.org/api/v1
+TARGETS_URL=https://cpr-production-targets-json-store.s3.eu-west-1.amazonaws.com
+CDN_URL=https://cdn.climatepolicyradar.org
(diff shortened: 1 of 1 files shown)


.env.example
```diff
@@ -3,22 +3,23 @@
...
 
 # Production
-API_URL=https://app.dev.climatepolicyradar.org/api/v1
-TARGETS_URL=https://cpr-staging-targets-json-store.s3.eu-west-1.amazonaws.com
-CDN_URL=https://cdn.dev.climatepolicyradar.org
+API_URL=https://app.climatepolicyradar.org/api/v1
+TARGETS_URL=https://cpr-production-targets-json-store.s3.eu-west-1.amazonaws.com
+CDN_URL=https://cdn.climatepolicyradar.org
 CONCEPTS_API_URL=https://api.climatepolicyradar.org
+NEXT_PUBLIC_APP_TOKEN=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhbGxvd2VkX2NvcnBvcmFfaWRzIjpbXSwic3ViIjoiY3ByIiwiYXVkIjoibG9jYWxob3N0IiwiaXNzIjoiQ2xpbWF0ZSBQb2xpY3kgUmFkYXIiLCJleHAiOjIwNTc4NDMxNTAuMCwiaWF0IjoxNzQyMzEwMzUwfQ.8e1XUvVmcDwXZH1ouEH635EDjSfjp2Qy0xq-Ez4-wp4
```
jamesgorrie, line 10: These were generated specifically for localhost

PR #430, opened 2025-03-21
refactor: remove ci-cd action

jamesgorrie:
# What's changed
- removes the `ci-cd` action
- removes the need to specify versioning in PR template as we'll now deploy with hashes

## Why

- We are now supporting two ways of deploying which has it's overhead. This reduces that
- The ci-cd workflow is completely replicated in other parts of the pull_request or merge_to_main workflows
- Fewer race condition breaks where builds don't happen because of semver weirdness

This will be a workflow change when deploying `main` in the sense of you will need to specify the GitHub hash when deploying

Screenshot 2025-03-21 at 16 25 46

Some nice knock effects

- we will have fewer builds as the PR changes won't kick one off (Percy might be helped here)
- quicker builds
- potentially easier to start adding builds we can deploy PRs to staging

- [x] Patch


.github/pull_request_template.md
```diff
@@ -1,23 +1,5 @@
...
-
-Notes:
-
-- If none of these options are selected, auto-tagging will fail
-- Where multiple options are selected, the most senior option ticked will be
-  used -- e.g. Major > Minor > Patch
-- If you are selecting the version in the list above using the textbox, make
-  sure your selected option is marked `[x]` with no spaces in between the
-  brackets and the `x`
+## Screenshots?
```
odrakes-cpr, line 5: do we not want to keep this ? @jamesgorrie I might have missed this but is this now how we determine the version tag when creating a PR ?

jamesgorrie: Nope - there is no versioning on `navigator-frontend` after this.

There has been numerous discussions as what versioning means on this app, and there is no definitive answer as semver really only works for things that have clients e.g. APIs, libs etc.

[We currently tag images with the Git commit hash](https://github.com/climatepolicyradar/navigator-frontend/blob/main/.github/workflows/merge_to_main.yml#L118) - this is the reference we'll use for deploying. It's got the added benefit of being able to trace it through the system directly (we could with the release number, but indirectly).

Is there value that you think we're losing by moving to this system? The overhead of maintaining it is quite large.


By the by - If we are to continue using versioning for things like libraries and apis, I'd suggest we use something like [Towncrier](https://github.com/twisted/towncrier) or [changesets](https://github.com/changesets/changesets).

jamesgorrie: For clarity though - the inspiration for this change is to stop doing things two ways, and choose the hash route.

odrakes-cpr (approved):

.github/pull_request_template.md
odrakes-cpr: No happy with all that after your explanation.

PR #435, opened 2025-03-26
feat: add smoke-tests workflow

jamesgorrie:
# What's changed

- adds a smoke test that can be triggered from other events e.g. deploys, merges etc.

The first event I would like to spawn this from is `AppRunner: Update service: Succeeded`.
(@katybaulch hopefully allowing as to subscribe to `Rollback succeeded`)

**Why not make this a reusable job as we use it in a few places?**
It's such a small step that I think the flexibility out-weighs the reduction in code that this would present

## Why?

there are a few services that the frontend is dependant on and we would like to know if they've broken anything.

### DIFF
--- a/.github/workflows/smoke_tests.yml
+++ b/.github/workflows/smoke_tests.yml
@@ -1,0 +1,28 @@
+name: Smoke tests
+on: workflow_call
+# trunk-ignore(checkov/CKV2_GHA_1)
+permissions: write-all
+jobs:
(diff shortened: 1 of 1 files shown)


katybaulch (approved):
Nice one!!!

jamesgorrie:
> I'm so excited/grateful for this 🙏

This isn't yet the implementation of it, but should facilitate it

PR #439, opened 2025-04-02
fix: remove CPR.corpus.i00000001.n0000

jamesgorrie:
# What's changed
- removes `CPR.corpus.i00000001.n0000` from being shown on the frontend as the data as we are currently not happy with the data quality

I am not removing this from the app tokens as we might want to add it back in the next few weeks.


katybaulch (approved):
lgtm

PR #437, opened 2025-04-01
APP-386: Concept picker & Slideout

PatFawbertMills:
# What's changed
- New concept picker organism which encapsulates all the functionality related to the concepts as interactable checkboxes
- New `Slideout` menu component
- Various new generic core components, such as `Select`.

## Why?
- Users needed a way to see what concepts are available (have been classified)
- We need a way to generically handle concepts being selected so that we can replicate the functionality in other places, such as the document page.

## Screenshots
New filter on the search page
Screenshot 2025-04-01 at 16 57 16

Sorting by "Grouped"
Screenshot 2025-04-01 at 16 57 59

Select box open
Screenshot 2025-04-01 at 16 58 26

PR #441, opened 2025-04-02
APP-375: Concepts on a document page

jesskelsall:
# What's changed

Adds the new concept multi-picker to the document page so that passage results can include concepts in filtering. This work largely builds upon @PatFawbertMills's work on the search page and reuses as many existing components as possible.

- `ConceptsDocumentViewer` (document page):
  - Rearranged sidebar and fiddled with a lot of padding to make everything flow better.
  - Added syntactic sugar logic variables so the JSX reads easier.
  - Added `SideCol` to the left of the document viewer.
    - Uses `ConceptPicker` to provide concept filtering options.
    - Show/hide button for `ConceptPicker` when in single column layout.
- `ConceptPicker`
  - Moved title into a prop so each use can have its own.
  - Added `containerClasses` for further customisation.

**Includes #437 changes, so don't review this until it is merged!**

## Why?

Implements the following tickets (which are all essentially the same):
- APP-375 Concepts on a document page
- APP-292 Multiple concept search on document page
- APP-388 Document page utilising search in header and concept picker

## Screenshots?

Wide, no concepts selected:
Screenshot 2025-04-02 at 17 16 57

Wide, one concept selected:
Screenshot 2025-04-02 at 17 17 12

Narrow, concepts hidden:
Screenshot 2025-04-02 at 17 17 31

Narrow, concepts visible:
Screenshot 2025-04-02 at 17 17 39


src/components/organisms/ConceptPicker.tsx
```diff
@@ -64,20 +66,7 @@ const onConceptChange = (router: NextRouter, concept: TConcept) => {
   router.push({ query: query }, undefined, { shallow: true });
 };
 
-const getSelectedConcepts = (concepts: TConcept[], router: NextRouter) => {
```
jesskelsall, line 67: Removed as this function wasn't being called anywhere.

PatFawbertMills: Yeh I left it in case we ever wanted to use the logic of showing the selected concepts first. Conor pointed out that we will already be showing applied filters in two other places so probably not needed.

PatFawbertMills: Hoarder-mentality 😆

src/components/documents/ConceptsDocumentViewer.tsx
PatFawbertMills, line 1: Question/reservatio: if policy signed off on the concept work tomorrow, would we be confident switching over to using this component in replacement for the current document viewer?

I need to dig into it a little more but I am slightly concerned we have repeated and or overlapping functionality between the page view component and this viewer component - if that makes sense?

There are also a few approaches in this file that I am not totally happy with, such as the `useEffectOnce` workaround, that to me, may indicate a bigger issue.

I can appreciate you didn't create this file and are merely updating it.

PatFawbertMills (approved):
Good work @jesskelsall

jesskelsall:
Quick note: @PatFawbertMills has approved this on the understanding that there will be some iteration early next week, but for now the main priority is having something that the policy team can test.

PR #443, opened 2025-04-03
fix: make env vars static

jamesgorrie:
# What's changed
- makes the config static vs an API call

This reduces complexity i.e. `env => api => async call => app` to `env => app`.

It also reduces an API call, so latency, but I can't see much benefit in terms of performance.

Saying that CLS seems to decrease a bit, not sure why (the LCP is very API dependant, so skipping over that)

## before
Screenshot 2025-04-03 at 11 26 18

## after
Screenshot 2025-04-03 at 11 45 42


PatFawbertMills (approved):
I see what you mean now, looks great!

PR #444, opened 2025-04-03
Fix: Sticky NavBar hides sticky search filters on scroll

jesskelsall:
# What's changed

Instead of `NavBar` being sticky to the top of the page and hiding content, the page's overall scroll area now excludes the navbar. This means the vertical scrollbar starts after the navbar.

- `NavBar`:
  - Remove `position: sticky` and associated CSS classes.
  - Lock height to 72px or 128px for the smallest breakpoint.
- Where `NavBar` is used:
  - Wrap main and footer in a div which is the height of the screen minus the navbar height (breakpoint dependent).
  - Ensure the vertical scrollbar shows when needed.
- Other sites: applied individually. Does not affect home pages as `NavBar` is applied separately there.

## Why?

Fixes a bug where the sticky search filters would stick to the top of the screen, meaning they would be hidden by the sticky navbar.

## Screenshots?

Desktop:
Screenshot 2025-04-03 at 16 04 15

Desktop, showing filters sticking correctly after scrolling:
Screenshot 2025-04-03 at 16 05 18

Mobile:
Screenshot 2025-04-03 at 16 03 58


PatFawbertMills (approved):
👏 ran it locally too - love it

PR #449, opened 2025-04-07
fix: use categories from 'All' as default in searchQuery

jamesgorrie:
# What's changed
- sets "All" as the default category if one isn't supplied by the search query

## Why?
This ensures that the categories are attached to the search query => backend.

Fixes https://linear.app/climate-policy-radar/issue/APP-458/make-all-filter-use-an-explicit-list-of-corpus-ids


vitest.xml
katybaulch, line 1: This needs to be removed & added to the git ignore list.

PR #455, opened 2025-04-08
Fix/sticky header partial revert

jesskelsall:
# What's changed

- `<main>` and `Footer` are no longer wrapped by a scrollable div that occupies most of the page.
- `NavBar` is sticky again.
- Search page filters (which are sticky) have a `top` value equal to the height of `NavBar` so that when it sticks to the top, it doesn't get hidden underneath it.
- Tweak to CCLW logo positioning.
- Removed `scrollToSearchTop` from `useSearch` to prevent the breadcrumb/download bar from being hidden on page load.

## Why?

Bug fixes.

## Screenshots?

CCLW search showing the filters not being overlapped by NavBar:
Screenshot 2025-04-08 at 11 16 42

### DIFF
--- a/src/components/modals/Popup.tsx
+++ b/src/components/modals/Popup.tsx
@@ -4,7 +4,7 @@ import { Icon } from "@/components/atoms/icon/Icon";
 const Popup = ({ active, onCloseClick, children }) => {
   return (
     <div
-      className={`absolute bg-black/50 pointer-events-none top-0 left-0 p-2 transition duration-250 z-50 transform w-full h-full flex justify-center items-start items-center md:p-4 ${
+      className={`absolute bg-black/50 pointer-events-none top-0 left-0 p-2 transition duration-250 z-100 transform w-full h-full flex justify-center items-start items-center md:p-4 ${
         active ? "opacity-100 visible" : "opacity-0 invisible"
       }`}
     >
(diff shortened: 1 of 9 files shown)


src/components/modals/Popup.tsx
```diff
@@ -4,7 +4,7 @@ import { Icon } from "@/components/atoms/icon/Icon";
 const Popup = ({ active, onCloseClick, children }) => {
   return (
     <div
-      className={`absolute bg-black/50 pointer-events-none top-0 left-0 p-2 transition duration-250 z-50 transform w-full h-full flex justify-center items-start items-center md:p-4 ${
+      className={`absolute bg-black/50 pointer-events-none top-0 left-0 p-2 transition duration-250 z-100 transform w-full h-full flex justify-center items-start items-center md:p-4 ${
```
PatFawbertMills, line 7: We need more Z's 🧟 !

src/hooks/useSearch.ts
```diff
@@ -19,11 +19,6 @@ type TConfig = {
   };
 };
 
-const scrollToSearchTop = () => {
```
PatFawbertMills, line 22: Be gone foul beast

PatFawbertMills (approved):
Like it, makes a bit more sense

PR #456, opened 2025-04-08
App 448: concept picker mobile support

PatFawbertMills:
# What's changed
- Remove the double menu that we had previously to support mobile phone users

On mobile viewports (< 768px wide):
- Filters menu now full screen takeover and slides in and out from bottom of screen, over the page, rather than be part of the page
- Concepts sub menu slides out from that (TODO: work out a better way to support different behaviours on desktop and mobile)

## Why?
- Provide better UX for smaller and mobile viewports
- Remove duplicated code

## Screenshots?
Screenshot 2025-04-08 at 11 55 33


src/hooks/usePrevious.ts
```diff
@@ -0,0 +1,9 @@
+import { useEffect, useRef } from "react";
```
PatFawbertMills, line 1: Added as I thought I needed. Didn't need, but thought it might come in handy later.

src/components/atoms/SlideOut/SlideOut.tsx
```diff
@@ -33,20 +34,28 @@ export const SlideOut = ({ children, showCloseButton = true }: SlideOutProps) =>
 
   return (
     <AnimatePresence initial={false}>
-      {currentSlideOut === "concepts" && (
+      {currentSlideOut !== "" && (
```
jesskelsall, line 37: Praise: This could just be `currentSlideOut &&` (as empty string is falsy) but I prefer the explicit string comparison to give a sense of what this variable stores.

src/components/atoms/SlideOut/SlideOut.tsx
```diff
@@ -33,20 +34,28 @@ export const SlideOut = ({ children, showCloseButton = true }: SlideOutProps) =>
...
         <motion.div
           ref={ref}
           key="slideOut"
           initial={{ translateX: "-100px" }}
-          animate={{ opacity: 1, translateX: 0, transition: { duration: 0.35, ease: [0.04, 0.62, 0.23, 0.98] } }}
+          animate={{ opacity: 1, translateX: 0, transition: { duration: 0.25, ease: [0.04, 0.62, 0.23, 0.98] } }}
           exit={{ opacity: 0, transition: { duration: 0 } }}
-          className="absolute top-0 left-full h-full bg-white p-5 pb-10 min-w-[400px]"
+          className="absolute z-20 top-0 left-0 h-full bg-white p-5 pb-[70px] w-screen md:z-0 md:w-auto md:min-w-[400px] md:left-full md:pb-0"
```
jesskelsall, line 44: Thought for later: We should install a linting rule that orders Tailwind classes. Here `z-20` and `md:z-0` aren't next to each other which makes it harder to follow. Nothing to correct for now, just a thought.

src/pages/search/index.tsx
```diff
@@ -63,6 +62,7 @@ const Search: InferGetServerSidePropsType<typeof getServerSideProps> = ({ theme,
   const [showCSVDownloadPopup, setShowCSVDownloadPopup] = useState(false);
   const [showOptions, setShowOptions] = useState(false);
   const [drawerFamily, setDrawerFamily] = useState<boolean | number>(false);
+  const [searchChanged, setSearchChanged] = useState(false);
```
jesskelsall, line 65: Thought: In terms of variable naming, the word "dirty" usually describes a form or input whose value has updated from the starting value. "changed" also works, but just FYI.

src/pages/search/index.tsx
```diff
@@ -365,90 +374,34 @@ const Search: InferGetServerSidePropsType<typeof getServerSideProps> = ({ theme,
...
-                        handleFilterChange={handleFilterChange}
-                        handleYearChange={handleYearChange}
-                        handleRegionChange={handleRegionChange}
-                        handleClearSearch={handleClearSearch}
-                        handleDocumentCategoryClick={handleDocumentCategoryClick}
-                        featureFlags={featureFlags}
-                      />
-                    </div>
-                    <SlideOut showCloseButton={false}>
+                  <div className="sticky md:top-[72px] h-screen md:h-[calc(100vh-72px)] px-5 bg-white border-r border-gray-300 pt-5 pb-[70px] overflow-y-auto scrollbar-thumb-gray-200 scrollbar-thin scrollbar-track-white scrollbar-thumb-rounded-full hover:scrollbar-thumb-gray-500 md:pb-0">
```
jesskelsall, line 387: Question: Is `calc(100vh-72px)` still relevant here after my change made the navbar sticky again?

jesskelsall (approved):
Looks good! A few comments but nothing blocking.

src/components/atoms/SlideOut/SlideOut.tsx
PatFawbertMills: Open to ideas on this, personally I've found it easier to read to have the mobile classes first, then subsequent breakpoint overrides at the end, so that they are grouped by breakpoint, not individual style.

src/pages/search/index.tsx
PatFawbertMills: Yeh otherwise it was trying to stick underneath the header.

src/pages/search/index.tsx
PatFawbertMills: Good idea

PR #457, opened 2025-04-08
feat: add isomorphic runtime env vars

jamesgorrie:
# What's changed
- adds a way for us to access env vars, isomorphically

## Why?
- NextJS supports runtime env vars on the server
- NextJS supports build-time env vars on the browser vars via the `NEXT_PUBLIC_` interface
- NextJS **DOES NOT** support run time browser vars

**Option A** could be that we move the env vars to build time. This is against the "[Build once, deploy many](https://www.mikemcgarr.com/blog/build-once-deploy-many.html)" that docker gives us, and we use across all of our other services, so I wouldn't want to go against that.

**Option B** this option ✨ 
- supports runtime env vars
- supports them isomorphically

I've used a subset of the `process.env` variables so we have to explicitly add them to expose them as this will make them public.

I have not hijacked the `NEXT_PUBLIC_` prefix as that already has build in functionality.  

### Hasn't this been done before?
There is an implementation here: https://github.com/expatfile/next-runtime-env.

I've chosen not to go with this as the other options in this thread https://github.com/vercel/next.js/discussions/44628 are relatively simple, and we don't have top worry about updates etc.

Dependant on: https://github.com/climatepolicyradar/navigator-infra/pull/1018
Part of: https://linear.app/climate-policy-radar/issue/APP-468/tame-env-vars-in-navigator-frontend

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -1,0 +1,35 @@
+export const publicRuntimeEnvConfig = ["API_URL", "THEME", "BACKEND_API_URL", "BACKEND_API_TOKEN"] as const;
+
+/**
+ * This is a custom config that is makes runtime env variables available to the the client environment.
+ *
+ * As per the docs:
+ * > public environment variables will be inlined into the JavaScript bundle during next build
+ *
+ * This does not allow us to "build once, deploy everywhere" - which we would like to do.
+ *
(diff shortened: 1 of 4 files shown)


jamesgorrie:
COnverting to draft as i am pretty certain these aren't available at runtime.

PatFawbertMills (approved):
Really nice upgrade to how we manage env vars

PR #458, opened 2025-04-09
fix: remove server.js

jamesgorrie:
# What's changed
- removes  server.js

## Why?
Less code, less confusion and cognitive load.


PatFawbertMills (approved):
As I said in Slack: I have memories of removing this before but ran into surprising issues. 

Don't see why we need this now though.

PR #459, opened 2025-04-09
APP-438: default sort

PatFawbertMills:
# What's changed
- Default sort when the user is browsing documents in search is now date - descending
- Previously we default to passing `null` to the API, which was resulting in an unpredictable "random" set of results

## Why?
- More predictable and easier to understand for the user
- Removes the confusing & vague "default" sort order option

### DIFF
--- a/src/utils/buildSearchQuery.ts
+++ b/src/utils/buildSearchQuery.ts
@@ -29,20 +29,26 @@ export default function buildSearchQuery(
     query.query_string = routerQuery[QUERY_PARAMS.query_string]?.toString();
   }
 
-  if (routerQuery[QUERY_PARAMS.exact_match]) {
-    query.exact_match = routerQuery[QUERY_PARAMS.exact_match] === "true";
-  }
-
-  if (routerQuery[QUERY_PARAMS.offset]) {
-    query.offset = Number(routerQuery[QUERY_PARAMS.offset]);
+  if (routerQuery[QUERY_PARAMS.sort_order]) {
+    query.sort_order = routerQuery[QUERY_PARAMS.sort_order]?.toString();
   }
 
   if (routerQuery[QUERY_PARAMS.sort_field]) {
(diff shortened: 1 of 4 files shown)


src/utils/buildSearchQuery.ts
```diff
@@ -29,20 +29,26 @@ export default function buildSearchQuery(
...
+    query.sort_order = "desc";
+    query.sort_field = "date";
+  }
+
+  if (routerQuery[QUERY_PARAMS.exact_match]) {
+    query.exact_match = routerQuery[QUERY_PARAMS.exact_match] === "true";
+  }
+
+  if (routerQuery[QUERY_PARAMS.offset]) {
+    query.offset = Number(routerQuery[QUERY_PARAMS.offset]);
```
jamesgorrie, line 51: [There's a `buildSearchQuery.test`](https://github.com/climatepolicyradar/navigator-frontend/blob/main/src/utils/buildSearchQuery.test.ts) where we can capture these decisions and ensure we don't regress on them.

src/utils/buildSearchQuery.ts
```diff
@@ -29,20 +29,26 @@ export default function buildSearchQuery(
...
+    query.sort_order = routerQuery[QUERY_PARAMS.sort_order]?.toString();
   }
 
   if (routerQuery[QUERY_PARAMS.sort_field]) {
     query.sort_field = routerQuery[QUERY_PARAMS.sort_field]?.toString();
   }
 
-  if (routerQuery[QUERY_PARAMS.sort_order]) {
-    query.sort_order = routerQuery[QUERY_PARAMS.sort_order]?.toString();
+  // If no sort order is provided, and we are in "browse" mode, we want to set the sort order to latest date "date:desc"
```
jamesgorrie, line 40: what is "browse" mode?

PatFawbertMills: No search query

PR #461, opened 2025-04-09
APP 470: unfccc filters

PatFawbertMills:
# What's changed
- Adding 2 new filters for UNFCCC category: author type and type
- Add new feature flag called "unfccc-filters" which must be switched on in order to see them
- Support new `_document.type` taxonomy in `QUERY_PARAM` and the api query builder

## Why?
- So that policy can test new filter functionality

### DIFF
--- a/src/components/blocks/FilterOptions.tsx
+++ b/src/components/blocks/FilterOptions.tsx
@@ -1,6 +1,6 @@
 import { useState } from "react";
 import { ParsedUrlQuery } from "querystring";
-import { find as Lfind, get as Lget } from "lodash";
+import { get } from "lodash";
 
 import { InputCheck } from "@/components/forms/Checkbox";
 import { InputRadio } from "@/components/forms/Radio";
@@ -11,7 +11,7 @@ import { TCorpusTypeDictionary, TThemeConfig, TThemeConfigFilter } from "@/types
 import { TextInput } from "@/components/forms/TextInput";
 
 const getTaxonomyAllowedValues = (corporaKey: string, taxonomyKey: string, corpus_types: TCorpusTypeDictionary) => {
-  const allowedValues = Lget(corpus_types[corporaKey].taxonomy, taxonomyKey)?.allowed_values || [];
+  const allowedValues = get(corpus_types[corporaKey].taxonomy, taxonomyKey)?.allowed_values || [];
 
   return allowedValues;
 };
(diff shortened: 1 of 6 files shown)


src/components/blocks/FilterOptions.tsx
```diff
@@ -10,7 +11,7 @@ import { TCorpusTypeDictionary, TThemeConfig, TThemeConfigFilter } from "@/types
 import { TextInput } from "@/components/forms/TextInput";
 
 const getTaxonomyAllowedValues = (corporaKey: string, taxonomyKey: string, corpus_types: TCorpusTypeDictionary) => {
-  const allowedValues = corpus_types[corporaKey].taxonomy[taxonomyKey]?.allowed_values;
+  const allowedValues = get(corpus_types[corporaKey].taxonomy, taxonomyKey)?.allowed_values || [];
```
jamesgorrie, line 14: pretty certain we can do this with standard JS with something like
```ts
(corpus_types?.[corporaKey]?.taxonomy ?? taxonomyKey)?.allowed_values ?? []
```

src/utils/buildSearchQuery.ts
```diff
@@ -37,7 +37,7 @@ export default function buildSearchQuery(
     query.sort_field = routerQuery[QUERY_PARAMS.sort_field]?.toString();
   }
 
-  // If no sort order is provided, and we are in "browse" mode, we want to set the sort order to latest date "date:desc"
+  // If no sort order is provided, and we are in "browse" mode (i.e. no search term), we want to set the sort order to latest date "date:desc"
```
jamesgorrie, line 40: I know I asked previously, I do think "browse" adds more complexity to understanding simply not having a `q` set.

jamesgorrie (approved):
Nice - the pattern seems to be working quite well.

src/components/blocks/FilterOptions.tsx
jamesgorrie: Actually I've got the logic above wrong, so maybe ignore this

PatFawbertMills: Yeh, the issue is with nested taxonomies, for example for "types" I want the getter to find the key _within_ the `_documents` object.

src/utils/buildSearchQuery.ts
PatFawbertMills: That's fair, with this comment I am trying to give some context around the _why_ because it is clearer from the code of the what

PR #462, opened 2025-04-09
fix: remove NEXT_PUBLIC_APP_TOKEN env var that isn't public

jamesgorrie:
# What's changed
- removes the deprecated `NEXT_PUBLIC_APP_TOKEN` in favour of `BACKEND_API_TOKEN`
- renames `API_URL` => `BACKEND_API_URL` for consistency

As per https://github.com/climatepolicyradar/navigator-infra/pull/1018 - which has deployed these values to the AppRunner instances - so we're good.

## Why?
Mainly a snag that tripped my up, and has the potential to break the app as the NEXT_PUBLIC_* is not actually available. You can read all about it here https://github.com/climatepolicyradar/navigator-frontend/pull/457


src/pages/api/env.ts
```diff
@@ -4,8 +4,8 @@ export default function handler(req: NextApiRequest, res: NextApiResponse) {
   res.status(200).json({
     env: {
       theme: process.env.THEME,
-      api_url: process.env.API_URL,
-      app_token: process.env.NEXT_PUBLIC_APP_TOKEN,
+      api_url: process.env.BACKEND_API_URL,
```
jamesgorrie, line 7: I've purposefully not renamed `api_url` as we're going to remove this soon.

PatFawbertMills (approved):
LGTM new names make more sense

odrakes-cpr (approved):
lgtm

PR #464, opened 2025-04-10
APP-476: menu on cclw homepage issue

PatFawbertMills:
# What's changed
- Always show bg on cclw header and if we only show menu force height to smaller version


jesskelsall (approved):
Looks good. 🙂

PR #465, opened 2025-04-14
Add timeout to the geography check due to API speed

PatFawbertMills:
# What's changed
- Add timeout for the country lookup in case the /geographies endpoint is slow

## Why?
- e2e tests are flakey because they are dependant on API response times

### DIFF
--- a/tests/search.spec.ts
+++ b/tests/search.spec.ts
@@ -41,7 +41,7 @@ test("search", async ({ page }) => {
   await expect(firstSearchResult.locator('[data-cy="family-metadata-category"]')).toBeVisible();
   await expect(firstSearchResult.locator('[data-cy="family-metadata-year"]')).toBeVisible();
   await expect(firstSearchResult.locator('[data-cy="family-description"]')).toBeVisible();
-  await expect(firstSearchResult.locator('[data-cy="country-link"]')).toBeVisible();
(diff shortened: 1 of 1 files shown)


tests/search.spec.ts
```diff
@@ -41,7 +41,7 @@ test("search", async ({ page }) => {
   await expect(firstSearchResult.locator('[data-cy="family-metadata-category"]')).toBeVisible();
   await expect(firstSearchResult.locator('[data-cy="family-metadata-year"]')).toBeVisible();
   await expect(firstSearchResult.locator('[data-cy="family-description"]')).toBeVisible();
-  await expect(firstSearchResult.locator('[data-cy="country-link"]')).toBeVisible();
+  await expect(firstSearchResult.locator('[data-cy="country-link"]')).toBeVisible({ timeout: 10000 });
```
jamesgorrie, line 44: question: should we be doing this? I know it fixes the test - but it's a cludge to a wider issue. 

If we are just looking to get the e2e tests less stable, could we perhaps create an issue and TODO to tidy this up in future?

jamesgorrie (approved):
Let's merge this to get the system up and running, but I think we should look at a more robust solution of "How do we deal with these slow API responses"

PR #467, opened 2025-04-14
Bug - playwright timeouts

PatFawbertMills:
# What's changed
- Rather than having timeouts for the visible check, wait for the response from the `searches` endpoint

## Why?
- This seems like a more logical approach considering it is the endpoint which delays the checks from passing


jamesgorrie (approved):
Small suggestion - great solution.

PR #472, opened 2025-04-17
APP-518: Litigation thin slice - search displays results

jesskelsall:
# What's changed

1. `CountryLinks` doesn't render an empty React node when it fails to find a name for a geographical subdivision.
2. `extractNestedData` is no longer hard coded to only extract 2 levels, but instead is recursive and returns an array of `n` levelled nodes.

## Why?

1. Prevents a double bullet (• •) as part of the country display on a search result.
2. Preparation for displaying the geographical subdivision on the new search result card, family page, and geography page. As each of these involves redesigns, I feel it's sufficient to stop here for now.

## Screenshots?

Litigation results' country, year, and category are only separated by single bullets:
Screenshot 2025-04-17 at 11 12 47

### DIFF
--- a/.size-limit.js
+++ b/.size-limit.js
@@ -2,6 +2,6 @@ module.exports = [
   {
     name: "Next static files",
     path: [".next/static/chunks/**/*.js"],
-    limit: "375 KB",
+    limit: "400 KB",
   },
 ];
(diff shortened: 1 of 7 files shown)


PatFawbertMills:
Looks like you'll need to bump up the size limit in `/size-limit.js`

PatFawbertMills (approved):
Nice work Jess

PR #473, opened 2025-04-17
APP-524: Litigation thin slice: family page displays litigation and its documents

jesskelsall:
# What's changed

- "Other documents in this entry" document heading doesn't include "other" when there are no main documents.
- Removed padding on this heading to fix the tooltip alignment.
- Notes heading is hidden if there is no corpus note to show.

## Why?

Making litigation play nice with the existing CPR website.

## Screenshots?

Screenshot 2025-04-17 at 14 03 35


PatFawbertMills (approved):
LGTM - nice that we don't have tons to do

PR #474, opened 2025-04-22
APP-527: Litigation thin slice: document page displays case documents

jesskelsall:
# What's changed

- `ConceptsDocumentViewer` only renders the concept picker panel if there is at least 1 concept present in the document.
  - We are expecting this to eventually be true of all documents, but right now litigation data on staging has not been ran against classifiers, and this is a good edge case to cover off anyway.
- Added the first litigation corpus ID to the Litigation and All categories.
  - Essential to get passage matches back from the API.

## Why?

Ensuring the existing CPR documents page plays nice with litigation documents.

## Screenshots?

Litigation document showing hidden concepts panel, PDF viewer, and passage matches:
Screenshot 2025-04-22 at 13 42 08


annaCPR (approved):
lgtm

PR #476, opened 2025-04-23
Fix: add Litigation corpus to CI env vars

jesskelsall:
# What's changed

Adds the Litigation corpus to the `BACKEND_API_TOKEN` used by the CI process.

## Why?

Necessary for litigation work that includes making API requests with this corpus.


jamesgorrie (commented):
Fab. Note to self `Academic.corpus.Litigation.n0000` = `Litigation`

jamesgorrie (approved):

PR #477, opened 2025-04-28
Wrap wait promises in a promise.all

PatFawbertMills:
# What's changed
- Investigation into the issues surrounding the e2e tests flakeyness
- For now, in the short term, not preventing merging of PRs based on the failing e2e tests - now a greater emphasis on manually checking work

## Why?
- We are currently being blocked by the e2e tests failing
- We think this is probably caused by the issues at the API / back-end layers upstream from the front-end


jesskelsall (approved):
Happy with this pragmatic approach for now in order to keep things moving.

PR #478, opened 2025-04-29
Add SortableTable component

jesskelsall:
# What's changed

- Add `SortableTable` component:
  - Uses the `ColumnKey` generic to define which keys the row object should have (string union).
  - `columns` prop is an array of column objects used to define the column order and which ones are sortable.
  - `rows` prop is an array of data objects with keys that match the `ColumnKey` generic.
    - Can use simple data types such as strings or numbers.
    - Can define separate display values (`ReactNode`) and sort values for more complex data in cells.
  - Column sorting:
    - Defaults to the order that rows are specified.
    - Can be provided a default sort through the `defaultSort` prop.
    - Controlled through internal state.
    - Click the sort buttons in the table header cells (that have sorting enabled) to sort by or reverse the sort order.
- Add `SortTable` stories.

TODO:

- [ ] All styling.
- [ ] Better sort icons - which is active, which direction, which orders can be sorted by.

## Why?

Confirmation by Conor that the family/document page revamp will include tabulated entries.

## Screenshots?

*[once ready for PR]*


PatFawbertMills (approved):
LGTM

jesskelsall:
Superseded by #608

PR #481, opened 2025-04-29
canPreview logic checks whether the CDN object is a PDF

kdutia:
# What's changed

`canPreview` logic checks whether the CDN object (rather than the document at its source URL) is a PDF.

Also adds a missing variable to `.env.example` that's needed for dev.

## Why?

As we have started converting other document types to PDF and also want to preview these, this changes the `canPreview` logic to check the file extension of the CDN object (and that it exists). See [related ADR](https://github.com/climatepolicyradar/architecture-hub/pull/17/files) for more context.

### Family with an HTML and a PDF 

image

### HTML with text

Soon these will be converted to PDF, which will test whether this code handles that correctly. I'll start with one document so the impact of any short-term issues are minimised.

image

### PDF

image


PatFawbertMills (approved):
LGTM

PR #486, opened 2025-04-30
Fix: Litigation results only behind the feature flag

jesskelsall:
# What's changed

A few changes to protect our environments from accidentally showing litigation results before the project is finished.

- Removed litigation corpus ID from "All" category in CPR config.
- Config can be changed dynamically depending on feature flags:
  - With litigation feature flag: add litigation to "All" category.
  - Without litigation feature flag: remove "Litigation" category.

This PR doesn't lock down litigation data entirely - it is still possible to manually enter URLs for litigation families/documents, though this can't really be achieved without looking at the API response, which is beyond a reasonable user experience.

## Why?

- We don't want users browsing the CPR website to find litigation cases/documents until they are ready for launch.
- The old method of hiding search results on the Litigation category no longer works as there is a lot on the page that shows this is smoke and mirrors (search results count, CSV download, API response in network tab).


jesskelsall:
Closed for a simpler approach.

PR #487, opened 2025-04-30
Climate Case Chart app setup

jesskelsall:
# What's changed

- Adds a new theme: `ccc` (Climate Case Chart).
- Taken as a copy of CCLW + CPR's `config.json`.
- Added litigation corpus ID to config.
- Homepage, header, and footer stripped back and made purple as a temporary clear distinction from other apps.
- Much of the theme code & components is expected to change later when we have designs and content.

## Why?

We need to spin up a CCC staging app.

## Screenshots?

Home page:
Screenshot 2025-04-30 at 16 00 29

Search:
Screenshot 2025-04-30 at 16 00 44


PatFawbertMills (approved):
CCC is alive! 🚀

katybaulch (approved):
lgtm - i'll add the infra bits!

PR #490, opened 2025-05-01
Revert 471 feature/display litigation in search

PatFawbertMills:
# What's changed
- Reverts the change that was exposing litigation docs to users on the CPR version of the app

## Why?
- We have no intention of showcasing the documents there yet


annaCPR (approved):
lgtm

PR #493, opened 2025-05-01
APP-561: Add document passage sorting options

jesskelsall:
# What's changed

- Added `sp` query param for `sort_within_page` search API param.
- `sp` is treated as `false` by default (sort by relevance).
- Added support for this query param when performing searches.
- Updated `SearchSettings` component to render "Relevance" and "Document position" options on the document page.

## Why?

Following a recent demo, there is a strong need to change the default passage searching back to relevance, but also provide the user flexibility.

## Screenshots?

Document page with default passage sorting (by relevance):
Screenshot 2025-05-01 at 14 37 46

Search settings dropdown options:
Screenshot 2025-05-01 at 14 37 54

Document page with passages sorted by document position:
Screenshot 2025-05-01 at 14 38 06

### DIFF
--- a/src/components/filters/SearchSettings.tsx
+++ b/src/components/filters/SearchSettings.tsx
@@ -8,12 +8,13 @@ import { QUERY_PARAMS } from "@/constants/queryParams";
 import { sortOptions, sortOptionsBrowse } from "@/constants/sortOptions";
 
 type TProps = {
-  queryParams: ParsedUrlQuery;
-  handleSortClick?: (sortOption: string) => void;
+  extraClasses?: string;
+  handlePassagesClick?: (passagesOption: string) => void;
   handleSearchChange?: (key: string, value: string) => void;
+  handleSortClick?: (sortOption: string) => void;
+  queryParams: ParsedUrlQuery;
   setShowOptions?: (value: boolean) => void;
   settingsButtonRef?: MutableRefObject<any>;
-  extraClasses?: string;
 };
 
 const getCurrentSortChoice = (queryParams: ParsedUrlQuery, isBrowsing: boolean) => {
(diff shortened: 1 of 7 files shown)


PatFawbertMills:
Not blocking, but wonder if the settings menu would benefit from a label saying "Sort by" above the 2 options

src/components/filters/SearchSettings.tsx
```diff
@@ -34,27 +35,41 @@ const getCurrentSemanticSearchChoice = (queryParams: ParsedUrlQuery) => {
...
+  queryParams,
   setShowOptions,
   settingsButtonRef,
-  extraClasses = "",
 }: TProps) => {
   const searchOptionsRef = useRef(null);
   const [options, setOptions] = useState(sortOptions);
 
+  // no query string OR query string is empty
   const isBrowsing = !queryParams[QUERY_PARAMS.query_string] || queryParams[QUERY_PARAMS.query_string]?.toString().trim() === "";
```
jamesgorrie, line 55: note+praise: I've mentioned this before - but "browsing" feels very unclear in code (maybe good for user stories etc) - so nice to have the comment.

src/hooks/useSearch.ts
```diff
@@ -55,19 +55,20 @@ const useSearch = (query: TRouterQuery, familyId = "", documentId = "", runFresh
...
       sort_order: searchQuery.sort_order,
-      offset: searchQuery.offset,
-      family_ids: searchQuery.family_ids,
-      document_ids: searchQuery.document_ids,
-      continuation_tokens: searchQuery.continuation_tokens,
-      corpus_import_ids: searchQuery.corpus_import_ids,
-      concept_filters: searchQuery.concept_filters,
-      metadata: searchQuery.metadata,
+      sort_within_page: searchQuery.sort_within_page,
+      year_range: searchQuery.year_range,
```
jamesgorrie, line 71: note: These reordering refactors make it a little tricky to see what the actual change is.

I've found it useful to do those refactors as no-ops first, and then introduce the change, but also understanding we're linting as we go.

src/components/filters/SearchSettings.tsx
```diff
@@ -106,6 +121,24 @@ export const SearchSettings = ({
...
+                  onClick={(e) => handlePassagesOrderClick(e, "false")}
+                  isActive={getCurrentPassagesOrderChoice(queryParams) === false}
+                >
+                  Relevance
+                </SearchSettingsItem>
+                <SearchSettingsItem
+                  onClick={(e) => handlePassagesOrderClick(e, "true")}
+                  isActive={getCurrentPassagesOrderChoice(queryParams) === true}
+                >
+                  Document position
```
jamesgorrie, line 137: subjective+note: document position => position in document?

jesskelsall:
@PatFawbertMills probably, but this is in line with what's already on the search page.

jamesgorrie (approved):
Simple and quickly done! Nice.

src/constants/queryParams.ts
```diff
@@ -1,25 +1,26 @@
...
+  exact_match: "e",
+  offset: "o",
+  query_string: "q",
+  region: "r",
   sort_field: "sf",
   sort_order: "so",
-  offset: "o",
-  active_continuation_token: "act",
-  continuation_tokens: "cts",
+  sort_within_page: "sp",
```
PatFawbertMills, line 13: Question: this sort option is more of a passage level sort than explicitly a sort within a particular page? Would `passage_sort`, or `sort_passages` (keeps as sp) would read easier?

src/hooks/useSearch.ts
jesskelsall: Apologies, I'm trying to find the right balance with this as I also don't particularly like the PR diff noise it creates.

PatFawbertMills (approved):
LGTM couple of non-blocking comments.

src/constants/queryParams.ts
jesskelsall: I had tried to keep this the same as the API query param, but you're right, it's not intuitive. I've gone with `passages_by_position` as I'd like to keep it a boolean.

src/components/filters/SearchSettings.tsx
jesskelsall: What might be nice here is a bit of syntactic sugar around query params - a getter function that handles the nuance of "what if there's a param but it has a falsy value".

PR #497, opened 2025-05-06
APP 540 - concept picker update

PatFawbertMills:
# What's changed
- UI changes
- Some behaviour changes regarding opening accordion items when searching
- Fix bug on mobile

## Why?
- New design updates and UX improvements

## Screenshots?
- See linear ticket

### DIFF
--- a/src/components/FaqSection.tsx
+++ b/src/components/FaqSection.tsx
@@ -24,9 +24,9 @@ export const FaqSection = ({ title, faqs, accordionMaxHeight = "464px" }: TProps
       </Heading>
       <VerticalSpacing size="md" />
       <div className="text-content mb-14">
-        {faqs.map((faq, i) => (
+        {faqs.map((faq, index) => (
           <Fragment key={faq.title}>
-            <Accordian title={faq.title} headContent={faq.headContent ?? null} open={i === 0} fixedHeight={accordionMaxHeight}>
+            <Accordian title={faq.title} headContent={faq.headContent ?? null} open={index === 0} fixedHeight={accordionMaxHeight}>
               {faq.content}
             </Accordian>
             <hr />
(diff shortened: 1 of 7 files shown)


src/components/FaqSection.tsx
```diff
@@ -26,7 +26,7 @@ export const FaqSection = ({ title, faqs, accordionMaxHeight = "464px" }: TProps
       <div className="text-content mb-14">
         {faqs.map((faq, i) => (
           <Fragment key={faq.title}>
-            <Accordian title={faq.title} headContent={faq.headContent ?? null} startOpen={i === 0} fixedHeight={accordionMaxHeight}>
```
jesskelsall, line 29: Nitpick: `index` is much more descriptive than `i` at a glance.

src/components/accordian/Accordian.tsx
```diff
@@ -17,15 +18,19 @@ type TProps = {
 
 export const Accordian = ({
```
jesskelsall, line 19: Thought for later: It'd be nice to replace this entirely with Base UI's accordion when we get some time.

src/components/accordian/Accordian.tsx
```diff
@@ -17,15 +18,19 @@ type TProps = {
...
+  open = false,
   overflowOverride,
   fixedHeight = "300px",
   children,
   showFade = "false",
   headContent,
   ...props
 }: TProps) => {
-  const [isOpen, setIsOpen] = useState(startOpen);
+  const [isOpen, setIsOpen] = useState(open);
```
jesskelsall, line 29: Concern: If the open state is now being controlled by a prop, why is internal state needed if it will mirror the prop value?

katybaulch (approved):
lgtm

jesskelsall (approved):
Some comments to address as you see fit. Otherwise approved.

src/components/accordian/Accordian.tsx
PatFawbertMills: It isn't solely controlled by the prop, I've made this change so that we can optionally control the accordion from outside of itself. It still has the ability to control itself from within the component.

PR #498, opened 2025-05-07
Fix: Passage sorting added to documents page with concepts

jesskelsall:
# What's changed

- `ConceptsDocumentViewer` shows passage sorting options and reacts appropriately when it is changed.
- Added typing on internal component state.

## Why?

Accidentally missed off from #493


PatFawbertMills (approved):
lgtm

Note for future: review the use of `useReducer` in this component and evaluate if it is necessary. Can appreciate it is not your code Jess.

PR #502, opened 2025-05-08
Allow for more padding in the bottom of the nav menu on mobile

PatFawbertMills:
# What's changed
- On mobile devices the "close" or "apply" button is not visible. This is a big issue as it means mobile users cannot close the menu when using search
- I have applied a bottom padding on the navigation to alleviate this

Note: this needs to be improved upon in future
- This is quite a heavy-handed approach of just applying a lot of padding at the bottom

## Why?
- Mobile users are blocked from using search correctly

## Screenshots?
First one shows the issue (hidden button), second one shows the solution on an android device (button visible)

image
Screenshot 2025-05-08 at 16 44 49


jesskelsall (approved):
Definitely needs revisiting later but I agree that this will resolve the immediate issue.

PR #510, opened 2025-05-12
APP-541: Search labelling

jesskelsall:
# What's changed

- Revisions to `Popover` component:
  - Prefers passing title, description, and link props, but still supports throwing children at it.
    - Allows existing usage of Popover to be refactored (more DRY).
  - Top arrow exported from a utils file for future use with other Base UI components (Menu, Preview Card).
- Search page:
  - Added `Info` component to filters and search results.
  - Removed `NoOfResults` component (now redundant with wording changes).
  - Updated search settings from a Lucide icon to "Sort & Display" with open state UI.
- Document page:
  - Same treatment for "Sort & Display" (both inside and outside of feature flag).

## Why?

Satisfies [APP-566](https://linear.app/climate-policy-radar/issue/APP-566/concept-popover) and [APP-541](https://linear.app/climate-policy-radar/issue/APP-541/labelling).

## Screenshots?

Search page:
Screenshot 2025-05-12 at 09 23 28

Filters popover:
Screenshot 2025-05-12 at 09 23 40

Results popover:
Screenshot 2025-05-12 at 09 23 48

Search settings:
Screenshot 2025-05-12 at 09 23 55

Document page (Concepts feature flag):
Screenshot 2025-05-12 at 09 24 22


PatFawbertMills:
I was about to say "we do limit the results" but I prefer this way of presenting the data. Nice.

src/components/atoms/popover/Popover.tsx
```diff
@@ -1,29 +1,42 @@
...
+}
+
+interface PopoverChildrenProps extends PopoverGenericProps {
+  children: ReactNode;
+  title?: never;
+  description?: never;
+  link?: never;
+}
+
+type PopoverProps = PopoverElementProps | PopoverChildrenProps;
```
PatFawbertMills, line 37: Request: we decide between ourselves a naming convention for types and interfaces.

Personally I quite like the prefix approach so that one instantly knows what the nature of the thing is, e.g. `TTypeName`, `IInterfaceName`, or even `TProps` if it is only used in the component.

src/pages/documents/[id].tsx
```diff
@@ -244,8 +243,11 @@ const DocumentPage: InferGetServerSidePropsType<typeof getServerSideProps> = ({
                           )}
                         </div>
                         <div className="relative z-10 flex justify-center">
-                          <button className="p-2 text-textDark text-xl" onClick={() => setShowOptions(!showOptions)}>
-                            <LuSettings2 />
+                          <button
+                            className={`px-1 py-0.5 -mt-0.5 rounded-md text-sm text-text-primary font-normal ${showOptions ? "bg-surface-ui" : ""}`}
+                            onClick={() => setShowOptions(!showOptions)}
+                          >
+                            Sort & Display
```
Comment (PatFawbertMills) [inline src/pages/documents/[id].tsx:250]: Same question RE: ampersand as above

src/styles/theme.css
```diff
@@ -75,6 +75,7 @@
   --color-text-primary: #202020;
   --color-text-secondary: #505050;
   --color-text-tertiary: #707070;
+  --color-text-quaternary: #858585;
```
PatFawbertMills, line 78: TIL: the name for 4th of a thing

src/components/atoms/popover/Popover.tsx
jesskelsall: My usage here is purely out of habit, I agree that it'd be best to standardise.

Personally I don't see a ton of value in using a prefix. There are rarely cases where the type/interface distinction is useful for TypeScript in React, and IDEs make it easy to contextualise the type on hover.

That being said, the prefix is useful at a glance (if a bit ugly), and I like that it adds a bit more intentionality to deciding whether to use a type or interface for an object.

PR #512, opened 2025-05-12
Use lvh over screen for min-height

PatFawbertMills:
# What's changed


jesskelsall (approved):
Looks good to me, as long as you're confident this is the only place we're using `-screen` or `vh]`.

PR #513, opened 2025-05-13
APP-582: Standardise TypeScript type/interface naming

jesskelsall:
# What's changed

- All TypeScript types and interface names are prefixed with `T` and `I` respectively.
- React component argument objects are always interfaces where possible.
- Main React component exports in a file have an argument type called `Props`, therefore `IProps` or rarely `TProps`.

This is a very noisy PR without any functionality change.

This PR _doesn't_ enforce naming via linting as it would require importing an entire ESLint package for the sake of one rule ([typescript-eslint's naming-convention](https://typescript-eslint.io/rules/naming-convention/)). If we have the appetite for this, please challenge me during the review.

## Why?

- Making our preferred naming conventions stick.
- Strongly hinting to other contributors what the standard is by example.

### DIFF
--- a/src/components/Alert.tsx
+++ b/src/components/Alert.tsx
@@ -1,9 +1,9 @@
-type TProps = {
+interface IProps {
   message: string | React.ReactNode;
   icon?: React.ReactNode;
-};
+}
 
-export const Alert = ({ message, icon }: TProps) => {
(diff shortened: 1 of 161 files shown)


odrakes-cpr:
@jesskelsall  could we add this as a rule to our current linter ?

odrakes-cpr (approved):

PR #514, opened 2025-05-13
Fix: Remove offset query param on new search

jesskelsall:
# What's changed

- Removes the `o` search offset URL parameter when submitting a new search through the `NavSearch` component.

## Why?

- Parity with the original functionality prior to the search in navbar piece of work.
- If the new search has less results than the offset number, no results are displayed despite there being more than 0 results.


PatFawbertMills (approved):
Nice catch and find.

PR #519, opened 2025-05-22
Fix feature flagged FAQs

katybaulch:
# What's changed

- Added a new function to get the cookies from the document to fix the conditional rendering of concept feature flags.
- Use CPR as template for CCC FAQs instead of CCLW, as CCLW is the exception in terms of content and formatting

## Why?

Feature flagged FAQs were not showing as FAQs are read via our custom CPR [page].tsx rather than via NextJS, so server side props getting was not working.


PatFawbertMills:
Question: when we are getting the cookies, is it necessary to get all the cookies to then pass into the `getFeatureFlags(featureFlagsCookie)` method?

themes/ccc/pages/faq.tsx
```diff
@@ -1,100 +1,52 @@
...
+    The FAQs page is read in not by using Next.JS, but by our CPR specific page reading logic.
+    This means that we cannot fetch the feature flags directly by using the page context.
+    This function provides a means of working around this so we can conditionally display the
+    concept FAQs.
+
+    TODO: Remove this once we have hard launched concepts in product.
+  */
+  const [featureFlags, setFeatureFlags] = useState({});
+  async function getFeatureFlag() {
+    const featureFlagsCookie = getAllCookies();
```
PatFawbertMills, line 24: Question: as far as I can tell - this function is not returning only feature flag cookies, but all cookies currently saved for this domain. Is there intention to filter out the non-feature-flag cookies from this list?

katybaulch: Renamed for clarity :)

jamesgorrie (approved):
Looks good with a little naming tweaks

PR #526, opened 2025-05-28
Add auto deploys to staging & gated deploys to prod

katybaulch:
# What's changed

Add auto deploys to staging & gated deploys to prod

## Why?

I'm a chaos monkey.

### DIFF
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -1,0 +1,53 @@
+name: Deploy to production
+on:
+  workflow_dispatch:
+    inputs:
+      # This is a gated deploy workflow.
+      # User inputs are intentional and protected by:
+      # 1. Environment protection rules requiring approval
+      # 2. Limited theme choices
+      # 3. Only deploying existing images
+      # trunk-ignore(checkov/CKV_GHA_7)
+      theme:
+        description: Theme to deploy
+        type: choice
+        required: true
+        options:
+          - cpr
+          - cclw
+          - mcf
+          - ccc
+      sha:
+        description: Git SHA to deploy
+        type: string
+        required: true
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  deploy-production:
+    runs-on: ubuntu-latest
+    environment:
+      name: production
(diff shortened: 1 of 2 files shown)


PatFawbertMills (approved):
LGTM good work!

.github/workflows/deploy-production.yml
```diff
@@ -0,0 +1,53 @@
...
+        with:
+          role-to-assume: arn:aws:iam::${{ secrets.AWS_ACCOUNT_ID }}:role/navigator-new-frontend-github-actions
+          role-session-name: GitHub_to_AWS_via_FederatedOIDC
+          aws-region: eu-west-1
+
+      - name: Build, tag, and push to ECR with 'production' tag
+        id: build-image
+        env:
+          ECR_REGISTRY: ${{ secrets.DOCKER_REGISTRY }}
+          IMAGE_TAG: production
```
jamesgorrie, line 46: Is there a reason to use `production` and not `latest`?

.github/workflows/deploy-staging.yml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          role-to-assume: arn:aws:iam::${{ secrets.AWS_ACCOUNT_ID }}:role/navigator-new-frontend-github-actions
+          role-session-name: GitHub_to_AWS_via_FederatedOIDC
+          aws-region: eu-west-1
+
+      - name: Build, tag, and push to ECR with 'latest' tag
+        id: build-image
+        env:
+          ECR_REGISTRY: ${{ secrets.DOCKER_REGISTRY }}
+          IMAGE_TAG: staging
```
jamesgorrie, line 34: Same question as above, as we're pushing to the different AWS environments - could we just use `latest`?

Or are we pushing to the same environment?

.github/workflows/deploy-production.yml
katybaulch: image

.github/workflows/deploy-staging.yml
katybaulch: image

PR #533, opened 2025-05-29
APP-621: Enable us to phase key features onto different custom apps in stages

jesskelsall:
# What's changed

This PR adds app-level features via app config so that each app can have different features. This is a separate mechanism to feature flags, but they can be combined together to provide more control.

### Config Features

- Updated `TThemeConfig` to include a `features` key of known feature keys and required booleans for each.
- `knowledgeGraph` is the first theme config feature and is enabled for both CPR and CCLW.

### Determining if a feature is enabled

- Added `isFeatureEnabled` function which considers the intersection of feature flags and theme config features to determine if a feature is active (see table below).
- All permutations are tested as this needs to be rock solid.
- Added shorthand functions for each feature we currently have, allowing you to call a syntactic sugar function with minimal arguments, e.g. `isLitigationEnabled(featureFlags)`.

This table represents how `isFeatureEnabled` determines its output. Note that either of its two arguments are optional, so it supports an app feature without a feature flag and vice versa.

- Horizontal axis: app config feature
- Vertical axis: feature flag

|| `true` | `false` | `undefined` |
|---|---|---|---|
| `true` | `true` | `true` | `true` |
| `false` | `true` | `false` | `false` |
| `undefined` | `true` | `false` | `false` |

What does this mean in practice?
- If a feature flag and app feature are both undefined, there's no need to use this function.
- If the app feature is `true`, the feature flag is ignored and the feature is on.
- If the app feature is `false` and there is no feature flag, the feature is off.
- Otherwise use the feature flag value.

### Putting it into practice

- All existing instances of direct feature flag comparison has been replaced by the new shorthand functions!
- Some `getServerSideProps` async calls for feature flags and theme config has been standardised in service of this.
- Extra tidy up around typings and state defaults.

### Dynamic theme pages

As the `concepts-v1` feature flag is already in use on dynamic theme pages, these also needed addressing:
- Added `ThemePageFeaturesContext` to roll up the theme config and feature flags.
- Applied `ThemePageFeaturesContext` around the dynamic page component.
- Replaced feature flag loading within each page with context use and shorthand feature functions.

## Why?

- Satisfies [APP-621](https://linear.app/climate-policy-radar/issue/APP-621/enable-us-to-phase-key-features-onto-different-custom-apps-in-stages).
- Feature branching should be bulletproof and a doddle for anyone to use or understand.

### DIFF
--- a/src/utils/features.ts
+++ b/src/utils/features.ts
@@ -7,7 +7,7 @@ interface IArgs {
 
 // Determines if an app feature is on or off by combining app-level config and PostHog feature flags
 export const isFeatureEnabled = ({ configFeature, featureFlag }: IArgs): boolean => {
-  if (configFeature === undefined && featureFlag === undefined) return false; // Shouldn't be using this fn
+  if (configFeature === undefined && featureFlag === undefined) return true; // No feature control = on
   if (configFeature === false) return false; // Config feature off = always off
   if (configFeature === true && featureFlag === undefined) return true; // Config feature on = on
   return featureFlag === true; // Config feature on + feature flag = use feature flag
--- a/src/utils/features.ts
+++ b/src/utils/features.ts
@@ -1,0 +1,37 @@
+import { TFeatureFlags, TThemeConfig } from "@/types";
+
+interface IArgs {
+  configFeature?: boolean;
+  featureFlag?: boolean;
+}
+
+// Determines if an app feature is on or off by combining app-level config and PostHog feature flags
+export const isFeatureEnabled = ({ configFeature, featureFlag }: IArgs): boolean => {
+  if (configFeature === undefined && featureFlag === undefined) return false; // Shouldn't be using this fn
+  if (configFeature === false) return false; // Config feature off = always off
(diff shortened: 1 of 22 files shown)


src/utils/features.ts
```diff
@@ -0,0 +1,37 @@
+import { TFeatureFlags, TThemeConfig } from "@/types";
+
+interface IArgs {
+  configFeature?: boolean;
+  featureFlag?: boolean;
+}
+
+// Determines if an app feature is on or off by combining app-level config and PostHog feature flags
+export const isFeatureEnabled = ({ configFeature, featureFlag }: IArgs): boolean => {
+  if (configFeature === undefined && featureFlag === undefined) return false; // Shouldn't be using this fn
```
PatFawbertMills, line 10: As mentioned in call - I think here we maybe default to true? Because if it isn't in the config and there isn't a feature flag for it, makes me think its just a 'normal' feature?

PatFawbertMills (approved):
Commented regarding the defaulting to true but we can debate that once its in.

PR #537, opened 2025-06-02
Fix author type key name in config and query builder

PatFawbertMills:
# What's changed
- The author type was incorrectly named, it only worked on cpr because there was another author type field
- Now added additional checks when building the search query object to check against the selected corpus (optional param) to prevent this in future


katybaulch (approved):
🫡 let's get this in

PR #539, opened 2025-06-02
Update feature flag logic

PatFawbertMills:
# What's changed
- Remove ff check on unfccc filters
- REVERT check of ff logic from posthog provider
- Turn on KG feature for CPR & CCLW themes


themes/cpr/config.ts
```diff
@@ -348,7 +348,7 @@ const config: TThemeConfig = {
   ],
   documentCategories: ["All", "Laws", "Policies", "UNFCCC", "Litigation", "MCF", "Reports"],
   features: {
-    knowledgeGraph: false,
+    knowledgeGraph: true,
```
katybaulch, line 351: does this set KG to permanently on? or is this like an extra check against our feature flags?

PR #540, opened 2025-06-03
APP-619: Create a column-based layout paradigm

jesskelsall:
# What's changed

Adds new column layout components that uses CSS grid to arrange content horizontally.

- Add `Columns` component:
  - 4-column CSS grid layout.
  - Drops progressively to 1 column using newly defined breakpoints.
  - Also applies a new maximum content width, so no other layout-based component needs wrapping around this.
  - Stories
    - Illustrated how to use `grid-template-columns: subgrid` to have a multi-column element subdivide onto the same column grid.
- Add `SubColumns` component:
  - 3-column CSS grid layout.
  - Does not drop to any other column layout (this is the main grid's job).
  - Stories
    - Shows how this looks/works in practice in combination with `Columns`.

I expect that the specific measurements (breakpoints, gaps) will be subject to change following design review, but that's okay as this isn't used anywhere but Storybook yet.

## Why?

- Satisfies [APP-619](https://linear.app/climate-policy-radar/issue/APP-619/create-a-column-based-layout-paradigm)
- We need a straightforward way of arranging our pages into a 4x3 column layout, as per the new family page designs and the thinking behind them.

## Screenshots?

Columns:
Screenshot 2025-06-03 at 10 16 24

SubColumns:
Screenshot 2025-06-03 at 10 16 31


PatFawbertMills (approved):
Neat 🖊️

PR #542, opened 2025-06-03
feat: manual deploys to staging

jamesgorrie:
# What's changed
- adds the ability to manually build and deploy to staging

I need to get this merged into `main` to allow the `workflow_dispatch` test.

## Why?
So that we can test on `staging` and potentially allow for auto-deploys to `production`

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -6,6 +6,7 @@ on:
       - completed
     branches:
       - main
+  workflow_dispatch: {}
 
 permissions:
   id-token: write


katybaulch:
Does this potentially link into our idea of having a `dev` server as well as `staging` and `prod` - for developers to test against separately from the ideally more stable `staging` server?

jamesgorrie:
> Does this potentially link into our idea of having a dev server as well as staging and prod - for developers to test against separately from the ideally more stable staging server?

If we deem it useful, indeed!

**TL;DR;** it would be good to have merge-to-main deploy to `production`.

Currently though this the step we spoke about where
- if branch is up to date with `main` (already enforced in GitHub rules)
- if someone can test on `staging` pre-merge
- when we merge-to-main we can deploy to `production` as it's been tested on staging already

--- 

I think what it could potentially also tie into is something more specific rather than a new environment.

e.g.

- "deploy this branch" action
- spin up an apprunner with a branch build
- expose the apprunner URL
- test test test
- merge-to-main deploys to production

PR #544, opened 2025-06-03
App 635 document page double search

PatFawbertMills:
# What's changed
Prevent double calls to the `/searches` endpoint by:
- Moving search out of the `ConceptsDocumentViewer` and instead pass in from the document page where the existing search was taking place

- Removed a couple of instances of `useCallback` inspired by Kent (https://kentcdodds.com/blog/usememo-and-usecallback). I don't believe we need to preserve the identity of this function in between renders

## Why?
- I believe this was leading to race conditions which was displaying oddities to the user (showing a loader when the API call had finished, etc)
- Reduced duplication between document page and component


odrakes-cpr (approved):
code looks good

PR #547, opened 2025-06-04
Don't clear concepts when switching between categories

PatFawbertMills:
# What's changed
- Don't clear concepts when switching between categories

## Why?
- Users want to keep their selection


katybaulch (approved):
beautiful

PR #548, opened 2025-06-04
Add two bits of business logic to force sort to be within a page

PatFawbertMills:
# What's changed
Setting the default sort order to "sort_within_page" for a specific search with conditions:
- no search query
- within a document view
- with concepts/classifiers

⚠️ This is known and accepted business logic / technical debt

## Why?
See ticket but essentially product and users are unhappy with the default logic when only classifiers are being applied to search

### DIFF
--- a/src/components/filters/SearchSettings.tsx
+++ b/src/components/filters/SearchSettings.tsx
@@ -37,6 +37,16 @@ const getCurrentSemanticSearchChoice = (queryParams: ParsedUrlQuery) => {
 };
 
 const getCurrentPassagesOrderChoice = (queryParams: ParsedUrlQuery) => {
+  // TODO: remove this
+  // Setting the default sort order to "sort_within_page" for a specific search with conditions:
+  // - no search query
+  // - with concepts/classifiers
+  if (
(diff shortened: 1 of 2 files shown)


src/components/filters/SearchSettings.tsx
```diff
@@ -37,6 +37,16 @@ const getCurrentSemanticSearchChoice = (queryParams: ParsedUrlQuery) => {
 };
 
 const getCurrentPassagesOrderChoice = (queryParams: ParsedUrlQuery) => {
+  // TODO: remove this
```
katybaulch, line 40: Ticket number for the TODO??

src/utils/buildSearchQuery.ts
```diff
@@ -50,6 +50,20 @@ export default function buildSearchQuery(
     query.sort_within_page = routerQuery[QUERY_PARAMS.passages_by_position] === "true";
   }
 
+  // TODO: remove this
```
katybaulch, line 53: Same as above

katybaulch (approved):
lgtm

PR #554, opened 2025-06-09
Remove loader over the pdf preview

PatFawbertMills:
# What's changed
- Removed the loading indicator from the PDF reader completely

## Why?
- Full page and smaller loader made page feel slower - decided it is best to display the data as it comes

## Screenshots?


odrakes-cpr (approved):
lgtm

PR #555, opened 2025-06-10
Display results on document view even when slugs are mismatched

PatFawbertMills:
Add additional check on the document source url if we get no results from the slug

# What's changed


katybaulch (approved):
let's get this in

PR #559, opened 2025-06-10
Feature/app 689 add suggested searches to add to homepages of cclw and cpr

katybaulch:
# What's changed

Add new suggested searches to CPR and CCLW but not MCF behind the KG feature flag

## Why?

Alan asked me to


PatFawbertMills (approved):
lgtm

PR #561, opened 2025-06-11
Move search results across when topics is open

PatFawbertMills:
# What's changed
- When the topics slideout is open, move the search results across and do not fade them or remove click events

## Why?
- Feedback on UX is that the search needs to be visible

## Screenshots?
https://github.com/user-attachments/assets/76214e8f-85a3-4da6-9fb7-0586ac35da8d


katybaulch (approved):
this is sexy!!!!!!!!

PR #564, opened 2025-06-11
Feature/app 693 clicking learn more again after the first time will open faqs

katybaulch:
# What's changed

After the new features popover is closed, clicking “learn more” again opens FAQs. It goes to the Topics FAQs heading within FAQs.

This opens as an external link though, so if the user stays in the FAQs page and navigates around there, their first click of learn more on that page will reopen the new features popover & subsequent clicks will open another FAQs page in a new tab.


odrakes-cpr (approved):
Nice

PR #570, opened 2025-06-11
Update drawer as per new design and layout

PatFawbertMills:
# What's changed
- Design of the drawer component has been updated


src/components/drawer/FamilyMatchesDrawer.tsx
```diff
@@ -1,14 +1,13 @@
 import { useRouter } from "next/router";
+import { LuFile } from "react-icons/lu";
```
katybaulch, line 2: issue: are we missing this import?

PR #587, opened 2025-06-17
Run type check in PR workflow

katybaulch:
# What's changed

Add a step to run the type checker (tsc) in the code quality job.

## Why?

We moved away from using our reusable workflows for validating precommits had been adhered to, but when the new pull_request workflow was made, the `tsc` check was omitted from the new workflow file.

### DIFF
--- a/src/utils/buildSearchQueryMetadata.test.ts
+++ b/src/utils/buildSearchQueryMetadata.test.ts
@@ -17,6 +17,9 @@ const testThemeConfig: TThemeConfig = {
   links: [],
   documentCategories: [],
   metadata: [],
+  features: {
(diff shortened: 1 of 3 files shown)


.github/workflows/pull_request.yml
```diff
@@ -89,6 +89,8 @@ jobs:
       - uses: trunk-io/trunk-action@v1
         with:
           arguments: --ci
+      - run: npm run type-check
```
jamesgorrie, line 92: question: so trunk doesn't do this automatically?

katybaulch: it doesn't, may i ask why you think it would?

PR #589, opened 2025-06-17
Feature/app 730 use geographies api for jurisdiction filter on search page

annaCPR:
# What's changed
- source list of countries for the "Published jurisdiction" filter from the new geographies API when no regions are selected
NOTE: the new API currently does not include regions so we're leaving this part of the code alone for now
- fix broken unit tests
- configure vitest to run all tests in the /tests directory but skip playwright tests 

## Why?
- we are gradually switching over to using the new geographies API


src/helpers/getCountriesFromRegions.ts
```diff
@@ -6,7 +6,7 @@ interface IProps {
   countries: TGeography[];
 }
 
-export const getCountriesFromRegions = ({ regions, countries, selectedRegions }: IProps) => {
+export const getCountriesFromRegions = async ({ regions, countries, selectedRegions }: IProps) => {
```
PatFawbertMills, line 9: Question: Does this function need to be asynchronous? I don't see any promises or any `await` within it

vitest.config.mts
```diff
@@ -12,7 +12,7 @@ export default defineConfig({
   test: {
     globals: true,
     environment: "jsdom",
-    exclude: ["**/node_modules/**", "**/.trunk/**", "**/tests/**"],
+    exclude: ["**/node_modules/**", "**/.trunk/**", "**/tests/*.spec.ts", "**/themes/**"],
```
PatFawbertMills, line 15: Interesting - so the page and component tests weren't actually running?

src/hooks/useCountries.ts
```diff
@@ -0,0 +1,25 @@
...
+  alpha2: string;
+  alpha3: string;
+  name: string;
+  official_name: string;
+  numeric: string;
+  flag: string;
+};
+
+export default function useCountries() {
+  const { CONCEPTS_API_URL } = useEnvConfig();
```
PatFawbertMills, line 16: Question: is this usable within this context as opposed to the SSR code of a given page?

Elsewhere we've used: `const { data } = await getEnvFromServer();` to get the env from the FE-server.

src/hooks/useCountries.ts
```diff
@@ -0,0 +1,25 @@
+import { useQuery } from "react-query";
+
+import { ApiClient } from "@/api/http-common";
+import { useEnvConfig } from "@/context/EnvConfig";
+
+export type TCountry = {
```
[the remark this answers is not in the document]
annaCPR: Oops missed that - good spot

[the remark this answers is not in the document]
annaCPR: removed

vitest.config.mts
annaCPR: The ones in the tests folder were not. Only a small portion of the tests were skipped because of this and the ones that were colocated with the source code were running fine. We should decide on an approach though - do we colocate all tests and leave the test folder for playwright tests only? Or do all tests go in the tests folder?

odrakes-cpr:
Just thinking that this can't fully be merged until we update the dataset to include No Geography, International, etc

src/hooks/useCountries.ts
PatFawbertMills: We resolved this and realised we should retire the method i referred to as soon as we can in favour of your approach.

katybaulch (approved):
lgtm

PR #605, opened 2025-06-25
feat: send slack message on github deploy

jamesgorrie:
# What's changed
- sends a slack message to the `#{{env}}-updates` channel when a deploy happens

## Why?

To give us more context of what's being deployed

## notes

There were two options
1. code living in the deploy actions
1. code decoupled from the deploy actions and hooking into the GitHub deploy

I've opted for `2.` for
- the decoupling
- makes it quite an easy lift / shift to other repos or apps
- allows us to capture all deploys rather than tied to things like the `theme` deploys (great for microservices)

## screenshots
Screenshot 2025-06-25 at 09 29 25


jamesgorrie:
Closing until people think it's the right thing

PR #608, opened 2025-06-25
APP-779 Component: Interactive table

jesskelsall:
# What's changed

Added a new generic interactive table for use on subsequent family page work, as well as supporting front-end components.

- Added `Tooltip` component:
  - Builds upon Base UI's [Tooltip](https://base-ui.com/react/components/tooltip) component.
  - Can specify the display direction (top/bottom) and whether to render an arrow.
  - Accompanying Storybook stories.
- Added menu components:
  - `MenuPopup` is a thin styling wrapper around Base UI's [Menu](https://base-ui.com/react/components/menu) `Menu.Popup` (the menu itself).
  - `MenuItem` is a thin styling wrapper around Base UI's [Menu](https://base-ui.com/react/components/menu) `Menu.Item` (each button in the menu).
  - Full pass-through of the Base UI props so these can be used flexibly as normal.
  - Allows for consistently styled reuse when building future menus.
- Added `InteractiveTable` component:
  - Requires a generic for the column IDs (string union type).
  - `columns` prop describes each column: position in the table, width fraction, display name, sortable, and optional tooltip.
  - `rows` is an array of objects with a key for each column.
    - Can provide a primitive string or number for simple data displaying.
    - Can provide a more complex object to differentiate the display and underlying value (used for sorting).
    - Supports `null` values which default to a dash placeholder.
  - Styling can be applied to the table, each row, and each column (via its cells).
  - Sorting functionality:
    - Columns that are sortable show a sort button in their header when hovered.
    - Each sortable column can be sorted ascending or descending, or cleared if already sorted.
    - Uses the new menu components.
  - Chunky Storybook story example.
- Fixed the `Popover` component's arrow which was missing its border.
  - Added a new prop to specify the border colour.

## Why?

- [APP-779](https://linear.app/climate-policy-radar/issue/APP-779/component-feature-rich-table)
- Satisfies [Conor's designs](https://www.figma.com/design/Lg0Vfj7EbUUz8HeqLWRZuK/Q3-2025?node-id=0-1&m=dev) for an interactive table as part of the new family page's "Entries" block.
- It is reasonable to make this table generic for future consistent tabulation.

## Screenshots?

Example table:
Screenshot 2025-06-25 at 14 56 26

Sortable column's heading cell on hover:
Screenshot 2025-06-25 at 14 56 36

Sort menu when column not sorted:
Screenshot 2025-06-25 at 14 56 52

Sort menu when column is sorted:
Screenshot 2025-06-25 at 14 57 02

Example of adding styling to columns and rows:
Screenshot 2025-06-25 at 14 57 11

Tooltip component:
Screenshot 2025-06-25 at 15 15 33

### DIFF
--- a/src/components/organisms/interactiveTable/InteractiveTable.stories.tsx
+++ b/src/components/organisms/interactiveTable/InteractiveTable.stories.tsx
@@ -13,7 +13,7 @@ export default meta;
 
 /* Generic */
 
-type TWainwrightColumns = "height" | "link" | "region" | "wainwright";
+type TWainwrightColumns = "height" | "link" | "region" | "summited" | "wainwright";
 const linkToCell = (link: string): TInteractiveTableCell => ({
   display: (
     <a href={link} className="text-text-brand underline">
@@ -49,6 +49,11 @@ export const Generic: TStory<TWainwrightColumns> = {
         id: "link",
         name: "WalkLakes",
       },
+      {
+        id: "summited",
+        name: "Summited",
+        sortable: true,
+      },
     ],
     defaultSort: {
       column: "height",
@@ -62,6 +67,10 @@ export const Generic: TStory<TWainwrightColumns> = {
           region: "The Eastern Fells",
           height: 950,
           link: linkToCell("https://www.walklakes.co.uk/hill_2515.html"),
+          summited: {
+            display: "23/05/2025",
+            value: "2025-05-23T10:24:00.000Z",
+          },
         },
       },
       {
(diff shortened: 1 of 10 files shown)


src/components/organisms/interactiveTable/InteractiveTable.stories.tsx
```diff
@@ -0,0 +1,133 @@
...
+  component: InteractiveTable,
+  argTypes: {},
+} satisfies Meta<typeof InteractiveTable>;
+type TStory<ColumnKey extends string> = StoryObj<typeof InteractiveTable<ColumnKey>>;
+
+export default meta;
+
+/* Generic */
+
+type TWainwrightColumns = "height" | "link" | "region" | "summited" | "wainwright";
```
katybaulch, line 16: praise: creativity ✨

katybaulch (approved):
looks great, nice one Jess!! 🚀

PR #612, opened 2025-06-30
Feature/app 795 add reporting for other themes other than just cpr

katybaulch:
# What's changed

Add LHCI reporting for all themes across both desktop and mobile

## Why?

Currently our LH reports cover only the CPR app (which is arguably the least important to be reported on, as we do not prioritising SEO yet).

NOTE: We will need to update the baselines for mobile and desktop once we have the new baselines on main.


jamesgorrie (approved):
Fantastic

PR #615, opened 2025-07-02
feat: familyConceptsSearch

jamesgorrie:
# What's changed
- adds `familyConceptsSearch` theme feature flag
- adds getting [concepts from the families-api](https://api.climatepolicyradar.org/families/concepts) behind a theme feature flag
- adds docs on theme feature flags

## Why?
We'd like to start rendering these in the filter panel of CCC.

### DIFF
--- a/src/utils/familyConcepts.ts
+++ b/src/utils/familyConcepts.ts
@@ -1,0 +1,39 @@
+export type FamilyConcept = {
+  relation: "category" | "jurisdiction" | "principle_law";
+  preferred_label: string;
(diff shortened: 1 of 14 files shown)


src/utils/familyConcepts.ts
```diff
@@ -0,0 +1,39 @@
+export type FamilyConcept = {
+  relation: "category" | "jurisdiction" | "principle_law";
```
PatFawbertMills, line 2: Question: is the plan to use this `relation` field to control how and what we display in the filters?

PR #616, opened 2025-07-02
Feature/app 800 update lhci baselines per device

katybaulch:
# What's changed

Updated LHCI to use different config files based on whether its a desktop or mobile run

## Why?

Previously we only had one set of assertions that applied to both desktop and mobile. However, mobile performance is worse, so by splitting them out we can raise the baseline for desktop to prevent regression.


PatFawbertMills:
I siggest we make the performance: `0.65` and accessibility `0.75` for now

PR #619, opened 2025-07-03
Feature/app 831 update applied filters logic to work with new geographies api

katybaulch:
# What's changed

New countries that were not previously in our database will now show the selected pill when they are selected from the published jurisdiction filter typeahead. 

## Why?

The geographies we had in our database are not identical to the new geographies we pull down from our API. There is overlap (Spain, Colombia etc) -- but the differences are geographies such as the Falklands, Faroe Islands etc.

### DIFF
--- a/src/components/CountryLink.tsx
+++ b/src/components/CountryLink.tsx
@@ -1,7 +1,7 @@
 import { FC, ReactNode } from "react";
 
 import { systemGeoCodes } from "@/constants/systemGeos";
-import { getCountrySlug } from "@/helpers/getCountryFields";
+import { getCountrySlug, getCountrySlugOld } from "@/helpers/getCountryFields";
 import useConfig from "@/hooks/useConfig";
 
 import { LinkWithQuery } from "./LinkWithQuery";
@@ -16,9 +16,9 @@ type TCountryLink = {
 
 export const CountryLink: FC<TCountryLink> = ({ countryCode, className = "", emptyContentFallback, children, showFlag = true }) => {
   const configQuery = useConfig();
-  const { data: { countries = [] } = {} } = configQuery;
+  const { data: { countries = [] } = {} } = configQuery; // TODO: Update as part of APP-841
 
-  const slug = getCountrySlug(countryCode, countries);
+  const slug = getCountrySlugOld(countryCode, countries);
   // Force render without any empty content fallback the children without a link
   if (!slug && emptyContentFallback) return <>{emptyContentFallback}</>;
   if (!slug && !emptyContentFallback) return <>{children}</>;
--- a/src/components/CountryLink.tsx
+++ b/src/components/CountryLink.tsx
@@ -1,7 +1,7 @@
 import { FC, ReactNode } from "react";
 
 import { systemGeoCodes } from "@/constants/systemGeos";
-import { getCountrySlug, getCountrySlugOld } from "@/helpers/getCountryFields";
+import { getCountrySlugOld } from "@/helpers/getCountryFields";
 import useConfig from "@/hooks/useConfig";
 
 import { LinkWithQuery } from "./LinkWithQuery";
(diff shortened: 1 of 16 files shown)


src/components/filters/AppliedFilters.tsx
```diff
@@ -61,10 +67,10 @@ const handleFilterDisplay = (
       filterLabel = configCategory ? configCategory.label : value;
       break;
     case "country":
-      filterLabel = handleCountryRegion(value, countries);
+      filterLabel = handleCountry(value, countries);
       break;
     case "region":
-      filterLabel = handleCountryRegion(value, regions);
+      filterLabel = handleRegion(value, regions);
```
jamesgorrie, line 73: ✨

src/helpers/getCountryFields.ts
```diff
@@ -1,21 +1,47 @@
-import { TGeography } from "@/types";
+import { TCountry, TGeography } from "@/types";
 
-const findCountryObject = (countrySearch: string, dataSet: TGeography[]) => {
+// Old
```
jamesgorrie, line 3: note: Using [deprecated](https://jsdoc.app/tags-deprecated) can also work for work like this (not suggesting we do it, just signposting).

src/helpers/getCountryFields.ts
```diff
@@ -1,21 +1,47 @@
...
+  if (!country) return null;
+  return country;
+};
+
+export const getCountrySlugOld = (countrySearch: string, dataSet: TGeography[]) => {
+  return findCountryObjectOld(countrySearch, dataSet)?.slug;
+};
+
+export const getCountryNameOld = (countrySearch: string, dataSet: TGeography[]) => {
+  // Official name is sometimes null.
```
jamesgorrie, line 17: question: I'm unclear what this comment is sign posting?

src/components/CountryLink.tsx
```diff
@@ -1,7 +1,7 @@
 import { FC, ReactNode } from "react";
 
 import { systemGeoCodes } from "@/constants/systemGeos";
-import { getCountrySlug } from "@/helpers/getCountryFields";
+import { getCountrySlug, getCountrySlugOld } from "@/helpers/getCountryFields";
```
jamesgorrie, line 4: question: I am not sure if I am missing something here - why do we import both?

jamesgorrie (commented):
Question on why we're importing both Old and the new one? Might just be late in the day, but I can't wrap my head around it.

src/helpers/getCountryFields.ts
katybaulch: it's not deprecated yet, still used by 90% of the application :)

src/components/CountryLink.tsx
katybaulch: we shouldn't be -- great spot!

src/helpers/getCountryFields.ts
```diff
@@ -1,21 +1,46 @@
...
+export const getCountryNameOld = (countrySearch: string, dataSet: TGeography[]) => {
+  return findCountryObjectOld(countrySearch, dataSet)?.display_value;
+};
+
+export const getCountryCodeOld = (countrySearch: string, dataSet: TGeography[]) => {
+  return findCountryObjectOld(countrySearch, dataSet)?.value;
+};
+
+// New geographies-api
+const findCountryObject = (countrySearch: string, dataSet: TCountry[]) => {
```
jamesgorrie, line 25: non-blocking: Is this a little clearer? Not 

```suggestion
const findCountry = (countrySearch: string, dataSet: TCountry[]) => {
```

src/helpers/getCountryFields.ts
katybaulch: end of week brain 🙃

jamesgorrie (approved):

PR #622, opened 2025-07-07
Feature/app 759 exact matches checkbox on homepage defaults to ticked

katybaulch:
# What's changed

- Exact matches checkbox on homepage defaults to checked (checked if exact_match is true or not present, false if exact_match explicitly set to false)
- Update router param for exact_match to true by default 
- Updated the AppliedFilter logic and make sure that we don't remove the router param for exact match so that we preserve the correct query string

## Why?

Semantic search is no longer the default

## Screenshots?

image


odrakes-cpr (approved):
lgtm

PR #623, opened 2025-07-07
Feature/app 823 fix seo issue with lighthouse in ci

katybaulch:
# What's changed

Pass robots env var value directly to lighthouse

## Why?

we were building with robots set to true but not passing this value on to LHCI


PatFawbertMills (approved):
LGTM

PR #632, opened 2025-07-08
Add playright tests for OEP landing page & search

katybaulch:
# What's changed

Added playright tests for OEP landing page and search

## Why?

I'm finding it hard to remember all the different user flows that manually need testing whenever I make a change, so going to start automating these to avoid mishaps.


PatFawbertMills (approved):
Great to add this.

PR #638, opened 2025-07-08
APP 882 - posthog persistence

PatFawbertMills:
# What's changed
- We only set the config value based on consent. This approach fixes the previous issue of not persisting user data between sessions
- Previously to this, each time we initialise posthog (i.e. on page refresh) it removed the previous cookie and replaced wth a new one, which treated every user as a unique visitor.

### DIFF
--- a/src/context/PostHogProvider.tsx
+++ b/src/context/PostHogProvider.tsx
@@ -59,19 +59,21 @@ export function PostHogProvider({ children, consent }: IProps) {
    * @see: https://posthog.com/docs/privacy#is-it-ok-for-my-api-key-to-be-exposed-and-public
    */
   useEffect(() => {
-    /** If consent is granted, initialize our cookied PostHog instance */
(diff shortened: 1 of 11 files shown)


src/context/PostHogProvider.tsx
```diff
@@ -59,19 +59,21 @@ export function PostHogProvider({ children, consent }: IProps) {
...
     window.sessionStorage.setItem("posthogLoaded", "true");
   }, []);
 
   useEffect(() => {
+    // We only set the config value based on consent
+    // This approach fixes the previous issue of not persisting user data between sessions
     if (consent) {
       posthog.set_config({ persistence: "localStorage+cookie" });
+    } else {
+      posthog.set_config({ persistence: "memory" });
```
jamesgorrie, line 76: Interesting - I do remember _thinking_ persistence was set to memory by default - [but apparently it's not](https://posthog.com/docs/libraries/js/persistence).

jamesgorrie (approved):
Nice and private

PR #640, opened 2025-07-09
Connect frontend litigation filters to vespa search

odrakes-cpr:
# What's changed

Send litigation concept filters through to the vespa search query. 

We have followed the other formats of updating the metadata query param. 

The screenshot below is from the staging endpoint, currently our staging dataset is not as rich with all the available concepts (principle laws etc) so I use a concept label that I knew existed on these cases to test. 

## Screenshots?
Screenshot 2025-07-10 at 10 30 59


PatFawbertMills (approved):
lgtm

PR #642, opened 2025-07-09
Add playright tests for CPR homepage

katybaulch:
# What's changed

- Add playright tests for CPR homepage and search from the homepage
- Fixed CCLW empty search test, as we would expect to be redirected to the search page

## Why?

We didn't have any


PatFawbertMills (approved):
You're _testing_ my patience here! 😉

PR #645, opened 2025-07-10
Feature/app 887 revamp geography filter on search page

annaCPR:
# What's changed
- add new button "Geography" to the filters on the search page which opens a slideout to the right
- move existing existing geography filters ("Region" and "Published Jurisdiction") to the new slideout
- change the country filter to be a full alphabetised list of multiselect checkboxes instead of a TypeAhead component

- drive-by: install @testing-library/jest-dom to get better test matchers and update all relevant tests to assert on items being actually rendered and visible in the DOM as the vitest matchers we were using before were not doing that

## Screenshots?
Screenshot 2025-07-10 at 12 57 18

### DIFF
--- a/src/pages/search/index.tsx
+++ b/src/pages/search/index.tsx
@@ -135,6 +135,8 @@ const Search: InferGetServerSidePropsType<typeof getServerSideProps> = ({
     keyword_filters: { countries: countryFilters = [], regions: regionFilters = [] },
   } = searchQuery;
 
+  const alphabetisedCountries = countries.sort((c1, c2) => c1.display_value.localeCompare(c2.display_value));
+
   const availableCountries = useMemo(() => {
     return regionFilters.length > 0
       ? getCountriesFromRegions({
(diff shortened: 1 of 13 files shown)


src/pages/search/index.tsx
```diff
@@ -465,6 +484,50 @@ const Search: InferGetServerSidePropsType<typeof getServerSideProps> = ({
                       )}
                     </SlideOut>
                   )}
+                  <SlideOut showCloseButton={false}>
```
PatFawbertMills, line 475: Suggestion: see how with the `ConceptPicker` above on line `481` could we potentially move the contents of the slideout into a separate component? And also while doing that move the associated logic? Would be nice to make the search page smaller

PatFawbertMills: Couple of reasons I see for this:
1. Any state logic can be encapsulated within the separate component rather than adding more to the search page (already a far too big component imo @ > 700 lines). And more broadly the behaviour can be managed separately to the search page
2. Managing changes to the component becomes more traceable and less chance for conflicts
3. Ideally the search page would only consist of core search logic and all state and interactivity concerns would be managed in the appropriate components, we aren't there now but that's the ambition. Moving this into a separate component I think gets us a little closer to that.

src/pages/search/index.tsx
```diff
@@ -465,6 +472,50 @@ const Search: InferGetServerSidePropsType<typeof getServerSideProps> = ({
...
+                    <div className="text-sm text-text-secondary flex flex-col gap-5">
```
PatFawbertMills, line 476: We also need to add this to the `SlideOut` above (see line `468`) with a check on the `currentSlideOut` context value to ensure we are displaying the correct slide out content.

Currently if you click on the "Topics" item, this new geography slide out appears, taking UI precedent over the others.

src/pages/search/index.tsx
```diff
@@ -465,6 +472,50 @@ const Search: InferGetServerSidePropsType<typeof getServerSideProps> = ({
...
+                              checked={regionFilters && regionFilters.includes(region.slug)}
+                              onChange={() => {
+                                handleRegionChange(region.slug);
+                              }}
+                              name={`region-${region.slug}`}
+                            />
+                          ))}
+                        </InputListContainer>
+                      </Accordian>
+                      <Accordian
```
PatFawbertMills, line 496: I think we should add the quick search attribute to this so that we can use the filtering

PR #648, opened 2025-07-14
APP-901: Display alternative pages when the litigation feature is active

jesskelsall:
# What's changed

When the litigation feature (feature flag or config feature) is active, show a completely different page for families and geographies. This also includes the new collection page which already doesn't show without this feature enabled.

- Page splitting:
  - Added components for each page display in `src/components/pages`.
  - There is a pair of components for each page e.g. `GeographyOriginalPage` and `GeographyLitigationPage`.
  - Each pair of components shares the same `IProps` interface, which is also used by the Next.js page.
  - The Next.js page renders one of these two components depending on the feature and passes all props through.
    - Ensures a consistent interface and that API calls happen in `getServerSideProps` as much as possible.
  - Once litigation is launched, the intent is to collapse this back into the original Next.js pages.
- Page rendering:
  - All 3 pages have been pared back to `Layout`, `Columns`, `ContentsSideBar` (typically non-functional yet), and a `<pre>` to render the raw data object for in-dev interrogation.

## Why?

- As the UI changes on these pages are significant, we want to see the new UI in its entirety, not mixed in with the existing pages. This is especially useful as the way we're laying out the pages is changing (which is fundamental).
- Satisfies [APP-901](https://linear.app/climate-policy-radar/issue/APP-901/display-alternative-pages-when-the-litigation-feature-is-active).

## Screenshots

Family page:
Screenshot 2025-07-14 at 14 13 31

Geography page:
Screenshot 2025-07-14 at 14 13 46

### DIFF
--- a/src/components/molecules/warning/Warning.stories.tsx
+++ b/src/components/molecules/warning/Warning.stories.tsx
@@ -1,0 +1,42 @@
+import { Meta, StoryObj } from "@storybook/react";
+import { useState } from "react";
+
+import { Warning } from "./Warning";
+
+const meta = {
+  title: "Molecules/Warning",
+  component: Warning,
+  argTypes: {
+    children: { control: false },
+  },
+} satisfies Meta<typeof Warning>;
+type TStory = StoryObj<typeof Warning>;
+
+export default meta;
+
+export const Primary: TStory = {
+  args: {
+    children: (
+      <p>
+        The warning component is used to display important information or alerts to users. It can be styled with different variants such as "info" or
+        "error". The component can also be hideable, allowing users to dismiss it and not see it again in the current session.
+      </p>
+    ),
+    variant: "info",
+    hideableId: "info-warning-id",
+    className: "",
+  },
+};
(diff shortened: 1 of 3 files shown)


jamesgorrie (commented):
thought: we could avoid such a large set of diffs and changes going forward by by returning early if it is the new page and continue as before if not

Currently we have:

```ts
import NewPage
import OldPage
 
function Page() {
    if (isNewPageEnabled) return <NewPage />
    else return <OldPage />
}

// Slightly simpler diff and way to migrate forward post having two pages

import NewPage
function Page() {
  if (isNewPageEnabled) return <NewPage />
  
  //... continue as before
}
```

☝️ this might also help with when we add new props from the data fetching later...

Slight simplification, but not too fussed as it's work we'll probably tidy once we have 1 page layout.

PatFawbertMills (approved):
Nice one.

Tested locally to confirm SSR renders the correct dynamic page component on the server by viewing the page source.

PR #654, opened 2025-07-15
Add Warning component to search page

PatFawbertMills:
# What's changed
- Add new `Warning` component
- Component only currently handles variants "Info" and "Disclaimer" - expand it later to include warnings and errors
- Dismissing it will store a session token to prevent showing again curing session

## Screenshots?
**Info**:
Screenshot 2025-07-15 at 15 58 57

**Disclaimer**:
Screenshot 2025-07-15 at 15 59 33


src/components/molecules/warning/Warning.tsx
```diff
@@ -0,0 +1,60 @@
...
+    if (!hideableId) return;
+    if (typeof window === "undefined") return;
+    sessionStorage.setItem(hideableId, "true");
+    setDisplayWarning(false);
+  };
+
+  useEffect(() => {
+    // Need to check if the hideableId is set in sessionStorage
+    // Also check if the window object is available (to avoid SSR hydration issues)
+    typeof window !== "undefined" && sessionStorage.getItem(hideableId) !== "true" ? setDisplayWarning(true) : setDisplayWarning(false);
```
jesskelsall, line 45: Nice, I like that this component encapsulates the nitty-gritty logic so that implementing it is simple. ✨

jesskelsall (approved):
Looks good, and thank you for taking inspiration from my components when building this!

I do have to wonder if there should be this much difference in styling - I'd personally push back on them looking any different except for the close button at this stage. 🤔

PatFawbertMills:
> Looks good, and thank you for taking inspiration from my components when building this!
> 
> I do have to wonder if there should be this much difference in styling - I'd personally push back on them looking any different except for the close button at this stage. 🤔

That's fair. I'll catch up with Conor tomorrow.

PatFawbertMills:
Have removed the "disclaimer" variant after consulting Conor. He was happy with me to make the decision and I agree with you Jess - for consistency it is nicer with one style. 

I also added an example variant for "error" so that the road is paved for whoever adds to this component next.

PR #664, opened 2025-07-23
APP-908: Text block component

jesskelsall:
# What's changed

- Revised some existing components that no longer match the most recent designs / our use-cases:
  - Removed `ShowMore` component as there won't be a "one size fits all" approach to revealing more content.
  - `Section` component is now a thin wrapper in service of `ContentsSideBar` with little more than a title and section markup.
- Added `TextBlock` component:
  - Inherits show more functionality from `ShowMore` but with a style overhaul.
  - ReactNode as children (cares not for our `dangerouslySetInnerHTML` tomfoolery).
- Added `TextBlock` to pages:
  - Family page
  - Collection page

## Why?

Satisfies [APP-908](https://linear.app/climate-policy-radar/issue/APP-908/text-block-component)

## Screenshots?

TextBlock in Storybook:
Screenshot 2025-07-23 at 11 14 05

Updated family page:
Screenshot 2025-07-23 at 11 28 48


src/styles/styles.css
```diff
@@ -353,9 +353,8 @@ ul li.selected {
 .text-content a {
   @apply text-blue-600 underline hover:text-blue-800;
 }
-.text-content p,
-.search-results p {
-  @apply my-5;
+.text-content p + p {
```
jesskelsall, line 356: This may be naïve, but the intent here is to prevent leading and trailing margins on paragraphs. The issue is that creates gaps that affect the layout of neighbouring elements, which is undesirable.

src/components/blocks/textBlock/TextBlock.tsx
```diff
@@ -0,0 +1,44 @@
...
+    <Section id="section-summary" title="Summary">
+      <div className="relative">
+        <div className={containerClasses} style={{ maxHeight: isOpen ? undefined : maxHeight }}>
+          {/* DOM element to measure height of */}
+          <div ref={contentRef}>{children}</div>
+        </div>
+        {contentIsOverflowing && (
+          <div className="pt-4">
+            <button type="button" onClick={onToggle} className="text-sm text-text-tertiary hover:underline">
+              {isOpen ? "+ View less" : "+ View more"}
```
PatFawbertMills, line 37: Question: should it be a `-` for viewing less?

src/components/molecules/section/Section.tsx
```diff
@@ -1,35 +1,15 @@
...
-
 export interface IProps {
   children: ReactNode;
-  className?: string;
-  controls?: ReactNode;
-  explanation?: ReactNode;
-  title: string | ReactNode;
-  titleBadge?: string;
+  id: string;
+  title?: string;
```
PatFawbertMills, line 6: Question: does limiting the title to be a string limit us to adding things like tags/badges whatever later?

src/styles/styles.css
PatFawbertMills: Makes sense, prefer this to `:first-child { @apply mt-0; }` !

PatFawbertMills (approved):
Code looks good, couple of small Qs

PatFawbertMills:
Is it missing a fade-to-white bottom overlay? Or are we not doing in first pass?

jesskelsall:
@PatFawbertMills that has disappeared from [the latest designs](https://www.figma.com/design/Lg0Vfj7EbUUz8HeqLWRZuK/Q3-2025?node-id=308-6&m=dev). I know Conor wasn't happy with how the previous gaussian blur fade looked.

PR #667, opened 2025-07-24
APP 858 - metadata block

PatFawbertMills:
# What's changed
- New metadata block
- A start on creating a handler to construct the data to serve the new component from the available family data

## Why?
- New designs and content to display on the family page

## Screenshots?
- See designs in ticket


src/components/blocks/metadataBlock/MetadataBlock.tsx
```diff
@@ -0,0 +1,24 @@
+import { Section } from "@/components/molecules/section/Section";
+import { IMetadata } from "@/types";
+
+interface IProps {
+  metadata: IMetadata[];
+}
+
+export const MetadataBlock = ({ metadata }: IProps) => {
+  return (
+    <Section title="About this case" id="metadata-block">
```
katybaulch, line 10: question: is the metadata block component coupled with litigation, or is it intended to be more generic?

PatFawbertMills: good spot, needs to be more generic, I should be passing in the title prop!

PR #669, opened 2025-07-28
Improve Subdivision Filter Experience by Showing Only Relevant Subdivisions with Data

odrakes-cpr:
feat: switch useSubdivisions query to only return divs with data

- we already have an endpoint in the families backend that returns a list of geographies and their doc counts where this value is greater than 0

Our current ui is not suitable for users as we were return our entire trove c.5000 subdivisions in the list that users would use to filter or select sub divisions - so we have decided to fix this

## Screenshots?
Screenshot 2025-07-28 at 13 57 54


annaCPR (approved):
very nice!

PR #676, opened 2025-07-29
Feature/app 935 add world map to ccc landing page

annaCPR:
# What's changed
- update the names of countries in the json file that powers the world map so that they match the names in the data returned from the config endpoint

## Why?
- for some of the countries on the map, when a user would hover over that country we were showing a fallback message because the country was not found in the data returned from the config endpoint

## Screenshots?
Screenshot 2025-07-29 at 12 29 46


odrakes-cpr (approved):
looks gd

jamesgorrie (approved):

PR #677, opened 2025-07-29
APP 932 update metadata block display

PatFawbertMills:
# What's changed
- ensure correct values are displayed
- ensure hierarchy of concepts is respected
- ensure parent concepts are repeated when different hierarchy trees are present

## Why?
- so that we can accurately display the correct date for a given litigation case

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -198,6 +198,7 @@ export type TFamilyPage = {
   published_date: string | null;
   last_updated_date: string | null;
   status?: string;
+  concepts: TFamilyConcept[];
 };
 
 export type TDocumentContentType = "application/pdf" | "text/html" | "application/octet-stream";
@@ -356,6 +357,15 @@ export type TConcept = {
   subconcept_of: string[];
   wikibase_id: string;
   type?: "principal_law" | "jurisdiction" | "category";
+};
+
+export type TFamilyConcept = {
+  id: string;
+  ids: string[];
+  type: string;
+  relation: string;
+  preferred_label: string;
+  subconcept_of_labels: string[];
 };
 
 export type TSearchResponse = {
--- a/src/utils/buildConceptHierarchy.ts
+++ b/src/utils/buildConceptHierarchy.ts
@@ -26,9 +26,9 @@ function buildTree(
   }
   visited.add(concept.id);
 
-  // Find children: concepts where this concept is a parent
+  // Find children: concepts where this concept is a parent and of the same type
   const children = allConcepts
-    .filter((child) => child.subconcept_of_labels.includes(concept.preferred_label))
+    .filter((child) => child.subconcept_of_labels.includes(concept.preferred_label) && child.type === concept.type)
     .map((child) => buildTree(child, labelMap, allConcepts, new Set(visited)));
 
   return {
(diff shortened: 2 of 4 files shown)


src/types/types.ts
```diff
@@ -198,6 +198,7 @@ export type TFamilyPage = {
   published_date: string | null;
   last_updated_date: string | null;
   status?: string;
+  concepts: TFamilyConcept[];
```
jesskelsall, line 201: Note to self: we need to be intentional about whether `TFamilyPage` is purely the shape of the V1 API response, or if it will be an combination of new and old.

src/utils/buildConceptHierarchy.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  return {
+    ...concept,
+    children,
+  };
+}
+
+export function buildConceptHierarchy(concepts: TFamilyConcept[]): TFamilyConceptTreeNode[] {
+  // Root concepts: those with no parents
+  const roots = concepts.filter((concept) => concept.subconcept_of_labels.length === 0);
```
jesskelsall, line 27: Praise: Nice, I was hoping the starting point for this would be straightforward, and it is!

src/utils/buildConceptHierarchy.ts
```diff
@@ -0,0 +1,31 @@
+import { TFamilyConcept } from "@/types";
+
+export type TFamilyConceptTreeNode = TFamilyConcept & {
+  children: TFamilyConceptTreeNode[];
+};
+
+function buildTree(concept: TFamilyConcept, allConcepts: TFamilyConcept[], visited: Set<string>): TFamilyConceptTreeNode {
+  // Prevent infinite cycles
+  if (visited.has(concept.id)) {
```
jesskelsall, line 9: Question: I'm assuming this is just in case the API response has circular data, rather than any error of our FE processing?

src/utils/buildConceptHierarchy.ts
```diff
@@ -0,0 +1,31 @@
...
+    throw new Error(`Cycle detected at concept id: ${concept.id}`);
+  }
+  visited.add(concept.id);
+
+  // find concepts where this concept is a parent and of the same type
+  const children = allConcepts
+    .filter((child) => child.subconcept_of_labels.includes(concept.preferred_label) && child.type === concept.type)
```
jesskelsall, line 16: Observation: Pretty good. Could be performance heavy with a lot of concepts, but I don't think we'll ever have enough for this to be significant.

jesskelsall (approved):
Looks good! A few comments (food for thought), but nothing blocking.

src/utils/buildConceptHierarchy.ts
PatFawbertMills: Yeh, if the API has given us the same concept somehow as a child of itself, we don't want to crash the app in an infinite loop!

src/utils/buildConceptHierarchy.ts
PatFawbertMills: Yeh defo a good shout. Trouble is the requirement is "infinite depth" so our hands are tied to an extent. Most I've seen is around 4/5 so its not end of world.

PR #678, opened 2025-07-30
Feature/app 968 render the correct canonical in navigator frontend

odrakes-cpr:
Per title we want to use the attribution url on a family to render the correct canonical url on our frontend document and family pages.


jamesgorrie (approved):
YES 🤑 !

PR #679, opened 2025-07-30
APP-950: Documents block: Component for collections page (per family)

jesskelsall:
# What's changed

- Added `FamilyBlock` component:
  - Renders an `InteractiveTable` of all events from the family (family events + document events).
  - When associated, links to an event's parent document is present.
  - Family title and the beginnings of some metadata above the table.
  - Example Storybook story with an especially unwieldy family present on CCC prod.
- API response types:
  - Added `TFamilyNew` as a "for now" extended alternative to `TFamilyPage`. 
  - Same goes for `TDocumentNew`.
  - Form of tech debt - old types will need removing after the V1 API is no longer used.
- Iterated `InteractiveTable` component:
  - `maxRows` prop limits how many sorted rows to render.
  - `showValues` prop is a debug mode for seeing the underlying values in each cell - useful for debugging if your table is sorting as expected.
- Misc:
  - `formatDateShort` helper function for DD/MM/YYYY (or local equivalent) date display.
  - Fixed `joinNodes` which wasn't accounting for essentially `null` React nodes.
  - Removed Storybook stories for pre component library components (was an example, no longer needed).

## Why?

Satisfies [APP-950](https://linear.app/climate-policy-radar/issue/APP-950/documents-block-component-for-collections-page-per-family)

## Screenshots?

A full-fat and wordy example:
Screenshot 2025-07-30 at 15 14 48


src/components/organisms/interactiveTable/InteractiveTable.tsx
```diff
@@ -126,6 +136,14 @@ export const InteractiveTable = <ColumnKey extends string>({ columns, defaultSor
     );
   };
 
+  const renderCellDisplay = (cell: Record<ColumnKey, TInteractiveTableCell>[ColumnKey]) => {
```
PatFawbertMills, line 139: Suggestion: I have a bit of an aversion to render methods contained within a component because it gets redefined on every render. It looks like only the `showValues` value is being used from within the context of this component so maybe we could pass that in when we call the render method if we put it outside the `InteractiveTable` parent?

PatFawbertMills (approved):
One suggestion but not blocking.

The page is really coming together now!

src/components/organisms/interactiveTable/InteractiveTable.tsx
jesskelsall: Good shout 🙂  Fixed

PR #682, opened 2025-07-31
Feature/app 972 world map missing legend and markers

odrakes-cpr:
# What's changed

Fix : map was missing legend / markers which would show little dots on the countries making it more accessible for users.

Driveby - removed some mcf code that i believe was added when we thought mcf would add a map to their home page

## Screenshots?
Screenshot 2025-07-31 at 11 14 14

### DIFF
--- a/src/components/map/WorldMap.tsx
+++ b/src/components/map/WorldMap.tsx
@@ -140,7 +140,7 @@ interface IProps {
   customApp: "ccc" | "cclw" | "cpr";
 }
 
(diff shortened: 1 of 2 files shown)


katybaulch (approved):
nice one!  do we just have roughly the same amount of litigation data per geo apart from the US??

odrakes-cpr:
umm @katybaulch us is the bulk of things and then its autralia and brazil which have the bulk of other litigation cases then its spread out fairly acorss the world. this screenshot is from staging

PR #686, opened 2025-08-04
Get subdivisions in server-side render code and load into metadata block

PatFawbertMills:
# What's changed
- Update metadata block to display the subdivision

## Why?
- Display the user-friendly name for a subdivision

### DIFF
--- a/src/helpers/getSubdivision.ts
+++ b/src/helpers/getSubdivision.ts
@@ -1,0 +1,20 @@
+import { TGeographySubdivision } from "@/types";
+
+const findSubdivision = (search: string, dataSet: TGeographySubdivision[]) => {
+  let subdivision = dataSet.find((c) => c.code.toLowerCase() === search.toLowerCase());
+  if (!subdivision) subdivision = dataSet.find((c) => c.name.toLowerCase() === search.toLowerCase());
+  if (!subdivision) return null;
+  return subdivision;
+};
+
+export const getSubdivisionName = (search: string, dataSet: TGeographySubdivision[]) => {
+  return findSubdivision(search, dataSet)?.name;
+};
+
+export const getSubdivisionCode = (search: string, dataSet: TGeographySubdivision[]) => {
+  return findSubdivision(search, dataSet)?.code;
+};
+
+export const getSubdivisionType = (search: string, dataSet: TGeographySubdivision[]) => {
+  return findSubdivision(search, dataSet)?.type;
+};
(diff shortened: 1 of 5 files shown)


src/helpers/getSubdivision.ts
```diff
@@ -0,0 +1,20 @@
+import { TGeographySubdivision } from "@/types";
+
+const findSubdivision = (search: string, dataSet: TGeographySubdivision[]) => {
+  let subdivision = dataSet.find((c) => c.code.toLowerCase() === search.toLowerCase());
+  if (!subdivision) subdivision = dataSet.find((c) => c.name.toLowerCase() === search.toLowerCase());
```
jesskelsall, line 5: Request: Is there a way this can avoid potentially iterating over the full data set twice?

src/pages/document/[id].tsx
```diff
@@ -85,6 +86,20 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
     } catch (error) {}
   }
 
+  if (familyData) {
+    const allSubdivisions = await Promise.all<TGeographySubdivision[]>(
+      familyData.geographies
+        .filter((country) => country.length === 3)
+        .map(async (country) => {
+          try {
+            const { data: subDivisionResponse } = await apiClient.get(`/geographies/subdivisions/${country}`);
```
Comment (jesskelsall) [inline src/pages/document/[id].tsx:95]: Observation: We need a better solution than making _n_ API queries, though out of scope of this PR.

src/utils/getFamilyMetadata.tsx
```diff
@@ -25,22 +26,27 @@ function displayConceptHierarchy(concept: TFamilyConceptTreeNode): React.ReactNo
...
   return familyMetadata;
 };
 
-function getLitigationMetaData(family: TFamilyNew, countries: TGeography[]): IMetadata[] {
+function getLitigationMetaData(family: TFamilyNew, countries: TGeography[], subdivisions: TGeographySubdivision[]): IMetadata[] {
   const metadata = [];
 
   // Structure concepts into a hierarchy we can use
   const hierarchy = buildConceptHierarchy(family.concepts);
+  const geosOrdered = family.geographies.sort((a, b) => {
```
jesskelsall, line 45: Suggestion: For a complex sort (especially if it would be beneficial to sort alphabetically _after_ length, consider [Lodash orderBy](https://lodash.com/docs/4.17.15#orderBy).

jesskelsall (approved):
Good stuff! A few comments for consideration, but nothing I'd consider a hard blocker. Use your own discretion. 🙂

src/utils/getFamilyMetadata.tsx
PatFawbertMills: Good shout on both lodash and sortby alophabetical

PR #687, opened 2025-08-05
Turn off KG and concepts for CCC

PatFawbertMills:
# What's changed
- There will not be KG as part of the CCC first launch.

## Why?
- To avoid the conversation each time we review the CCC app as to why the KG filters are present.


odrakes-cpr (approved):
looks good

themes/ccc/config.ts
```diff
@@ -70,8 +70,8 @@ const config: TThemeConfig = {
   documentCategories: ["All", "Laws", "Policies", "UNFCCC Submissions", "Litigation", "Climate Finance Projects", "Offshore Wind Reports"],
   defaultDocumentCategory: "All",
   features: {
-    familyConceptsSearch: true,
-    knowledgeGraph: true,
+    familyConceptsSearch: false,
```
jamesgorrie, line 73: This should be true as it includes case categories etc

```suggestion
    familyConceptsSearch: true,
```

PatFawbertMills: ah yeh same, sorry will revert.

PR #693, opened 2025-08-12
APP-1035: Data for family event table component

jesskelsall:
# What's changed

- Added new functions to build an interactive table of events:
  - `getEventTableColumns` determines which columns to display based on a few arguments.
  - `getEventTableRows` unpacks multiple families into an array of events and associated documents, then turns each into a row for the table.
  - As a result, each of the 3 tabular blocks has had its bespoke versions of the above replaced.

See the Linear ticket below for a greater understanding of what's been asked for, what I've actually applied, and what's potentially unclear currently.

## Why?

Satisfies [APP-1035](https://linear.app/climate-policy-radar/issue/APP-1035/data-for-family-event-table-component).

## Screenshots?

### DIFF
--- a/src/utils/getFamilyMetadata.tsx
+++ b/src/utils/getFamilyMetadata.tsx
@@ -129,5 +129,16 @@ function getLitigationMetaData(family: TFamilyPublic, countries: TGeography[], s
     value: <div className="grid">{principalLaws.length > 0 ? principalLaws.map((law) => displayConceptHierarchy(law)) : "N/A"}</div>,
   });
 
+  /* At issue */
+
+  metadata.push({
+    label: "At issue",
+    value: (
+      <div className="grid">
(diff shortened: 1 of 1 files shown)


src/utils/eventTable.tsx
```diff
@@ -0,0 +1,116 @@
...
+  if (courtConcepts.length === 0) return null;
+
+  // On each loop, remove legal entities without parents. Stops when the deepest level court remains
+  while (courtConcepts.length > 1) {
+    const moreSpecificConcepts = courtConcepts.filter((concept) =>
+      concept.subconcept_of_labels.some((id) => concepts.findIndex((con) => con.id === id) !== -1)
+    );
+
+    // Prevent a situation where number of concepts goes from 2 to 0 on the last loop
+    courtConcepts = moreSpecificConcepts.length === 0 ? courtConcepts.slice(0, 1) : moreSpecificConcepts;
```
PatFawbertMills, line 59: Pondering: in situations where there are two lowest level courts, should we display both rather than the first instance?

PR #702, opened 2025-08-18
Turn off video and screenshot in config to speed up tests in CI

annaCPR:
# What's changed
- explicitly set 'video' and 'screenshot' config options for playwright to 'off' in order to speed up tests in CI

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -5,12 +5,40 @@ const config = {
     baseURL: "http://localhost:3000",
     useWebserver: true,
   },
-  staging: {
+  // CPR
+  cpr_staging: {
     baseURL: "https://cpr.staging.climatepolicyradar.org",
     useWebserver: false,
   },
-  production: {
+  cpr_production: {
     baseURL: "https://app.climatepolicyradar.org",
+    useWebserver: false,
+  },
+  // CCLW
+  cclw_staging: {
+    baseURL: "https://cclw.staging.climatepolicyradar.org",
+    useWebserver: false,
+  },
+  cclw_production: {
+    baseURL: "https://climate-laws.org",
+    useWebserver: false,
+  },
+  // MCF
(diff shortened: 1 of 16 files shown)


playwright.config.ts
```diff
@@ -64,7 +64,8 @@ export default defineConfig({
   use: {
     /* Base URL to use in actions like `await page.goto('/')`. */
     baseURL: envConfig.baseURL,
-
+    video: "off",
```
jamesgorrie, line 67: [Videos are off by default](https://playwright.dev/docs/videos#record-video)?

playwright.config.ts
```diff
@@ -64,7 +64,8 @@ export default defineConfig({
...
+    screenshot: "off",
```
jamesgorrie, line 68: [These are off by default too](https://playwright.dev/docs/test-use-options#recording-options).

jamesgorrie:
@katybaulch @annaCPR I've found previously using a setting that captures videos and / or screenshots on failure, [like this](https://playwright.dev/docs/test-use-options#recording-options%23recording-options), is very useful for debugging what went wrong.

It shouldn't slow down builds in general, but might make a failed build take a second longer, but we'd gain that speed on debugging.

katybaulch:
> @katybaulch @annaCPR I've found previously using a setting that captures videos and / or screenshots on failure, [like this](https://playwright.dev/docs/test-use-options#recording-options%23recording-options), is very useful for debugging what went wrong.
> 
> It shouldn't slow down builds in general, but might make a failed build take a second longer, but we'd gain that speed on debugging.

We're not doing anything with screenshots or videos at the moment, so this was a way of explicitly trying to save some time until we do decide to do something with it. Equally - do we need to capture images/videos on pull request and merge to main and smoke test?? That feels like a way of eating space fast without a way of mananging it?

PR #717, opened 2025-08-26
feat: start using Vespa search results on geo pages

jamesgorrie:
# What's changed
Starts rendering search results from Vespa on the geographies page

## Why?
In aid of https://linear.app/climate-policy-radar/issue/APP-1066/use-document-type-as-the-grouped-lists-on-geography-pages

This is definitely a v0.1 - but trying to keep the PRs small (and hence the feature switch)


jesskelsall (approved):
Looks good for getting the data on the page, I'm excited to integrate this into my ongoing work! 💥

PR #718, opened 2025-08-26
feat: use theme config to generate family lists on geo pages

jamesgorrie:
# What's changed
Dependant on https://github.com/climatepolicyradar/navigator-frontend/pull/717

Uses the config files to generate the filters on the pages pages.

Part of: https://linear.app/climate-policy-radar/issue/APP-1066/use-document-type-as-the-grouped-lists-on-geography-pages

## Screenshots?

https://github.com/user-attachments/assets/f994db8c-1f10-4d84-af56-9cf333ad3ad6

### DIFF
--- a/src/components/pages/geographyOriginalPage.tsx
+++ b/src/components/pages/geographyOriginalPage.tsx
@@ -1,6 +1,7 @@
 import { useRouter } from "next/router";
 import { useState } from "react";
 
+import { ApiClient } from "@/api/http-common";
 import { BrazilImplementingNDCCard } from "@/cclw/components/BrazilImplementingNDCCard";
 import { Alert } from "@/components/Alert";
 import { ExternalLink } from "@/components/ExternalLink";
@@ -20,6 +21,7 @@ import { Event } from "@/components/timeline/Event";
 import { Timeline } from "@/components/timeline/Timeline";
 import { Heading } from "@/components/typography/Heading";
 import { QUERY_PARAMS } from "@/constants/queryParams";
+import { TPublicEnvConfig } from "@/context/EnvConfig";
 import {
   TDocumentCategory,
   TEvent,
@@ -33,6 +35,7 @@ import {
   TTheme,
   TThemeConfig,
 } from "@/types";
+import buildSearchQuery from "@/utils/buildSearchQuery";
 import { sortFilterTargets } from "@/utils/sortFilterTargets";
 
 export interface IProps {
@@ -44,6 +47,7 @@ export interface IProps {
   theme: TTheme;
   themeConfig: TThemeConfig;
   vespaSearchResults?: TSearch;
+  envConfig: TPublicEnvConfig;
 }
 
 const categories: { title: TDocumentCategory; slug: string }[] = [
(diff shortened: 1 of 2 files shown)


src/components/pages/geographyOriginalPage.tsx
```diff
@@ -217,6 +221,37 @@ export const GeographyOriginalPage = ({ geography, summary, targets, theme, them
     }
   };
 
+  /**
+   * Vespa search results
+   */
+  const [currentVespaSearchSelectedCategory, setCurrentVespaSearchSelectedCategory] = useState(themeConfig.categories.options[0].label);
```
jesskelsall, line 227: `themeConfig.categories` isn't set on all apps - on MCF and CCC where there is only one category ("All"), this property is omitted. I'd suggest referencing `documentCategories` and in particular `defaultDocumentCategory` when initialising this state variable.

As a rule of thumb, `themeConfig.categories` is for rendering the category search filters, whereas `themeConfig.documentCategories` is for showing documents by groups. I wish these were just one variable but they're not. 🤷‍♀️

jesskelsall (changes requested):
Needs some work to handle MCF and CCC apps where there is no categories config variable.

jesskelsall (approved):
My original comment has been addressed, so I'm happy to approve this now while any last tweaks go in. 🙂

src/components/pages/geographyOriginalPage.tsx
```diff
@@ -217,6 +221,48 @@ export const GeographyOriginalPage = ({ geography, summary, targets, theme, them
...
 
+  /** Vespa search results */
+  const vespaSearchTabbedNavItems = themeConfig.categories
+    ? themeConfig.categories.options.map((category) => {
+        return {
+          title: category.label,
+          /** We need to maintain the slug to to know what to send to Vespa for querying. */
+          slug: category.slug,
+          // TODO: Make this work
+          count: 0,
```
PatFawbertMills, line 232: There is an attribute on the search result object called `"total_family_hits"` that I believe we should be able to use for this number?

Only issue is retrieving the number _before_ we select the category though 🤔

src/components/pages/geographyOriginalPage.tsx
```diff
@@ -217,6 +221,48 @@ export const GeographyOriginalPage = ({ geography, summary, targets, theme, them
...
+    : /** We generate an `All` for when themeConfig.categories are not available e.g. MCFs */
+      [
+        {
+          title: "All",
+          slug: undefined,
+          // TODO: Make this work
+          count: 0,
+        },
+      ];
+  const [currentVespaSearchSelectedCategory, setCurrentVespaSearchSelectedCategory] = useState(vespaSearchTabbedNavItems[0].title);
```
PatFawbertMills, line 244: The default starting category is also stored within the config under the attribute `defaultDocumentCategory`. The intention here was to ensure this category is selected by default when navigating to the geo page. For the CPR app for instance, this is "Laws"

src/components/pages/geographyOriginalPage.tsx
jamesgorrie: Yea, I've left this as a "V2" to come up with a solution that works out all counts at once.

PR #727, opened 2025-08-28
Feature/app 987 Add methodology page for CCC

katybaulch:
# What's changed

Update CCC methodology page with first draft.

## Why?

It had CCLW content before and now it has draft content so we can see what it looks like and make adjustments

## Screenshots?

image

image


jesskelsall (approved):
Looks good. It'd be nice to use a map over the FAQs but I'm currently very happy with a "just get it on the page" mentality. 👍  Approved!

PR #728, opened 2025-08-28
APP-1042: Geo page: Targets block

jesskelsall:
# What's changed

- Added `TargetsBlock` component:
  - Renders all of the given targets in the order they are provided.
  - Links to the source on CCC go to the CPR app (as these policies don't exist in the CCC case list).
- Updated `getGeographyPageSidebarItems` to hide targets if there are none to display.
- `Section` component no longer has a `ReactNode` title but a title string and a count number (more opinionated props).

## Why?

Satisfies [APP-1042](https://linear.app/climate-policy-radar/issue/APP-1042/geo-page-targets-block).

## Screenshots

Screenshot 2025-08-28 at 11 19 01


src/components/blocks/subDivisionBlock/SubDivisionBlock.tsx
```diff
@@ -27,12 +22,12 @@ const get3ColumnClass = (index: number, length: number): string => {
...
 };
 
-export const SubDivisionBlock = ({ title, id, subdivisions }: TProps) => {
+export const SubDivisionBlock = ({ subdivisions }: TProps) => {
   if (!subdivisions || subdivisions.length === 0) {
     return null;
   }
   return (
-    <Section title={title} id={id}>
+    <Section id="section-subdivisions" title="Geographic sub-divisions" count={subdivisions.length}>
```
jamesgorrie, line 30: If Subdivision block is included multiple times (which I assume it is given it's name?), wouldn't this break as you can't have two matching ids?

jesskelsall: This will only ever be on any given page once.

jamesgorrie: Maybe for later - but maybe worth calling `GeographicSubdivisionBlock` or similar

jamesgorrie (approved):
very clear!

PR #731, opened 2025-09-01
Feature/app 1085 world map hide show aggregated eu data checkbox

katybaulch:
# What's changed

Hide show aggregated EU checkbox in World Map by default

## Why?

There is a bug that means that the counts of families is not aggregated for EU member states.

## Screenshots?

image


PatFawbertMills (approved):
Nice one

PR #732, opened 2025-09-01
Feature/app 1088 allow slideouts to be optionally opened on search page

katybaulch:
# What's changed

Allow slideouts to be optionally opened on search page.

Works for all slideout values in the type `TSlideOutContent`, which is derived from the values in the `SLIDEOUT_VALUES` constant in `/src/constants/slideOutValues.ts`

## Why?

So when redirecting, we can get the user closer to the filters they wanted to apply originally

## Screenshots?

image


jamesgorrie (commented):
suggestion: Should we be attributing the areas of the page that the hash refers to? e.g.
```html
<!-- pseudo code -->
<div class="slideout" id="concepts">
<!-- ... -->
</div>
```

jamesgorrie:
@PatFawbertMills @katybaulch - [hashes are used extensively to target fragments of a page](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Fragment).

If we just stick to this simple definition and keep use that API as a starter - would we need a hash manager?

Granted some would have special powers, but an API shaped similar to this might be less custom and simpler?

```tsx
// over-simplified Pseudo code
function openDrawer(id: string) {
   [...document.querySelectorAll('.drawer')].forEach(elem => elem.classList.remove('open'));
   document.getElementById(id).classList.add('open');
}

<a href="#principal-laws" onClick={openDrawer('principal-laws')}> Principal laws</a>

{/* ... */}

<div id="principal-laws" class="drawer">
{/* ... */}
</div>

```

This could then work for tabs etc, or whenever we want to target an element on the page.

katybaulch:
> suggestion: Should we be attributing the areas of the page that the hash refers to? e.g.
> 
> ```
> 
> <div class="slideout" id="concepts">
> 
> </div>
> ```

This is already happening via the `data-slideout` attribute

PR #734, opened 2025-09-02
feat: sunset geography v1 on litigation geography page

jamesgorrie:
# What's changed

- sunsets the geography V1 on the litigation geo page

## Why?

[The Geographies API now returns statistics in a performant way, so we should use that.](https://api.climatepolicyradar.org/geographies/united-states)


PatFawbertMills (approved):
lgtm

PR #736, opened 2025-09-02
Create dependabot.yml

katybaulch:
# What's changed

- Readd dependabot.yml
- Group minor and patch updates into one PR
- Raise separate PR for each major version bump

This will also autogenerate the lockfile updates


annaCPR (approved):
lgtm

PR #737, opened 2025-09-02
fix: add a little extra info to deploy labels

jamesgorrie:
# What's changed

- adds `sha` and `theme` infor when deploying to production & staging 
- removes the old deploy script that no longer works and isn't how we want to progress


katybaulch (approved):
lgtm - great adds for visibility!!

PR #739, opened 2025-09-02
Feature/app 1056 return list of subdivisions with data

annaCPR:
# What's changed

Updated useSubdivisions() hook, which calls `/families/aggregations/by-geography`, to include new filters:
'documents.document_status'  - filters by 'published' documents
'corpus.import_id' - filters by corpora listed in theme config so that we're only counting families for the relevant corpora

## Why?
So that we get accurate counts of published families per geography for the relevant corpora according to the theme.


PatFawbertMills (approved):
Neat 🖊️

PR #756, opened 2025-09-08
Feature/app 977 landing page about section

katybaulch:
# What's changed

Add landing page about section for CCC

NOTE: Just adding the text based off the MCF design for now -- whoever is doing the landing page will need to format according to whatever the design is.

## Why?

Didn't exist

## Screenshots?

image


PatFawbertMills (approved):

themes/mcf/pages/homepage.test.tsx
katybaulch, line 1: we're deleting this file because we already have MCF homepage tests that are more built out - this is duplication

themes/ccc/styles/styles.scss
katybaulch, line 1: we're deleting this file because it was copy pasted from CCLW theme and at the top of the file there is a note that the file isn't being used anymore -- so this is junk

PR #758, opened 2025-09-08
Deduplicate event list when events a re linked to both family and document

annaCPR:
# What's changed

Fixes a duplication issue on the collection page, where both the 'Cases' and 'Events' views showed duplicated entries for all events/documents. 

## Why?
This was because when fetching all the data for a family within a collection, we simply collated all the family events and document events into a single list to be displayed on the page. However, many events are linked both to the family AND a document so they needed to be deduplicated before the list was rendered.

## Screenshots?
Before:
Screenshot 2025-09-08 at 12 27 49

After:
Screenshot 2025-09-08 at 12 27 29


src/utils/eventTable.tsx
```diff
@@ -74,10 +74,16 @@ export const getCourts = (family: TFamilyPublic): string | null =>
     .join(" / ") || null;
 
 const getFamilyEvents = (family: TFamilyPublic): TEventWithDocument[] =>
-  [
-    ...family.events.map((event) => ({ event })),
-    ...family.documents.map((document) => document.events.map((event) => ({ event, document }))).flat(1),
-  ].flat(1);
+  Array.from(
```
PatFawbertMills, line 77: This is a pretty gnarly bit of code, personally I'd prefer to break down the solution to make it easier to read and understand what is going on and why we are doing it (captured in code).

jesskelsall (approved):
Paired with Anna on tweaking the event uniqueness code, and now I'm happy this is good to go. Approved.

PR #768, opened 2025-09-08
feat: use search API on Litigation geography page

jamesgorrie:
# What's changed
- fetches the initial data, the first documents list, on server, if litigation is enabled
- ~fetches the counts for (not sure this is necessary as the count is only necessary on opening the accordion)~ we didn't need this as the count loads on accordion open. We can always reinstate this if we need it.
- fetches the data for other accordions when you open them
- caches any loaded data (not sure this is necessary as we use `react-query`[^1])

## Why?
So we can sunset the `summaries` endpoint

## notes 

The UI could use some work - but we can update that once we're happy with the data fetching. It's a little slower than the current accordion, so felt OK as an initial pass.

## Screenshots?

https://github.com/user-attachments/assets/74d059a0-238c-41f5-8c68-883c5fe84c2d

[^1]: I tried this, but the caching didn't work so well and the caching didn't seem to catch as many use cases, so have opted for the simple version
```ts
  const { isPending, error, data } = useQuery({
    queryKey: ["repoData", lastSelectedCategory],
    queryFn: async ({ queryKey }) => {
      const [_key, category] = queryKey;
      const searchQuery = buildSearchQuery({ l: geographyV2.slug, c: category }, themeConfig);
      const search: TSearch = await backendApiClient
        .post("/searches", searchQuery, {
          headers: {
            accept: "application/json",
            "Content-Type": "application/json",
          },
        })
        .then((response) => response.data);

      setSearchResultsByCategory((currentFamilies) => ({
        ...currentFamilies,
        [category]: search,
      }));
    },
  });
```


jamesgorrie:
@jesskelsall I noticed [this comment](https://github.com/climatepolicyradar/navigator-frontend/blob/bca9bd5069b8584e1a4bb9214ac40cbd669fdd1c/src/components/blocks/recentFamiliesBlock/RecentFamiliesBlock.tsx#L17) - this implementation would allow you to do either.

katybaulch (changes requested):
All good to me - apart from enabling litigation on CPR. We're not ready for that quite yet as I understand it

katybaulch (approved):
Nice work!

PR #770, opened 2025-09-09
Comment out all instances of the debug component and remove from sidebar

PatFawbertMills:
# What's changed
- Commenting out instances of the debug component that we were using for debugging

## Why?
- Don't want to confuse any external users
- Didn't have time to incorporate a feature-flagged solution and will reinstate after any immediate demos


jesskelsall (approved):
LGTM

PR #775, opened 2025-09-09
fix: sunset summaries endpoint

jamesgorrie:
# What's changed

- removes all references to the geography summary endpoint and it's data

## Why?

- so we can move forward using more understood + performant endpoints


PatFawbertMills (approved):
Couple of questions but by no means blocking.

PR #778, opened 2025-09-09
Feature/app 1109 update breadcrumbs and add to ccc

katybaulch:
# What's changed

Updated breadcrumbs across all sites

## Why?

Because we wanted to update the breadcrumbs across all sites, including adding subdivison support

### CCC

image

image

image

image

image

### CCLW

image

image

image

image

### CPR

image

image

image

image

### MCF

image

image

image


package.json
```diff
@@ -34,6 +34,7 @@
     "posthog-js": "^1.261.1",
     "react": "18.3.1",
     "react-dom": "18.3.1",
+    "react-icons": "^5.5.0",
```
PatFawbertMills, line 37: Request: any way to avoid adding the entire react icons package? I have a distant memory of @jesskelsall and I attempting to remove this before

katybaulch: I didn't realise we were use Lucide - have updated

conordelahunty:
@katybaulch Just saw the screenshots now. the breadcrumbs should be above the heading on then CCC app.

src/components/pages/familyLitigationPage.tsx
```diff
@@ -61,6 +63,37 @@ export const FamilyLitigationPage = ({ countries, subdivisions, family, theme, t
...
+
+      breadcrumbGeography = { label: countryName, href: `/geographies/${countrySlug}` };
+      breadcrumbSubGeography = { label: subdivisionName, href: `/geographies/${subdivisionSlug}` };
+    } else {
+      const countryCode = firstGeography.split("-")[0];
+      const countrySlug = getCountrySlug(countryCode, countries);
+      const countryName = getCountryName(countryCode, countries);
+
+      breadcrumbGeography = { label: countryName, href: `/geographies/${countrySlug}` };
+      breadcrumbSubGeography = { label: subdivisionName, href: `/geographies/${subdivisionSlug}` };
```
PatFawbertMills, line 93: Is there a chance that we are setting the sub geo breadcrumb here, but we may not have subdivision data?

katybaulch: No - we're trying to set it if we have it - if not, the Breadcrumb component itself will not render that part of the breadcrumb if its value is falsey

katybaulch:
> @katybaulch Just saw the screenshots now. the breadcrumbs should be above the heading on then CCC app.

Done.

PR #811, opened 2025-09-15
Feature/app 1139 fe build about section on CCC homepage

katybaulch:
# What's changed

- Add powered by CPR section on homepage
- Driveby added label for world map as per design

Note: images for the cards are missing as we don't have them from Conor yet. Will need to add these later.

## Why?

https://www.figma.com/design/Lg0Vfj7EbUUz8HeqLWRZuK/H2-2025?node-id=1184-2&p=f&m=dev

## Screenshots?

image


themes/ccc/components/Feature.tsx
```diff
@@ -1,44 +1,66 @@
...
+          {subheading}
+        </Heading>
+      )}
+      {children}
+    </>
+  );
+
+  if (inline) {
+    return (
+      <Card heading={heading || ""} extraClasses="h-full" headingClasses="text-blue-900 text-2xl font-bold">
```
PatFawbertMills, line 35: We are moving away from the old blue colour set, think this is close enough to the `text-brand-darker`. See new colours in the `theme.css` file at the bottom.

themes/ccc/components/FeatureDiscover.tsx
```diff
@@ -1,34 +1,35 @@
...
+        <LinkWithQuery href="/about" className="text-blue-900 hover:text-blue-800 underline text-lg font-medium">
+          Read more →
+        </LinkWithQuery>
+      </Feature>
+      <Feature heading="Get the newsletter" subheading="Stay up to date" inline={true}>
+        <p className="text-lg text-gray-700 leading-relaxed mb-6">
+          Subscribe to the Sabin Center Climate Litigation Newsletter for twice-monthly updates. Each issue includes the latest case updates, event
+          announcements, and publication highlights.
+        </p>
+        <ExternalLink
```
PatFawbertMills, line 26: The links in the design look to be not underlined and to be bold. Also rather than the blue colour set could we use the `text-brand-darker` colour?

PatFawbertMills: Also text size 16px, which is base so we shouldn't need to specify size unless we have on a parent container

jesskelsall (approved):
Looks good following all of your changes! Approved

PR #817, opened 2025-09-16
Add CCC Google Analytics info

katybaulch:
# What's changed

Add CCC Google Analytics

## Why?

We copied and pasted from CCLW for the CCC template, so the Google analytics were pointing at the CCLW GA


PatFawbertMills (approved):
lgtm - just double checked we can use gtag with Google Analytics 4 (we know this because the code starts with `G-`)

PatFawbertMills (approved):
lgtm

PR #825, opened 2025-09-18
Feature/app 1190 Clarify the count on search results

katybaulch:
# What's changed

- Add search result count label for litigation app (configurable in theme config)
- Driveby fix by reverting the change to the stats block in the geo page rendering

## Screenshots?

Note the 151 cases at the top 

image


katybaulch:
> Adding a variable to config for this sort of divergent per-app copy change is not going to be scalable. I have a hollistic solution for this which I'm keen to build this afternoon.
> 
> On the basis that this just needs to look right today for litigation launch, I'll approve this on the proviso that the implementation will be replaced shortly.

Sounds good - let's pair :)

PR #827, opened 2025-09-18
fix: move to collections and slugs

jamesgorrie:
# What's changed
- moves collection => collections to sync with APIs
- uses slugs to sync with other pages

## Why?

Consistency is key. If we make a change, this will make it easier to do universally.

### DIFF
--- a/src/pages/collections/[id].tsx
+++ b/src/pages/collections/[id].tsx
@@ -105,13 +105,15 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
     return { notFound: true };
   }
 
-  const import_id = context.params.id;
-  const apiClient = new ApiClient(process.env.CONCEPTS_API_URL);
-
   let collectionData: TCollectionPublicWithFamilies;
 
   try {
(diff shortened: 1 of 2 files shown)


src/pages/collections/[id].tsx
```diff
@@ -105,13 +105,15 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
...
-  const import_id = context.params.id;
-  const apiClient = new ApiClient(process.env.CONCEPTS_API_URL);
-
   let collectionData: TCollectionPublicWithFamilies;
 
   try {
-    const { data: collectionResponse } = await apiClient.get(`/families/collections/${import_id}`);
+    const apiClient = new ApiClient(process.env.CONCEPTS_API_URL);
+    const id = context.params.id;
+    const { data: slugData } = await apiClient.get(`/families/slugs/${id}`);
```
Comment (PatFawbertMills) [inline src/pages/collections/[id].tsx:113]: Side note: kind of weird we call `/families/slugs/` with the collection ID rather than the family ID but no blocker, just threw me off a bit

PR #828, opened 2025-09-18
APP-1199 Add control to clear other values when interacting with AppliedFilters

PatFawbertMills:
# What's changed
- When a user selects to remove a root concept from the list of `AppliedFilters` we now clear all the child filters, to prevent unwanted behaviour

## Why?
- Bug that allowed root concepts to be removed but leaving the child concepts, which is invalid in the new structure

## Screenshots?
https://github.com/user-attachments/assets/4f3bd89b-1230-4019-82ca-beeca1f1f21f


jamesgorrie (approved):
neato!

PR #858, opened 2025-09-25
fix: add isCorpusIdAllowed

jamesgorrie:
# What's changed
- adds `isCorpusIdAllowed` to either 404 if a `family.corpus.import_id` isn't in the `allowed_corpora_ids` 

## Why?

Given SEO sensitivity - this felt prudent while we discuss attribution and the like.
This is an incomplete abstraction - but simple enough to understand.

This will be part of the APP_TOKEN conversation that will happen soon.

### DIFF
--- a/src/pages/document/[id].tsx
+++ b/src/pages/document/[id].tsx
@@ -126,7 +126,7 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
   }
 
   /** Check the family is in the "allowed_corpora" */
-  if (!isCorpusIdAllowed(process.env.BACKEND_API_TOKEN, familyData.corpus.import_id)) {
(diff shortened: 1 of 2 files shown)


src/pages/document/[id].tsx
```diff
@@ -124,6 +125,13 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
     };
   }
 
+  /** Check the family is in the "allowed_corpora" */
+  if (!isCorpusIdAllowed(process.env.BACKEND_API_TOKEN, familyData.corpus.import_id)) {
```
Comment (PatFawbertMills) [inline src/pages/document/[id].tsx:129]: Smart

PatFawbertMills (approved):
I say again: smart.

PR #860, opened 2025-09-25
refactor: refactor family page to use families API

jamesgorrie:
# What's changed
- refactors the family page datas fetching method to use the families-api only
- refactors out the use of `FamilyPage` in favour of `TFamilyPublic`

## Why?
- simple. simplicity.


src/pages/document/[id].tsx
Comment (jamesgorrie) [inline src/pages/document/[id].tsx:1]: Here be the 🍖 of the PR

PR #866, opened 2025-09-29
APP-1181: Can we change the text on sorting card for document type (table)

jesskelsall:
# What's changed

- Moved interactive table types into their own file and converted a few interfaces to types.
- `InteractiveTable` `sortOptions` prop:
  - Defaults to the current ascending/descending pair of options.
  - Can be used to specify custom labels, orders, or omit either `asc` or `desc` sort option.
- Tables of events' _Type_ column has one sort option: "Group by type".
- Fixed tests.

## Why?

Satisfies [APP-1181](https://linear.app/climate-policy-radar/issue/APP-1181/can-we-change-the-text-on-sorting-card-for-document-type-table).

## Screenshots?

Screenshot 2025-09-29 at 14 17 31


PatFawbertMills (approved):
Praise: I was skeptical we needed to make this change, but you've implemented it really well and I've changed my opinion because of that. Nice work

PR #867, opened 2025-09-29
Feature/app 1246 Remove links to international and no geo 

katybaulch:
# What's changed

Removed links to international and no geo:
- In breadcrumbs
- On family page metadata block
- On family page header

## Why?

Because we don't have geo pages for these, so we get 404 errors.


src/components/pages/familyLitigationPage.tsx
```diff
@@ -101,17 +102,21 @@ export const FamilyLitigationPage = ({ countries, subdivisions, family, theme, t
...
+
+            return !isSystemGeo(name) ? (
+              <LinkWithQuery key={code} href={`/geographies/${slug}`} className="underline">
+                {name}
+              </LinkWithQuery>
+            ) : (
+              <span key={code}>{name}</span>
+            );
+          })
+          .filter(Boolean),
```
PatFawbertMills, line 119: Quesiotn: do we need this `filter`?

PR #872, opened 2025-10-01
Re-hydrate the table rows in the client to display dates in local format

PatFawbertMills:
# What's changed
- Wait for page to load before updating the table rows to display dates in the client's language format

## Why?
- Prevents console errors caused by hydration errors cause by the client and server differing
- Displays dates in client's preferred format


jesskelsall (approved):
Looks decent with this approach, particularly with the language being passed into the row generating function. Happy to approve.

PR #885, opened 2025-10-06
APP-1249: Add geos and sub-geos to the generated sitemap

jesskelsall:
# What's changed

- Added constants for countries data and removed equivalent literal declarations.
- Refactored sitemap page:
  - Less individual functions for better readability.
  - Filtered out any geography node which we don't serve a geo page for (custom ISO codes, regions).

## Why?

Satisfies [APP-1249](https://linear.app/climate-policy-radar/issue/APP-1249/add-geos-and-sub-geos-to-the-generated-sitemap).

### DIFF
--- a/src/pages/sitemap.txt.ts
+++ b/src/pages/sitemap.txt.ts
@@ -1,37 +1,39 @@
 import { ApiClient } from "@/api/http-common";
+import { EXCLUDED_ISO_CODES, INCLUDED_GEO_TYPES } from "@/constants/geography";
 import { TDataNode, TGeography } from "@/types";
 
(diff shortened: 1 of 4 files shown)


annaCPR (approved):
lgtm

odrakes-cpr (approved):

src/pages/sitemap.txt.ts
jamesgorrie, line 1: The sitemap that's being used is https://github.com/climatepolicyradar/navigator-frontend/blob/main/src/app/ccc/sitemap.ts.

We should be removing this as deprecated.

PR #890, opened 2025-10-07
APP-1168: Map: countries with 0 cases and countries with 1 case should have different colour

jesskelsall:
# What's changed

- Updated the formula for what colour to shade a world map country as:
  - Countries with no entries have a static HSL lightness value.
  - Lowered the darkness of the lowest count value and increased the contrast to the highest value country.

## Why?

Satisfies [APP-1168](https://linear.app/climate-policy-radar/issue/APP-1168/map-countries-with-0-cases-and-countries-with-1-case-should-have).

## Screenshots?

Screenshot 2025-10-07 at 15 37 44


annaCPR (approved):
very nice

PR #1218, opened 2026-04-09
fix: use labels.value.value for filter

jamesgorrie:
# What's changed
- uses `labels.value.value` for the filter value, based on [this PR](https://github.com/climatepolicyradar/search/pull/239)

## Why?

- so that we can reimplement the `labels.value.id` filter properly, and then move the FE over
- this will avoid selecting a filter value across type e.g. `Law` will select `entity_type::Law` and `category::Law`

Towards https://linear.app/climate-policy-radar/issue/APP-1950/make-document-label-search-use-typevalue-instead-of-just-value


PatFawbertMills (approved):
lgtm

PR #1221, opened 2026-04-11
APP-1949 - search results documents toggle display

PatFawbertMills:
# What's changed
- Adding toggle for searching by either principal or documents
- Update search results to reflect if item is a principal or a document

## Screenshots?
Screenshot 2026-04-11 at 14 50 35


conordelahunty:
Can I quick check expected behaviour.

When the the toggle is in an "on" position the user should see both Principal and non principal results. When the toggle is "off" then the user should just see principal results

?

PatFawbertMills:
> Can I quick check expected behaviour.
> 
> When the the toggle is in an "on" position the user should see both Principal and non principal results. When the toggle is "off" then the user should just see principal results
> 
> ?

Ah, no that's not what I've done. Currently "on" position = everything that is _not_ Principal, i.e. only Documents. "off" position = everything that is Principal.

katybaulch (approved):
nice work!

PR #1223, opened 2026-04-13
Make documents toggle on by default

PatFawbertMills:
# What's changed
- Use`useQueryState` to save state of the toggle and default to being on


katybaulch (approved):
nice 🧹

PR #1224, opened 2026-04-13
Initial instructions to work with AI agents

PatFawbertMills:
# What's changed
- Add some Github copilot instructions MD file
- Ensures we all have a similar experience and output from copilot 
- See here for documentation https://docs.github.com/en/copilot/concepts/prompting/response-customization#about-repository-custom-instructions 
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -


AGENTS.md
```diff
@@ -0,0 +1,56 @@
...
+- The UI should be clean and intuitive, with a focus on usability and
+  accessibility.
+- When generating new components, do not apply too many tailwind classes, allow
+  for developer discretion.
+
+### Types
+
+- Always define types for component props, even if the props are simple. This
+  promotes consistency and makes it easier to add additional props in the future
+  if needed. Use the syntax `TProps` for component props.
```
katybaulch, line 28: Worth putting here to prefix interfaces with I?? so ISomething?

AGENTS.md
katybaulch, line 1: Might be worth working on a section to do with testing here too :))

AGENTS.md
```diff
@@ -0,0 +1,56 @@
+# Project Overview
+
+This project is a web application that serves as the front-end for the
+`navigator` system. It provides a user interface for searching and navigating
+through data, with support for various themes and configurations. The
+application is built using Next.js, React, and Tailwind CSS, and is designed to
+be run in a Docker container for ease of deployment and development.
```
henry-cpr: Tokens are very precious with this sort of file!

```suggestion
This is the front-end for the `navigator` system. It is used for searching and navigating
through law and policy data. The application is built using Next.js, React, and Tailwind CSS and deployed as a docker container.
```

(the for ease of... ) is fluff LLM-wise

AGENTS.md
```diff
@@ -0,0 +1,56 @@
...
+
+The core functions of the application is to allow users to perform searches of
```
henry-cpr: Is this paragraph relevant and necessary?

AGENTS.md
```diff
@@ -0,0 +1,56 @@
...
+
+- `/src`: Contains the source code for the frontend.
+- `/themes`: Contains specific files relating to each of the themes, this
+  includes components, pages, constants and redirects
+- `/tests`: Contains test files for the application.
+
+## Libraries and Frameworks
+
+- Typescript for type safety and improved developer experience.
+- Next, React and Tailwind CSS for the frontend.
```
henry-cpr: Said this already!

AGENTS.md
```diff
@@ -0,0 +1,56 @@
...
+- Playwright for end-to-end testing.
+- Vitest for unit testing.
+- Trunk.io for code quality and formatting.
+
+## UI guidelines
+
+### General
+
+- The UI should be clean and intuitive, with a focus on usability and
+  accessibility.
```
henry-cpr: These need defining more specifically imo. I'd move them to a separate "docs/ui-guidelines.md" file that can be referenced here, and then an agent can choose to read it or not as necessary (saving tokens and reducing distracting information). But also clean, intuitive, usability, accessibility, these are not well defined or precise enough for LLMs. What do you eman by those? Are there specific criteria that an LLM can use to cleanly judge "is usable", "is not usable", "is intuitive", "is not intuitive", and so on? If there aren't, scratch it. If there are, put them in detail in a separate file

AGENTS.md
```diff
@@ -0,0 +1,56 @@
...
+- When generating new components, do not apply too many tailwind classes, allow
+  for developer discretion.
+
+### Types
+
+- Always define types for component props, even if the props are simple. This
+  promotes consistency and makes it easier to add additional props in the future
+  if needed. Use the syntax `TProps` for component props.
+- When generating types for use within components, keep the type definition in
```
henry-cpr, line 29: praise: this is a good example of a specific, idiomatic, precise rule that is the sort that should be in this sort of file :)

AGENTS.md
```diff
@@ -0,0 +1,56 @@
...
+  the same file as the component. Allow the developer to move the type
+  definition if they feel it is necessary, but this should be the default
+  behaviour.
+
+### Utilities
+
+- Avoid bloating components with too much logic - use a utility function if the
```
henry-cpr: bloating is too imprecise. What counts as too much logic? Maybe this might be reframed as something like components should be presentation-layer only. If more than a single layer of conditional logic is required, or logic is being repeated, or ..etcetc...

henry-cpr (commented):
Really glad you're looking at this! Have left a few tweaks that should be helpful for LLM interpretability, but it's also worth looking at https://arxiv.org/abs/2602.11988 -- these files aren't always helpful by default. What I love about them is that it helps us document and encode precise idioms and approaches that we use in the codebase. But LLMs get REALLY screwed by even slightly-irrelevant info in the context window, so I'd recommend keeping this as light and tight as possible (like, maybe even 10 lines max), and being more a map of contents (eg for information on UI, see ui-guidelines.md, for information on how to correctly use types, see types.md, etc etc). That way the agent can decide whether its worth bloating the context window with the additional data.

AGENTS.md
PatFawbertMills: An interesting point because I honestly don't know what our guidelines are right now. 

For the experimental search work for example, I don't think we need or want any tests to be written right now until we are more confident with the solution.

PatFawbertMills:
Based on the above discussion and chats in slack I have rewritten the file and tried to remove a lot of the chaff that the Github Copilot documentation suggested I add.

henry-cpr (approved):
much tighter!

PR #1233, opened 2026-04-16
Remove taxonomy debug from search result

jesskelsall:
# What's changed

Removes a leftover debug string added to search page family results as part of testing the new taxonomy attribute.


PatFawbertMills (approved):
lgtm

PR #1236, opened 2026-04-16
Remove cloudfront package

PatFawbertMills:
# What's changed
- This is flagging in our security flags in github
- We are importing it in a js file in our infra folder, but I think this may not be using the package


PatFawbertMills:
This isn't required I don't think anymore.

PR #1235, opened 2026-04-16
Add ICCN reports to search on CPR

PatFawbertMills:
# What's changed
- Adding ICCN reports to CPR search

PR #1237, opened 2026-04-16
Revert "Add ICCN reports to search on CPR"

PatFawbertMills:
Reverts climatepolicyradar/navigator-frontend#1235

Decision from Dominyka is to wait until we have run the concept classifiers on the new corpus.


jesskelsall (approved):
Approved

PR #1244, opened 2026-04-20
fix: remove filter by label.value

jamesgorrie:
# What's changed

- removes the ability to filter by `label.value`

## Why?

Allowing filtering by non-explicit (vespa-speak: attribute) styles values causes way more confusion than it's worth. We should be using free-text search for that. 

AKA: Filtering should always be done on IDs.


PatFawbertMills (approved):
Nice catch

PR #1246, opened 2026-04-20
Remove counts from search results

PatFawbertMills:
# What's changed
- Remove counts as it is proving to cause more confusion than clarity at this moment.


katybaulch (approved):
nice & smol ✅

PR #1247, opened 2026-04-20
Feature/app 1891 Add sorting controls to shadow search

katybaulch:
Rustoleum


src/api/search.ts
```diff
@@ -54,15 +54,57 @@ export interface SearchDocumentsResponse {
...
+  return "relevance";
+}
+
 interface SearchDocumentsParams {
   query?: string;
   filters?: TQueryGroup;
   page_size?: string;
   page_token?: string;
   includeDocumentsInSearch?: boolean;
+  /** Result ordering; defaults to relevance when omitted */
```
PatFawbertMills, line 96: Nitpick: I'm presuming Claude has added these comments? we should be careful because if we ever change the default this comment will end up being way more of a problem than simply not bothering to have one - plus it is easy to deduce the default from the normalise function above

katybaulch: Yeah -- a lot of this was auto generated last week as I was trying to get a full e2e setup including local search/vespa with a local frontend. Gotta go back through a lot of this stuff to pull out only what we need (hence draft mode!)

src/api/search.ts
```diff
@@ -54,16 +54,47 @@ export interface SearchDocumentsResponse {
...
   page_token?: string;
   includeDocumentsInSearch?: boolean;
+  sort?: SearchDocumentsSortKey;
   excludeMergedDocuments?: boolean;
 }
 
-const SEARCH_DOCUMENTS_BASE_URL = "https://api.climatepolicyradar.org/search/documents";
+/**
+ * Same API origin as `useLabelSearch` / `loadLabels` (`NEXT_PUBLIC_API_URL`).
+ * Defaults to production; set e.g. `http://localhost:8000` for local search API.
```
PatFawbertMills: Question: Do we need this comment? we don't really ever run the API locally so that localhost:8000 is quite confusing. This function seems pretty self explanatory without a comment.

src/api/search.ts
```diff
@@ -54,16 +54,47 @@ export interface SearchDocumentsResponse {
   };
 }
 
+export const SEARCH_DOCUMENT_SORT_KEYS = ["relevance", "recent", "oldest", "title_asc", "title_desc"] as const;
+
+export type SearchDocumentsSortKey = (typeof SEARCH_DOCUMENT_SORT_KEYS)[number];
+
+export function orderByParamFromSortKey(key: SearchDocumentsSortKey): string {
```
PatFawbertMills:
```suggestion
export function getSortParamFromKey(key: SearchDocumentsSortKey): string {
```

PatFawbertMills: Suggestion: this naming was confusing me a bit!

PatFawbertMills: do we need to wrap this in a function? vs. just referring to the map on line 159?

PatFawbertMills: if we change this from a function that returns a value, to:

`sortParams = Record<SearchDocumentsSortKey, string> = {
    relevance: "relevance desc",
    recent: "attributes.published_date desc",
    oldest: "attributes.published_date asc",
    title_asc: "title asc",
    title_desc: "title desc",
  };`

then on line 159 we can say:
`sortParams[sortKey]`

src/components/_experiment/searchSort/SearchSortSelect.tsx
```diff
@@ -0,0 +1,64 @@
...
+  }
+  const item = SORT_ITEMS.find((i) => i.value === value);
+  return item ? `Sort: ${item.label}` : "Sort";
+}
+
+const triggerClassName =
+  "inline-flex h-9 max-w-full min-w-0 items-center justify-between gap-2 rounded-full border border-transparent-regular bg-[canvas] px-4 py-1.5 text-left text-sm font-medium text-neutral-600 select-none transition-colors hover:border-inky-black focus-visible:outline-2 focus-visible:-outline-offset-1 focus-visible:outline-blue-800 data-popup-open:border-inky-black";
+
+const popupClassName =
+  "max-h-(--available-height) min-w-(--anchor-width) origin-(--transform-origin) overflow-y-auto rounded-lg border border-transparent-regular bg-[canvas] py-1 shadow-lg shadow-gray-200/80 outline-none transition-[transform,scale,opacity] data-ending-style:scale-95 data-ending-style:opacity-0";
```
PatFawbertMills: Praise: nice use of base vars and animations.

Suggestion: we haven't been setting the shadow colour for the other surfaces on the _search page. Assuming these are adding by Claude - be cautious as it can try to add excessive tailwind classes. Less is more!

src/pages/_search/index.tsx
```diff
@@ -42,6 +43,8 @@ const ShadowSearch = ({ theme, themeConfig, features }: TProps) => {
   // pagination state
   const [currentPage, setCurrentPage] = useQueryState("page_token", parseAsString.withDefault("1"));
   const [pageSize, setPageSize] = useQueryState("page_size", parseAsString.withDefault("10"));
+  const [sortParam, setSortParam] = useQueryState("sort", parseAsString.withDefault("relevance"));
+  const sort = normaliseSearchDocumentsSortKey(sortParam);
```
PatFawbertMills: Request: this feels unnecessary to normalise this value here when all we are doing is passing it along into the component, which we then do transformations on anyway.

PatFawbertMills (approved):
LGTM 🚢

src/api/search.ts
```diff
@@ -54,16 +54,42 @@ export interface SearchDocumentsResponse {
...
   includeDocumentsInSearch?: boolean;
+  sort?: SearchDocumentsSortKey;
   excludeMergedDocuments?: boolean;
 }
 
-const SEARCH_DOCUMENTS_BASE_URL = "https://api.climatepolicyradar.org/search/documents";
+function searchDocumentsUrl(): string {
+  const origin = (process.env.NEXT_PUBLIC_API_URL || "https://api.climatepolicyradar.org").replace(/\/$/, "");
+  return `${origin}/search/documents`;
+}
```
jamesgorrie, line 92: suggestion: this could just be a const as there is no need to compute it every time.

src/api/search.ts
```diff
@@ -111,19 +137,18 @@ function configureDocumentsFilters(
...
 
-  // This enables `bolding` in vespa AKA highlighting, which highlights the matched terms in the results.
-  url.searchParams.set("bolding", "true");
-
   if (params.query) url.searchParams.set("query", params.query);
   if (filters) url.searchParams.set("filters", JSON.stringify(filters));
   if (params.page_size !== undefined) url.searchParams.set("page_size", params.page_size);
   if (params.page_token !== undefined) url.searchParams.set("page_token", params.page_token);
+  const sortKey = params.sort ?? "relevance";
+  url.searchParams.set("order_by", SEARCH_DOCUMENT_SORT_PARAMS[sortKey]);
```
jamesgorrie, line 151: I still don't fully understand why we transform between the UI => API?

Is there a reason not to keep them the same to avoid having to remember the hoop we jump through?

jamesgorrie (commented):
boom kaboom, love it

jamesgorrie (approved):
Could we add a description in the PR for future travellers?

PR #1248, opened 2026-04-20
Clicking on labels in search result applies the label correctly

PatFawbertMills:
# What's changed
- Clicking on a label in the search result now applies the filter correctly, by using the ID of the label.

## Why?
- Wasn't working before as we had not switched it over from using the value


katybaulch (approved):
nice work!!

PR #1259, opened 2026-04-21
Upgrade any eslint packages to latest safest

PatFawbertMills:
# What's changed
- Core eslint package upgraded to the latest within v9
- Two smaller eslint packages updated to latest that is safe and compatible

## Why
- Housekeeping
- v10 not supported by some dependancies yet

## AI attribution
- Checking version dependancy alignment between the ESLint packages


CLAUDE.md
katybaulch, line 1: question: do we need to remove the old agents.md file?

katybaulch (approved):
yum, sorry for the pain these are awful toilet cleaning updates

CLAUDE.md
PatFawbertMills: I hadn't but I will now

PR #1262, opened 2026-04-21
Feat/screw you bots

henry-cpr:
# What's changed

- Added vCPU and memory as pulumi stack config parameters
- Upped production frontend containers to 4 vCPUs and 8GB memory 
- Added WAF ACL to cloudfront distribution that counts (without block or challenge) user agents with Chrome/125.0.0.0 in the user agent string

The production frontend apprunners have been updated and are using the new config. 

CPR-staging has a deployment of the WAF rules, but I do not want to update production until this code has had a good going over. 

Before merging, if we're happy with the code, we should finish deploying to staging services and confirm all working, and then update production services incrementally. 

## Why

We got hit hard by scraper bots, and they all use that string in their user agent. If they come back, we'll be able to turn the count into a challenge and serve them with a captcha. This PR gets the infra in so we can test and tune. 

The interaction of this with the CCC redirects is an open question for me and therefore I would appreciate careful eyes on. 

## AI attribution

Claude wrote the WAF code, heavily supervised by myself and Katy

### DIFF
--- a/infra/__main__.py
+++ b/infra/__main__.py
@@ -197,6 +197,8 @@ if not is_review_template:
             if is_cpr_stack
             else default_min_instances
         ),
+        cpu=config.require("apprunner_frontend_vcpu_count")
+        memory=config.require("apprunner_frontend_memory_gb"),
         auto_deploy=True,
     )
 
--- a/infra/__main__.py
+++ b/infra/__main__.py
@@ -197,7 +197,7 @@ if not is_review_template:
             if is_cpr_stack
             else default_min_instances
         ),
-        cpu=config.require("apprunner_frontend_vcpu_count")
+        cpu=config.require("apprunner_frontend_vcpu_count"),
         memory=config.require("apprunner_frontend_memory_gb"),
         auto_deploy=True,
     )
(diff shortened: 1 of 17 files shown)


PatFawbertMills:
i'm seeing lots of errors with the linter on the yaml files

infra/__main__.py
```diff
@@ -604,12 +614,12 @@ class DomainVisibility(Enum):
...
 
-        infra_dir = Path(__file__).parent.parent
-        with open(infra_dir / "redirects.json", "r") as f:
+        infra_dir = Path(__file__).parent
+        with open(infra_dir / "lambda_code" / "redirects.json", "r") as f:
             redirects: list[dict[str, str]] = json.load(f)["redirects"]
 
         # Get the code as a string for the CloudFront Function
-        with open(infra_dir / "redirection.js", "r") as f:
+        with open(infra_dir / "lambda_code" / "redirection.js", "r") as f:
```
jamesgorrie, line 622: is this related? don't really care, but I am trying to be thorough and can't tell what it's doing?

jamesgorrie (approved):
This looks good. Validating my understanding, we are just counting traffic from this chrome version with the hope to identify if it's a bot, and then potentially captcha them?

infra/__main__.py
henry-cpr: Thoroughness good! We didn't look into the why, but the path was wrong before. Assume missed during a migration

henry-cpr:
> This looks good. Validating my understanding, we are just counting traffic from this chrome version with the hope to identify if it's a bot, and then potentially captcha them?

Correct. Textbook making the change easy and then if/when we get bots again, making the easy change to challenging them with captcha

PR #1265, opened 2026-04-23
Reinstate the THEME variable in the tsconfig base

PatFawbertMills:
# What's changed
- Fixing a bug with the dynamic theme in tsconfg base

## Why
- Ignoring themes on the other domain


katybaulch (approved):
lgtm

PR #1268, opened 2026-04-23
Add unit test for the tsconfig.base.json file on the critical alias

PatFawbertMills:
# What's changed
- Added a unit test for the tsconfig.base.json on the themes path alias

## Why
- This is a critical path that we recently caused a critical issue due to it being hardcoded to a specific theme

### DIFF
--- a/tsconfig.base.test.ts
+++ b/tsconfig.base.test.ts
@@ -1,7 +1,12 @@
 import config from "./tsconfig.base.json";
 
 describe("tsconfig.base.json", () => {
-  it("should not be hardcoding the theme alias path, and should contain the wildcard", () => {
-    expect(config.compilerOptions.paths["@/themes/*"]).toContain("./themes/__THEME__/*");
+  describe("the */themes alias", () => {
+    it("should only contain one path", () => {
+      expect(config.compilerOptions.paths["@/themes/*"]).toHaveLength(1);
+    });
+    it("should contain the wildcard", () => {
+      expect(config.compilerOptions.paths["@/themes/*"]).toContain("./themes/__THEME__/*");
+    });
   });
 });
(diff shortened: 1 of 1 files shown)


tsconfig.base.test.ts
```diff
@@ -0,0 +1,7 @@
+import config from "./tsconfig.base.json";
+
+describe("tsconfig.base.json", () => {
+  it("should not be hardcoding the theme alias path, and should contain the wildcard", () => {
```
annaCPR: Question: There are 2 requirements in the test name but only one assertion. Technically we're only asserting on the second. What would happen if you had 2 paths in the config, one with the wildcard and one with a hardcoded  theme? (is that possible)?

PatFawbertMills: Good point RE: test description.

As per JSON parsing, if duplicate keys the last instance will win.

annaCPR: In that case might be worth adding more test cases to cover that

annaCPR: So if we're only expecting one path and for that path to be the one with the wildcard, then this test is too broad, i.e. `.contains` versus `===`

PatFawbertMills:
Example of IDE informing the user of duplicate keys + failing test using last entry. I don't think we should be replicating this in the test, we only care about the final JSON file.

Screenshot 2026-04-23 at 13 55 21

tsconfig.base.test.ts
PatFawbertMills: I disagree on test cases for duplicate keys, but I agree on more explicit tests for this key. I have updated the test

PatFawbertMills: I've added a comment below documenting the IDE output for duplicate keys in a JSON file.

annaCPR: Nice one - that works.

annaCPR (approved):

PR #1269, opened 2026-04-23
fix: be bold by bolding

jamesgorrie:
# What's changed

- adds bolding to the [API request](https://github.com/climatepolicyradar/search/blob/main/api/routers.py#L39) by default

## Why

- highlight messages

## Screenshots

Screenshot 2026-04-23 at 13 33 04


PatFawbertMills (approved):
very bold of you James

PR #1270, opened 2026-04-23
Secret merge button

jamesgorrie:
# What's changed

- adds a small secret button (@conordelahunty approved ✅) that hides `status::Merged` documents

A merged document is
- a family that has a document that matches it's title exactly.
- a MCF family that has a document that is called "Project document"

The aim is for folks to be able to play around with this.

## Why

To highlight that a lot of our issues on duplication is actually data-quality, not modelling or ranking, that is the issue. 

e.g. a family with a document that both share the same title AKA - they are the same thing.

i.e. the family -> document model forced us into creating two records for a Law, Project etc, when one would resolve to a cleaner, easier to use and understand dataset.

Another non-search, higher level reason was planning for when we will consume other data-sources, where the model is very much 1-1 with a PDF and document record, we no longer have to perpetuate this convoluted pattern.

This is **_a_** reason why the model was designed this way from the start.

## Screenshots

https://github.com/user-attachments/assets/060b4ae9-fccd-4691-97a7-e9325ca8ea42

### DIFF
--- a/src/api/search.ts
+++ b/src/api/search.ts
@@ -60,40 +60,59 @@ interface SearchDocumentsParams {
   page_size?: string;
   page_token?: string;
   includeDocumentsInSearch?: boolean;
+  excludeMergedDocuments?: boolean;
 }
 
 const SEARCH_DOCUMENTS_BASE_URL = "https://api.climatepolicyradar.org/search/documents";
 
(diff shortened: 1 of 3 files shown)


PatFawbertMills (approved):
Nice experiment 👍

PR #1276, opened 2026-04-28
APP-1974: Apply transformations to document page

jesskelsall:
# What's changed

This PR applies data-in API data and BFF transformation to the document page.

- Generic `transformFamily` and `transformDocument` functions are page-agnostic.
- Reworked `DocumentSchema` to be recursive (documents have documents).
- Difference between different data-in API document types (family vs document vs collection) is made using `document.attributes` through 3 new attribute schemas e.g. `FamilyAttributesSchema` and corresponding validation functions e.g. `validateFamilyAttributes`.
- Added `getChildDocuments` and `getParentDocuments` for filtering out any given document's relations by hierarchy direction.
  - Attempted to demystify the relation type values through this as they are unintuitive and differ if litigation or not.
- Document page now uses data-in API data where possible.
- Document page shows debug information just like the family page.

## Why

Satisfies [APP-1974](https://linear.app/climate-policy-radar/issue/APP-1974/apply-transformations-to-document-page)

## AI attribution

Claude used for code analysis and planning but all code was hand-written.


src/bff/methods/getRelations.ts
```diff
@@ -0,0 +1,17 @@
+import { TDataInDocumentRelation, TDataInDocumentRelationType } from "@/schemas";
```
PatFawbertMills, line 1: Praise: really nice clean abstraction ✏️ neat

PatFawbertMills (approved):
LGTM keep that train rolling

PR #1277, opened 2026-04-28
Include ICCN corpus in the all category

PatFawbertMills:
# What's changed
- Adding ICCN corpus ID to the "All" category

## Why
- This category is used as part of a normal search when a category is not selected
- Not including ICCN in the all category meant that 0 passage results were being returned on ICCN reports unless the user included the category in the search - which is only possible if they come from the search page and had filtered by category beforehand


katybaulch (approved):
lgtm

PR #1279, opened 2026-04-29
fix: add published filter by default

jamesgorrie:
# What's changed
- adds `attributes.status=published` filter be default 

## Why
- to not show unpublished work to users

Dependent on https://github.com/climatepolicyradar/search/pull/272


src/components/_experiment/advancedFilters/AdvancedFilters.tsx
```diff
@@ -13,8 +13,8 @@ export type TQueryGroup = {
 };
 
 export type TQueryRule = {
-  field: "labels.value.id";
-  op: "contains" | "not_contains";
+  field: "labels.value.id" | "attributes.status";
+  op: "contains" | "not_contains" | "eq";
```
PatFawbertMills, line 17: whats the difference between `contains` and `eq`? in the UI i've been treated contains as "is"

jamesgorrie: `contains` is pretty much `eq`. I've avoided adding `eq` to the API to not have to try and explain the difference. I'll remove this as it's a mistake.

We should expect `contains` works as `eq` for now. On reflection, we probably should have just gone with `eq` as that is what we mean in all used instances, so happy to make that refactor later.

PatFawbertMills (approved):
lgtm

PR #1280, opened 2026-04-30
Shadow Search: Hook date filter controls up to API

katybaulch:
# What's changed

Add date filtering to searchfilters and applied filters

Don't add date filtering to advanced filters for now

## Why

Advanced filters are really complicated and we need more thought. Also, published_date isn't a label, but an attribute. That's a divergence from the current label based advanced filtering.


src/components/_experiment/appliedLabels/AppliedLabels.tsx
```diff
@@ -51,30 +73,40 @@ export function AppliedLabels({
...
+      )}
       {isFilterComplex(filters) ? (
-        <button
-          className="bg-white py-1 px-2 rounded-lg inline-flex gap-2 items-center border border-gray-300 hover:bg-gray-50"
-          onClick={() => onAdvancedClick?.()}
-        >
-          <SlidersHorizontal className="text-neutral-400 h-4 w-4" />
-          Advanced filters applied
-        </button>
+        <>
```
PatFawbertMills: Nitpick: don't think we need to add this fragment

src/components/_experiment/appliedLabels/AppliedLabels.tsx
```diff
@@ -51,30 +73,40 @@ export function AppliedLabels({
...
-            <AppliedLabel
-              key={i}
-              label={label}
-              type={type || ""}
-              onSelect={() => onSelectLabel?.(label, type || "")}
-              onRemove={() => onRemoveLabel?.(label)}
-            />
-          );
-        })
+        <>
```
PatFawbertMills: Again not sure we need to add this fragment. Did you add these for a particular reason?

src/components/_experiment/searchFilters/SearchFilters.tsx
```diff
@@ -47,8 +55,100 @@ type TProps = {
    * any aggregation‑based disabling.
    */
   query?: string;
+  dateRangeValue?: string | null;
+  onDateRangeChange?: (value: string | null) => void;
 };
 
+const DATE_RANGE_PRESETS: Array<{ value: TDateRangePreset; label: string }> = [
```
PatFawbertMills: Request: could we move the date filter into its own component? Otherwise we've doubled the size of the filters component due to one filter group.

src/components/_experiment/appliedLabels/AppliedLabels.tsx
```diff
@@ -51,30 +73,40 @@ export function AppliedLabels({
   onRemoveLabel?: (label: string) => void;
   onSelectLabel?: (label: string, type: string) => void;
   onAdvancedClick?: () => void;
+  dateRangeValue?: string | null;
+  onSelectDateRange?: () => void;
```
PatFawbertMills: Suggestion: I'm not sure we need another `onSelect` handler specifically for the date range. We should use the current `onSelectLabel("", "published_date")` should work

PatFawbertMills: We might need to review the naming conventions of these surfaces soon though, as strictly speaking this is the applied search controls vs. calling them labels everywhere. In this case the date is not a label - I am aware of that.

src/pages/_search/index.tsx
```diff
@@ -216,6 +229,14 @@ const ShadowSearch = ({ theme, themeConfig, features }: TProps) => {
                 setFilters((prev) => (prev ? removeLabelRule(prev, label) : createGroup()));
                 setCurrentPage("1");
               }}
+              onSelectDateRange={() => {
```
PatFawbertMills: Same as comment above but from the other side, we should just re-use the onSelectLabel handler we already have.

PatFawbertMills (commented):
Couple of minor points for some code cleanliness. Logic looks solid!

src/components/_experiment/appliedLabels/AppliedLabels.tsx
katybaulch: I think my IDE automatically added these ugh - will remove, ta for spotting

src/components/_experiment/appliedLabels/AppliedLabels.tsx
katybaulch: I think my IDE automatically added these ugh - will remove, ta for spotting

.trunk/trunk.yaml
katybaulch, line 1: Driveby fix for that gnarly linux/mac os package lock issue

src/components/_experiment/searchFilters/PublishedDateFilterSection.tsx
```diff
@@ -0,0 +1,105 @@
...
+};
+
+const DATE_RANGE_PRESETS: Array<{ value: TDateRangePreset; label: string }> = [
+  { value: "all_time", label: "All time" },
+  { value: "last_year", label: "Last year" },
+  { value: "last_5_years", label: "Last 5 years" },
+];
+
+/** Preset and custom year-range UI for published-date filtering in shadow search. */
+export function PublishedDateFilterSection({ value, onChange }: TProps) {
```
PatFawbertMills, line 23: Non-blocking context: this will expanded on in future to incorporate different _types_ of date, for example for MCF projects it will use the approval financial year as opposed to the published date.

Naming things is always a pain the butt! But if we can find a way to name things so that thet are both descriptive but also abstract enough to work in different conditions = winner chicken dinner. We can circle back to it when we get to that point.

PatFawbertMills (approved):
Good work!

PR #1281, opened 2026-04-30
APP-1928: Migrate litigation events

jesskelsall:
# What's changed

- One event per document is created when transforming litigation cases.
- Fixed an issue where multiple document events had matching fake import IDs (events with matching IDs are de-duplicated in the event table functions).
- Document events are duplicated over into the family events array as they're also needed here.
- Lots of testing to ensure this works on a diverse range of litigation cases.

## Why

Satisfies [APP-1928](https://linear.app/climate-policy-radar/issue/APP-1928/migrate-litigation-events)

## AI attribution

Analysis only.

## Screenshots

`Sabin.family.132485.0` showing an identical match to CCC prod:
Screenshot 2026-04-30 at 11 33 51


annaCPR (approved):
lgtm

PR #1288, opened 2026-05-05
Comments & minor cleanup

PatFawbertMills:
# What's changed
- Added some helpful comments to the new shadow seach page
- Cleaned up the code as part of a drive-by removing unused code

## Why
- Initially wanted to remove unused and discontinued experiment code but Claude also flagged unusued code elsewhere and not a big job to remove

## AI
- Yes to help find what is unusued


src/pages/_search/index.tsx
```diff
@@ -263,6 +271,18 @@ const ShadowSearch = ({ theme, themeConfig, features }: TProps) => {
           </div>
         )}
       </FiveColumns>
+      {/* 4. ADVANCED FILTERS */}
```
katybaulch: nitpick: can we not number these comments please, as its another thing to remember to maintain. also, 4 here comes after 5 instead of before?

src/pages/_search/index.tsx
```diff
@@ -33,6 +33,16 @@ const columnLayoutCss = "col-start-1 -col-end-1 cols-5:col-start-3 cols-5:-col-e
 
 type TProps = InferGetServerSidePropsType<typeof getServerSideProps>;
 
+/*
+ * SHADOW SEARCH is currently made up of 6 Core surfaces, each surface is commented in code below
+ * NB: this is not necessaruly the order they appear within this component
```
katybaulch: typo:
```suggestion
 * NB: this is not necessarily the order they appear within this component
```

src/pages/_search/index.tsx
PatFawbertMills: Sure I'll remove the numbers, I was hoping it would make sense as an order of visual hierarchy - as in the order of which a user might experience them - the first 4 being customising search, then the last 2 are about viewing the results

katybaulch (approved):
small lil typo - otherwise nice!! 🔨

PR #1289, opened 2026-05-05
Block late day production deploys

katybaulch:
# What's changed

Block late day production deploys via GitHub Actions.

## Why

It's good practice not to deploy too late in the day before people log off in case anything goes wrong. This is us codifying that in the frontend repo.

NOTE: Manual deploys can still be done via Pulumi and via the '[rollback](https://github.com/climatepolicyradar/navigator-frontend/blob/main/.github/workflows/deploy-frontend.yml)' github action, so if we really need to we're not blocked. I'll write up how to do this the runbook I'm writing on how to deploy/rollback.

## AI attribution

Used to help generate the Makefile commands with extra usage help!

## Screenshots

n/a let's test at 5:30pm!

### DIFF
--- a/.github/workflows/deploy-all-production.yml
+++ b/.github/workflows/deploy-all-production.yml
@@ -8,7 +8,54 @@ permissions:
   contents: read
 
 jobs:
+  precheck:
+    name: Validate window before production deploy
+    runs-on: ubuntu-latest
+    environment:
+      name: production
+    env:
+      SLACK_WEBHOOK: ${{ secrets.SLACK_UPDATES_CHANNEL_WEBHOOK_URL }}
+    outputs:
+      blocked: ${{ steps.time-window.outputs.blocked }}
+      current_hm: ${{ steps.time-window.outputs.current_hm }}
+    steps:
+      - if: ${{ github.ref != 'refs/heads/main' }}
+        run: |
+          echo "This workflow can only be run from main."
+          exit 1
+      - name: Block deploy outside local change window
+        id: time-window
+        run: |
+          current_hm="$(TZ=Europe/London date +%H%M)"
+          echo "current_hm=$current_hm" >> "$GITHUB_OUTPUT"
+          if [[ "$current_hm" -ge 1730 || "$current_hm" -lt 0900 ]]; then
+            echo "blocked=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "blocked=false" >> "$GITHUB_OUTPUT"
+          fi
+      - name: Notify Slack about blocked deploy
+        if: steps.time-window.outputs.blocked == 'true' && env.SLACK_WEBHOOK != ''
+        run: |
+          COMMIT_URL="https://github.com/${{ github.repository }}/commit/${{ github.sha }}"
+          COMMIT_SHORT=$(echo "${{ github.sha }}" | cut -c1-7)
+          MESSAGE="⛔ Deploy blocked outside local change window (09:00-17:30 Europe/London). Current local time: ${{ steps.time-window.outputs.current_hm }}. Attempted deploy of all production themes from <$COMMIT_URL|$COMMIT_SHORT>. @application_engineers"
+          jq -n --arg msg "$MESSAGE" '{text: $msg, mrkdwn: true}' > slack-payload.json
+      - name: Send Slack message
+        if: steps.time-window.outputs.blocked == 'true' && env.SLACK_WEBHOOK != ''
+        uses: slackapi/slack-github-action@v3.0.2
+        with:
+          webhook: ${{ env.SLACK_WEBHOOK }}
+          webhook-type: incoming-webhook
+          payload-file-path: slack-payload.json
+      - name: Fail blocked deploy
+        if: steps.time-window.outputs.blocked == 'true'
+        run: |
+          echo "Deploy blocked outside local change window (09:00-17:30 Europe/London). Current local time: ${{ steps.time-window.outputs.current_hm }}"
+          exit 1
+
   deploy-production:
+    needs: precheck
+    if: needs.precheck.outputs.blocked != 'true'
     strategy:
       matrix:
         theme: [cpr, cclw, mcf, ccc]
@@ -17,10 +64,6 @@ jobs:
     environment:
       name: production
     steps:
-      - if: ${{ github.ref != 'refs/heads/main' }}
-        run: |
-          echo "This workflow can only be run from main."
-          exit 1
       - uses: actions/checkout@v6
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v6.1.0
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -27,16 +27,47 @@ jobs:
     runs-on: ubuntu-latest
     environment:
       name: production
+    env:
+      SLACK_WEBHOOK: ${{ secrets.SLACK_UPDATES_CHANNEL_WEBHOOK_URL }}
     steps:
       - if: ${{ github.ref != 'refs/heads/main' }}
         run: |
           echo "This workflow can only be run from main."
           exit 1
+      - name: Block deploy outside local change window
+        id: time-window
+        run: |
+          current_hm="$(TZ=Europe/London date +%H%M)"
+          echo "current_hm=$current_hm" >> "$GITHUB_OUTPUT"
+          if [[ "$current_hm" -ge 1730 || "$current_hm" -lt 0900 ]]; then
+            echo "blocked=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "blocked=false" >> "$GITHUB_OUTPUT"
+          fi
+      - name: Notify Slack about blocked deploy
+        if: steps.time-window.outputs.blocked == 'true' && env.SLACK_WEBHOOK != ''
+        run: |
+          COMMIT_URL="https://github.com/${{ github.repository }}/commit/${{ github.sha }}"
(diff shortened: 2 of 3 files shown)


PatFawbertMills:
I saw your comment about manually overriding I thin we should hold off with this block until we enable and document a work around, just in case we immediately run into an emergency or similar and need to deploy.

jamesgorrie (commented):
I don't really work that much on frontend, so little skin in the game, but really feel as though this will sting us in an emergency, even given the override. 

I also feel like this penalises folks and reduces accessibility of the deployment for those who work outside of these hours (I do on occasion due to not being able to work in hours for example).

> Good practise

☝️ I would say it's good practise to for a person to be responsible for shepherding their PR to production, yes, not to work between certain hours. 

What has happened that we want to put this in place?

@PatFawbertMills mentioned overriding - what is the mechanic?

katybaulch:
> I don't really work that much on frontend, so little skin in the game, but really feel as though this will sting us in an emergency, even given the override.
> 
> I also feel like this penalises folks and reduces accessibility of the deployment for those who work outside of these hours (I do on occasion due to not being able to work in hours for example).
> 
> > Good practise
> 
> ☝️ I would say it's good practise to for a person to be responsible for shepherding their PR to production, yes, not to work between certain hours.
> 
> What has happened that we want to put this in place?
> 
> @PatFawbertMills mentioned overriding - what is the mechanic?

We shouldn't be deploying to production outside of core working hours, unless there is an emergency. You can still merge PRs, deploy to staging, have review stacks etc -- you just can't use Github Actions to deploy to production. So no accessibility block there - you can work as normal.

Yes - we as devs are responsible for this (alongside a bunch of other things). But we still build in automated checks to catch us when we miss something because, after all, we're only human!

This is an outcome from the incident last week (or week before?) -- alongside documenting how to rollback/what to do in a frontend incident/how to deploy. The two support each other - we can still do it IF we need to, but the default becomes to block production deploys outside of core working hours.

jamesgorrie:
@katybaulch I really get the sentiment, but I am not sure a codified, blocking process here is most effective, and also blocks work that sometimes is appropriate.

The thoughts I have...

## This adds gates not safety 

Incident responses often lead to more testing (🥳), not fewer deployments (🤔). This feels potentially regressive in terms of CD, it adds friction, not safety. an indicator that we should be addressing the issue that our coverage needs improvement, be it in PRs and post deployment (smoke tests).

Smoke tests would be a more elegant solution in my eyes, and has been suggested a lot by the team. It is mostly setup, and has been or a while. We don't need to go large. We could use the mentioned incident as the first test, and leave a pattern for others to follow (and get us closer to CD 🍾).


## Blocking deploys when we are on fire

Most hotfixes and incidents I have worked on, we deploy after hours as that is when the fix goes in. When speed is of the essence, I can only see this adding fuel to a fire AKA higher [MTTR](https://sre.google/sre-book/testing-reliability/#relationships-between-testing-and-mean-time-to-repair) leading us to leave later.


## We shouldn't be deploying to production outside of core working hours

Where has this come from? In terms of a policy? I don't always work the hours in this PR, so how do we make this inclusive without being basically all hours? This is a far reaching change that affects people's workflows, and I am trying to get to the heart of how it was made?


## "Work as normal" <= normal?

Normal for me, and potentially others, is not within these hours, and dependant on circumstances (see the above on being on alert).

This would block
- working in different timezones
- starting early to leave early
- any other flexibility we might want


## What is the override?

If we still think this is the way we want to respond to the incident, I am still unclear as to what the "escape hatch deployment" is when we need it? 

---

Personally I think we can add more safety and fewer gates. We would still block a deploy - just based on tests, rather than the very indirect time of day.

References
- https://charity.wtf/tag/friday-deploys/
- https://charity.wtf/2021/08/27/software-deploys-and-cognitive-biases/

katybaulch:
> @katybaulch I really get the sentiment, but I am not sure a codified, blocking process here is most effective, and also blocks work that sometimes is appropriate.

To clarify: we are not completely blocking production deploys - I think there might be a misunderstanding.
> 
> The thoughts I have...
> 
> ## This adds gates not safety
> Incident responses often lead to more testing (🥳), not fewer deployments (🤔). This feels potentially regressive in terms of CD, it adds friction, not safety. an indicator that we should be addressing the issue that our coverage needs improvement, be it in PRs and post deployment (smoke tests).
> 
> Smoke tests would be a more elegant solution in my eyes, and has been suggested a lot by the team. It is mostly setup, and has been or a while. We don't need to go large. We could use the mentioned incident as the first test, and leave a pattern for others to follow (and get us closer to CD 🍾).

I fully agree with you - I've been pushing for regression suites since I started, but these have never prioritised in the nearly three years I've been here now - and we have fewer frontend tests now than we did even at the end of last year now. The incident in reference could have been avoided with a regression suite (which we did actually have, but got ripped out when the cull of our e2e tests happened by the looks of things). 

This (blocking deploys) is a hack and feels like the best we can do for now until someone sees the value in us making these smoke test suites over new feature work (which I really hope happens). It's grotty. My ideal right now would be merging this in & removing it once we get a chance to add smoke tests. But again, we could be waiting months for this to happen... I'm not sure any of our incidents have actually led to much more testing than what we were doing before the incident? We always talk about it, and never act.

To be clear - I want fewer gates. But until we have more safety, more of the right tests etc, this feels like a reasonable tradeoff? We can still deploy to prod if needed [see below]. 

> ## Blocking deploys when we are on fire
> Most hotfixes and incidents I have worked on, we deploy after hours as that is when the fix goes in. When speed is of the essence, I can only see this adding fuel to a fire AKA higher [MTTR](https://sre.google/sre-book/testing-reliability/#relationships-between-testing-and-mean-time-to-repair) leading us to leave later.

We can still deploy to production (manually via the new makefile commands, or via the [deploy-frontend.yml action](https://github.com/climatepolicyradar/navigator-frontend/blob/main/.github/workflows/deploy-frontend.yml) if really needed (in these cases you've mentioned - hotfixing/rollbacks/fastforwards). 

> ## We shouldn't be deploying to production outside of core working hours
> Where has this come from? In terms of a policy? I don't always work the hours in this PR, so how do we make this inclusive without being basically all hours? This is a far reaching change that affects people's workflows, and I am trying to get to the heart of how it was made?

I believe the policy has always been don't deploy to production outside core hours unless its needed imminently for a demo, or unless you need to for an emergency ([core hours](https://www.notion.so/climatepolicyradar/Development-Environment-Set-Up-984ffd4470734a87bdec965ac005f364?source=copy_link#8d1e6704100d478da37a4126b47f7bd0) used to be defined in the handbook, I wouldn't know where to find that now) (? at least this is what I was told when I started, and not sure where/if this is written down), and it used to be don't deploy on Thursday afternoons past a point too (in lieu of working Fridays). 

I'm not sure I understand the pushback on this point - we are not blocking auto-deploys to staging, we are not blocking production deploys in their entirety. We are blocking 'push-the-red-button' auto deploy all themes or deploy main of this other theme - _only_ in the frontend repo - outside of core working hours when other people aren't around.

> ## "Work as normal" <= normal?
> Normal for me, and potentially others, is not within these hours, and dependant on circumstances (see the above on being on alert).
> 
> This would block
> 
> * working in different timezones
> * starting early to leave early
> * any other flexibility we might want

Perhaps this could be better rephrased as when is it reasonable to deploy to production outside of core hours, vs when is it less reasonable?

Reasonable (to me): - emergency situations (rollbacks, hotfixes, fastforwards), something needed for a demo/COP that can't be deployed the next working day for whatever reason etc

Less reasonable? :- Someone deploying to production at 7:30pm because they've worked different hours that day or 'just really want to get' what they've been working on into prod maybe? That could in theory wait until the next working day

What other situations do you think its reasonable vs less reasonable to deploy, to production, outside of the core CPR working hours?

> ## What is the override?
> If we still think this is the way we want to respond to the incident, I am still unclear as to what the "escape hatch deployment" is when we need it?

I am writing a runbook on deploys for the frontend, another action of the postmortem, which includes how to use the [deploy-frontend.yml action](https://github.com/climatepolicyradar/navigator-frontend/blob/main/.github/workflows/deploy-frontend.yml) to rollback/fastforward/hotfix in case of emergencies (which doesn't have this temporary block on it). I've also added new makefile commands to allow building & deploying images from a local machine to remote AWS, as a fallback for that, and will be including how to use them in said runbook (part of Patrick's suggestion from the last portmortem).

> 
> Personally I think we can add more safety and fewer gates. We would still block a deploy - just based on tests, rather than the very indirect time of day.
> 

Totally onboard with that solution in the long run. What do you suggest for the interim solution, until that work is refined, prioritised, included in a sprint, and completed?

I'm not opposed to blocking merges instead as a temporarily alternative solution, but I think that as a solution would be better if we had ephemeral environments so people could test with real data in a production-esque environment whilst their PRs couldn't be merged for testing on staging (which is still the process in this repo).

jamesgorrie:
@katybaulch - as I do not work a tonne in this repo - definitely take the steer from others. I think if we think this is a solution - we can just use a people process, and ask people not to do this.

re: This PR - I think it is moving us in the wrong direction, culturally and technically. Culturally is actually why I am being so vocal about this. Which it sounds like we agree on a bit.

Where I disagree is that I cannot see the safety this adds, and, similar to using [security through obscurity](https://en.wikipedia.org/wiki/Security_through_obscurity), I think it gives us a false sense of security.

After this PR, I can still, at any time, not be responsible for seeing my code through to `production`, still break things and not be aware, and still think it is fine as it is between the hours stated above. And if I wanted to deploy "to get it in" I could just use the local command.

---

On a side note of these specific hours - my hours are often 08:00 -> 16:00 as, dependant on circumstance, that is my best working times.

PR #1290, opened 2026-05-06
APP-1975: Apply transformations to collection page

jesskelsall:
# What's changed

- Updated the collection page to do the now standard processes for fetching data, handling one or more errors, and debugging the data for testing purposes.
- `getCollectionData` makes an API request for the collection and then for each family (as the family's documents and their events are needed to populate the page's tables).
- Fixed generative event IDs matching which causes the events table to remove one as it is considered a duplicate.
- `TCollectionPublicWithFamilies` is now just `TCollectionPublic` with one extra property rather than a new type definition.

## Why

Satisfies [APP-1975](https://linear.app/climate-policy-radar/issue/APP-1975/apply-transformations-to-collection-page)

## AI attribution

Analysis only.


src/bff/methods/getCollectionData.ts
```diff
@@ -36,7 +36,33 @@ export const getCollectionData = async (slug: string, features: TFeatures): Prom
...
+      );
+    } catch (error) {
+      errors.push(error as Error);
+    }
+  }
+
   /* Transform API data for presentation */
 
-  return collectionTransformer({ collection }, null, errors);
+  return collectionTransformer({ collection }, { collection: dataInCollection, families: dataInFamilies }, errors);
```
PatFawbertMills, line 67: Appreciation, part 1: I like the neat separation of concerns. Load the data...

src/bff/transformers/collectionTransformer.ts
```diff
@@ -8,11 +9,23 @@ export const collectionTransformer = (
...
   if (collectionApiOldData === null) return { data: null, errors };
 
-  if (collectionApiNewData) {
-    // TODO: introduce transformations for new data model API data
-    return { data: null, errors };
+  if (collectionApiNewData.collection && collectionApiNewData.families) {
+    try {
+      return {
+        data: {
+          collection: transformCollection(collectionApiNewData.collection, collectionApiNewData.families),
```
PatFawbertMills, line 16: Part 2: choose a transformer

src/bff/transformers/partials/transformCollection.ts
```diff
@@ -0,0 +1,18 @@
+import { transformFamily } from "@/bff/transformers/partials/transformFamily";
+import { TDataInDocument, validateCollectionAttributes } from "@/schemas";
+import { TCollectionPublicWithFamilies } from "@/types";
+
+export const transformCollection = (document: TDataInDocument, familyDocuments: TDataInDocument[] = []): TCollectionPublicWithFamilies => {
+  const collectionAttributes = validateCollectionAttributes(document.attributes);
+  const families = familyDocuments.map(transformFamily);
+
+  return {
+    description: document.description || families[0]?.summary || "",
```
PatFawbertMills, line 10: Part III: return the transformed model ✏️ neat.

PatFawbertMills (approved):
Appreciate the architecture of the approach taken.

PatFawbertMills:
Looks like we've tipped over the 650kb size limit. Suggest we up to 750kb and aim to circle back and do some optimisations after we have fully migrated to the new model.

PR #1291, opened 2026-05-06
feat: transform author type and author

odrakes-cpr:
# What's changed

Slight updates to the transformer to map the author and author metadata for international agreements. 

- author type comes back as a nested value of author, so mapping accordingly

## Why

Required for full transformation of the related corpora.

## AI attribution

NA lol

## Screenshots

Screenshot 2026-05-06 at 16 57 18
Screenshot 2026-05-06 at 16 57 32


jesskelsall (approved):
Not sure why I'm unable to leave a comment on a line of code, but: use `firstCase` instead of your own capitalize function (we have a util already that's very similar to your method).

Once that's sorted, this is good to go! Approved in advance.

PR #1294, opened 2026-05-07
Slim down merge to main

PatFawbertMills:
# What's changed
- Removed repeated jobs from the pull_request steps
- Remove comments

## Why
- When we need to bypass PR rules (see recent incident) we are blocked on the same e2e tests in the merge to main step (hi prio)
- Arguably unnecessary to run all steps again when they are under the same conditions
- Speed (low prio)

## Consensus?
- Yes consensus from the application team


odrakes-cpr (approved):

katybaulch (approved):
thanks! 🧹

PR #1295, opened 2026-05-07
Handle config GET failures resiliently

jesskelsall:
# What's changed

- When requests to `/config` fail, an empty config data structure is returned instead.
- Any error making this request are logged.

## Why

This means that if there is a bug or other site-wide issue with the config endpoint, pages will load with some missing data rather than fail to load entirely.

## AI attribution

None used.

### DIFF
--- a/src/api/http-common.ts
+++ b/src/api/http-common.ts
@@ -1,7 +1,7 @@
 import axios from "axios";
 import type { AxiosInstance, AxiosResponse } from "axios";
 
-import { TApiCorpusTypeDictionary, TApiDataNode, TApiGeography, TApiLanguages } from "@/types";
+import { TApiConfig } from "@/types";
 
 export async function getEnvFromServer() {
   return await axios.get("/api/env").then((res: any) => res);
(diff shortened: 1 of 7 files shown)


src/pages/geographies/[id].tsx
```diff
@@ -69,7 +70,9 @@ export const getServerSideProps = (async (context) => {
       const parentGeographyV2Data = await apiClient.get<TApiItemResponse<GeographyV2>>(`/geographies/${geographyV2.subconcept_of[0].slug}`);
       parentGeographyV2 = parentGeographyV2Data.data.data;
     }
-  } catch {}
+  } catch {
+    // Do nothing
```
Comment (jesskelsall) [inline src/pages/geographies/[id].tsx:74]: Added to keep the linter happy but this API call will be getting removed shortly anyway.

src/api/http-common.ts
```diff
@@ -72,12 +72,20 @@ class ApiClient {
...
-      corpus_types: TApiCorpusTypeDictionary;
-      languages: TApiLanguages;
-    }>("/config");
+  async getConfig(): Promise<{ config: TApiConfig; error: Error | null }> {
+    try {
+      const config = await this.get<TApiConfig>("/config");
+      return { config: config.data, error: null };
+    } catch (error) {
+      return {
+        config: {
```
PatFawbertMills, line 81: ✏️ neat idea

PatFawbertMills (approved):
lgtm

PR #1305, opened 2026-05-12
APP-2007: [CCC] Display summaries for docs without pdfs

jesskelsall:
# What's changed

- Populating an events table with "documents only" keeps family events that look like document events but have no corresponding document.
- Event table document cells will render the action taken and summary without a document to link to.
- "View more" buttons have been reverted to actually expanding the summary rather than passing the click event through to the table row to open the drawer. 
- Table rows without a document still aren't clickable to open the drawer with more info.

## Why

Satisfies [APP-2007](https://linear.app/climate-policy-radar/issue/APP-2007/ccc-display-summaries-for-docs-without-pdfs)

## AI attribution

None

## Screenshots

Screenshot 2026-05-12 at 14 40 44

### DIFF
--- a/src/utils/eventTable.tsx
+++ b/src/utils/eventTable.tsx
@@ -246,7 +246,7 @@ export const getEventTableRows = ({
 
   rowsData.forEach(({ family, event, document }) => {
     // Some document events don't have corresponding documents, so keep those even when documentEventsOnly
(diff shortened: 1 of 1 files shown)


odrakes-cpr (approved):

annaCPR (approved):
nicely done

PR #1308, opened 2026-05-13
Feature/app 2069 bug missing litigation documents

annaCPR:
# What's changed

A new document status value has been added in the new `data-in-api` document response, specifically for litigation documents that were transformed from events due to document files not being available.

## Why

FE filters out any documents that do not have status of `published`. Hence a new status to explicitly describe the state of litigation documents that have not been `published` but should still be displayed as there is relevant information for the user.

## AI attribution
None


jesskelsall (approved):
A nice simple good solution, approved!

PR #1311, opened 2026-05-14
APP-2010: Hide family documents that are only created

jesskelsall:
# What's changed

Update the way that documents are displayed by only showing published documents and hiding created ones - these are essentially drafts.

## Why

Part of feedback on [APP-2010](https://linear.app/climate-policy-radar/issue/APP-2010/documents-dont-delete-properly-from-admin-service-cclw)

## AI attribution

Analysis of other places family/document filtering might also be occurring.


PatFawbertMills (approved):
lgtm

PR #1318, opened 2026-05-20
fix: use standard author type

jamesgorrie:
# What's changed

- we added a regular [`author` and `author_type`](https://github.com/climatepolicyradar/navigator-backend/pull/1230) so we can remove this piece of custom code that was reading from the irregular `author` `Label`

## Why

- simplicity

## Screenshots
Screenshot 2026-05-20 at 20 22 35


jamesgorrie:
Merging as I don't want this to become a bug.

PR #1319, opened 2026-05-21
fix: add waf-acl-detection header

jamesgorrie:
# What's changed

- adds the `waf-acl-detection` header which can take the value `chrome125-user-agent` (or others later)

## Why

- so @juliesaigusaCPR can use this in posthog to do some analysis of the bot data

--

part of https://linear.app/climate-policy-radar/issue/APP-2109/tag-suspected-bots-so-they-can-be-filtered-downstream


jamesgorrie:
Merging as this is deployed.

PR #1324, opened 2026-05-26
APP-2053 :: search result document card

PatFawbertMills:
# What's changed
- New component for `DocumentCard` which ideally can be used in any place we display a document.
- We will need to expand this component in future for non-principal documents, for now the focus has been on displaying it in the search results

## Why
- Start to componentise and move away from the experimental work

## AI attribution
- Generated the storybook component for me
- General advice/autocomplete
- Generate initial component test

## Screenshots
Screenshot 2026-05-26 at 11 33 41


src/components/molecules/documentCard/DocumentCard.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+  title: "Molecules/DocumentCard",
+  component: DocumentCard,
+  parameters: { layout: "centered" },
+  args: { onClick: () => {} },
+} satisfies Meta<typeof DocumentCard>;
+type TStory = StoryObj<typeof DocumentCard>;
+
+export default meta;
+
+const baseDocument: SearchDocument = {
```
jesskelsall, line 17: Praise: Nice repeatable pattern.

src/components/molecules/documentCard/DocumentCard.stories.tsx
```diff
@@ -0,0 +1,92 @@
+import { Meta, StoryObj } from "@storybook/nextjs-vite";
+
+import { SearchDocument } from "@/api/search";
+
+import { DocumentCard } from "./DocumentCard";
+
+const meta = {
+  title: "Molecules/DocumentCard",
```
jesskelsall, line 8: Musing: This is technically an atom (it uses no other component in our library) but it feels more molecule-ish so I get it. Definitely some quibbling to do once we componentise more things.

src/components/molecules/documentCard/DocumentCard.tsx
```diff
@@ -0,0 +1,81 @@
...
+      data-ph-capture-attribute-link-purpose={context ?? "document-card"}
+      data-ph-capture-attribute-position-page={page}
+      data-ph-capture-attribute-position-total={page !== undefined && positionOffset !== undefined ? positionOffset + page : undefined}
+    >
+      <span className="basis-12.5 grow-0 shrink-0">
+        <span>{getDocumentPublishedYear(document)}</span>
+      </span>
+      <span>
+        {/* CORE DETAILS */}
+        <h3 className="font-semibold text-base mb-3">
```
jesskelsall, line 43: Question: Why H3? Are we expecting to have a H1 and H2 on the page?

src/components/molecules/documentCard/DocumentCard.tsx
```diff
@@ -0,0 +1,81 @@
...
+          {linkHref(document) ? (
```
jesskelsall: Concern: `linkHref` returns a URL string but is only being used in this component here for branching logic. Is this missing the actual link markup?

src/components/molecules/documentCard/DocumentCard.tsx
```diff
@@ -0,0 +1,81 @@
...
+            }}
+          />
+        )}
+        <span className="flex gap-4">
+          <span>{document.labels.find((label) => label.type === "category")?.value.value}</span>
+          {/* DISPLAYING GEOS */}
+          <ul className="flex flex-wrap gap-2 text-base text-neutral-600">
+            {document.labels
+              .filter((label) => label.type === "geography")
+              .filter((label) => label.value.type === "country")
```
jesskelsall: Optimisation: Filter once with an `&&` to prevent iterating twice.

src/components/molecules/documentCard/DocumentCard.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+  });
+
+  it("sets analytics data attributes on the button", () => {
+    render(<DocumentCard document={baseDocument} onClick={() => {}} analytics={{ context: "search-results", page: 2, positionOffset: 10 }} />);
+    const button = screen.getByRole("button");
+    expect(button).toHaveAttribute("data-ph-capture-attribute-link-purpose", "search-results");
+    expect(button).toHaveAttribute("data-ph-capture-attribute-position-page", "2");
+    expect(button).toHaveAttribute("data-ph-capture-attribute-position-total", "12");
+  });
+});
```
jesskelsall, line 98: Praise: A good set of tests!

src/components/molecules/documentCard/DocumentCard.tsx
```diff
@@ -0,0 +1,81 @@
...
+function linkHref(doc: SearchDocument): string | undefined {
+  if (doc.attributes.deprecated_slug)
+    if (doc.labels.find((label) => label.value.value === "Principal")) {
+      return `/document/${doc.attributes.deprecated_slug}`;
+    } else {
+      return `/documents/${doc.attributes.deprecated_slug}`;
+    }
+}
+
+function getDocumentPublishedYear(doc: SearchDocument) {
```
jesskelsall: Question: Did I miss the memo about using `function` over `const () =>`?

jesskelsall (approved):
Looks pretty solid! A variety of comments for you but nothing I'd consider blocking, so approved.

src/components/molecules/documentCard/DocumentCard.stories.tsx
PatFawbertMills: My thinking was at some point we might atomise some of the things in this molecule, a bit like what we did with the country links / flags component.

src/components/molecules/documentCard/DocumentCard.tsx
PatFawbertMills: I think carried over from the old `FamilyListItem` component. For reference Google uses h3 for the search result items.

I think generally an h3 feels semantically safe in this instance?

src/components/molecules/documentCard/DocumentCard.tsx
PatFawbertMills: Great catch, didn't evolve this when we moved to the button -> trigger -> open drawer approach

src/components/molecules/documentCard/DocumentCard.tsx
PatFawbertMills: Good question and good for us to be strict about. Let's discuss.

PR #1325, opened 2026-05-26
Add node to ignore for dependabot upgrades

PatFawbertMills:
# What's changed
- Adding an ignore section to the dependabot yaml
- Add Node to this list as we should do this manually

## Why
- To ignore certain packages from upgrading

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,9 +4,11 @@ updates:
     directory: /
     schedule:
       interval: weekly
+    # Manually decide when to update certain packages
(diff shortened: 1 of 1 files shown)


katybaulch (approved):
lgtm - tho unsure if this syntax is correct

PatFawbertMills:
Looks OK from the spec: https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#ignore--

I realised I had the wrong package name in there though

PR #1327, opened 2026-05-26
Add contextual info for Documents

PatFawbertMills:
# What's changed
- Adding contextual info to the `DocumentCard`

## Why
- Requirements

### DIFF
--- a/src/components/molecules/documentCard/DocumentCard.test.tsx
+++ b/src/components/molecules/documentCard/DocumentCard.test.tsx
@@ -64,8 +64,8 @@ describe("DocumentCard", () => {
       ],
     };
     render(<DocumentCard document={doc} onClick={() => {}} />);
-    expect(screen.getByText("France")).toBeInTheDocument();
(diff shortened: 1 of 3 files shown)


src/components/molecules/documentCard/DocumentCard.test.tsx
```diff
@@ -64,8 +64,8 @@ describe("DocumentCard", () => {
       ],
     };
     render(<DocumentCard document={doc} onClick={() => {}} />);
-    expect(screen.getByText("France")).toBeInTheDocument();
-    expect(screen.getByText("Germany")).toBeInTheDocument();
+    expect(screen.getByText("France", { exact: false })).toBeInTheDocument();
```
PatFawbertMills, line 67: The comma in the string was failing this test, I think we are fair to assume as long as the country name appears in some capacity this test should pass.

jesskelsall (approved):
Approved - looks good enough for now.

PR #1328, opened 2026-05-28
APP-1963 remove aggregations

PatFawbertMills:
# What's changed
- Remove the aggregations from the shadow search page

## Why
1. Decision against blocking user behaviour
2. Gives us a cleaner slate to start layering up the new requirements around filters, rules, typeahead, etc


katybaulch (approved):
fab tidies 🧹

PR #1329, opened 2026-06-01
APP-1912: ICCN landing page

jesskelsall:
# What's changed

- Added `LandingPage` component with a configuration spec for defining all of the page's details (essentially a templated landing page).
- Added an ICCN landing page, images, and route config.

## Why

Satisfies [APP-1912](https://linear.app/climate-policy-radar/issue/APP-1912/iccn-handshake-page).

## AI attribution

Analysis only.

## Screenshots

Screenshot 2026-06-01 at 11 44 49


src/components/organisms/LandingPage.tsx
```diff
@@ -0,0 +1,114 @@
...
+    title: string;
+    content: ReactNode;
+  }[];
+};
+
+type TProps = {
+  config: TLandingPageConfig;
+};
+
+export const LandingPage = ({ config }: TProps) => (
```
PatFawbertMills: Question: do we need a section and type changes to manage the HTML meta data? E.g. a page description?

PatFawbertMills: Happy if the answer is: when we have more detail.

themes/cpr/routes.json
```diff
@@ -18,5 +18,10 @@
     "title": "Rio submissions",
     "path": "rio-submissions",
     "contentPath": "rio-submissions"
+  },
+  {
+    "title": "ICCN",
```
PatFawbertMills: Checking: have we not been given anything more than this so far? I would presume we wouldn't only use the acronym here.

PatFawbertMills (approved):
Liking the configuration approach.

Couple of non-blocking comments on HTML metadata and name/url of new page.

themes/cpr/routes.json
jesskelsall: Great spot, this should match the title on the page - fixed.

src/components/organisms/LandingPage.tsx
jesskelsall: This was actually an accidental omission on my part - I had intended to wrap with `Layout` and apply the config's hero title and description to that, which I've now done. Thanks for spotting this!

PR #1332, opened 2026-06-02
fix: temporary fix, formatting issue for linebreaks and breakpoints

odrakes-cpr:
# What's changed

Fix summary formatting so line breaks are rendered correctly in case summaries.

Previously, we ignored literal \r\n / \n sequences in the summary data. This change updates the rendering to preserve real line breaks, ensuring multi-paragraph summaries display as intended.

Impact

Restores paragraph and line-break formatting in summary text.
Improves readability of long case descriptions.
Prevents summaries from appearing as a single block of text.

## Why

This was a request from the sabin team, we would normally handle this on the backend but the platform do not have capacity to deal with the data changes downstream this would cause if fixed in the backend. 

## Screenshots
Screenshot 2026-06-02 at 11 54 04
Screenshot 2026-06-02 at 11 54 28

### DIFF
--- a/src/components/documents/DocumentHead.tsx
+++ b/src/components/documents/DocumentHead.tsx
@@ -74,7 +74,7 @@ export const DocumentHead = ({ document, family, features, handleViewOtherDocsCl
             <Heading level={1}>{document.title}</Heading>
             <DocumentMetaRenderer family={family} isMain={isMain} document={document} />
 
-            <div className="text-content" dangerouslySetInnerHTML={{ __html: summary }} />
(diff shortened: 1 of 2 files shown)


PatFawbertMills:
We may also want to do the same on the document page where we display the summary again, please see here: https://www.climatecasechart.com/documents/opinion-of-the-national-human-rights-commission-on-the-constitutional-complaints-on-constitutionality-of-carbon-neutrality-act-opinion_56e1?q=opinion+of+the+national+human+rights

odrakes-cpr:
updated doc page


Screenshot 2026-06-02 at 12 44 00

PatFawbertMills (approved):
Nice 1 m8

PR #1334, opened 2026-06-03
Lazy load posthog

PatFawbertMills:
# What's changed
- Lazy load posthog to reduce initial page size

## Why
- Unnecessary to bundle Posthog when we can defer it to the client

## AI attribution
- Some help figuring how to split the Posthog init from the Provider wrapper


jamesgorrie (approved):
Just for my clarity - this does not actually decrease the amount a user will download, but rather speeds up the download of the initial JS by deffering the PostHog load till after?

src/utils/features.ts
```diff
@@ -40,15 +38,3 @@ export const getFeatures = (themeConfig: TThemeConfig, featureFlags: TFeatureFla
 
   return features;
 };
-
-// Add A/B test feature flags into TFeatures. Does not work if called in getServerSideProps
-export const getFeaturesWithABTests = (features: TFeatures): TFeatures => {
```
jesskelsall, line 45: Just a heads up that removing this is effectively removing our A/B test implementation.

jesskelsall (changes requested):
I'm requesting changes as what you have here removes our ability to run A/B tests with no alternative.

PatFawbertMills:
> Just for my clarity - this does not actually decrease the amount a user will download, but rather speeds up the download of the initial JS by deffering the PostHog load till after?

Yes exactly. I was also confused why the bundle size was not lower but according to Claude:

> The size-limit config checks all chunks — including page-specific ones for /search and /documents/[id], which still import useSearch.ts → posthog-js/react. That’s expected and fine; those are lazy-loaded when navigating to those pages.

So the initial page load _should_ be faster as it doesn't need to download posthog, but then the instant the page loads in the browser the deferred load will kick off and load posthog.

PatFawbertMills:
> I'm requesting changes as what you have here removes our ability to run A/B tests with no alternative.

Thanks for flagging. Is it acceptable to move it into its own utility, just for AB checks? That way we can lazy load it as and when we need it?

The issue here was we were importing from `posthog-js` every time we importing a function from `features.ts`

jesskelsall (approved):
I'm happy with `featuresAB.ts`, approved!

src/utils/features.ts
PatFawbertMills: Resolved with move to separate file.

PR #1336, opened 2026-06-03
APP-2147: UI: Checkbox

jesskelsall:
# What's changed

- Evolved `src/components/checkbox` (shadow search) into `src/components/atoms/checkbox`.
- Handled all 3 states and disabled state visually.
- Keyboard controls and ARIA labels.
- Storybook stories and unit tests.

## Why

Satisfies [APP-2147](https://linear.app/climate-policy-radar/issue/APP-2147/ui-checkbox).

## AI attribution

Claude used to analyse accessibility considerations & make changes, and write unit tests.

## Screenshots

Screenshot 2026-06-03 at 11 25 42


jamesgorrie (commented):
Looks good - but your supporting test is failing

src/styles/theme.css
```diff
@@ -1,4 +1,10 @@
 @theme {
+  /* CPR Design System variables (June 2026) */
+
+  --color-border-normal: #101a2429;
+  --color-border-light: #101a241f;
+  --color-border-checkbox: #101a2452;
```
PatFawbertMills, line 6: Thanks for starting on the new vars - minor thing though: not sure I like that our checkbox has its own var

Defo not blocking but if we need to support a separate border colour I'd prefer us to get design to give us one for all inputs

src/components/atoms/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const handleChange = (value: TCheckboxState) => {
+    if (rootProps.disabled) return;
+    rootProps.onCheckedChange?.(value);
+  };
+
+  return (
+    <label className={labelClasses}>
+      <BaseCheckbox.Root className={rootClasses} {...rootProps} aria-label={label} onCheckedChange={handleChange}>
+        <BaseCheckbox.Indicator className="flex items-center justify-center text-white">
+          {rootProps.indeterminate ? <HorizontalRuleIcon aria-hidden={true} /> : <Check size={16} />}
```
PatFawbertMills: Consideration: rather than maintain our own components, we could optionally use the Lucide `<Minus>` icon for the indeterminate. It might mean we have fewer alignment issues as they should always remain the same scale rules.

src/components/atoms/checkbox/Checkbox.test.tsx
PatFawbertMills, line 1: Request: can we include a test that checks the label is displayed the same as what is passed in?

src/styles/theme.css
jesskelsall: Good point, I'll raise this with Conor as part of the work I do this afternoon to get all the new variables in and the old ones out.

src/components/atoms/checkbox/Checkbox.tsx
jesskelsall: Oh that's what it's called! I'd tried searching the library for a few different keywords and couldn't find it. Will definitely replace!

src/components/atoms/checkbox/Checkbox.test.tsx
jesskelsall: Absolutely.

PatFawbertMills (approved):
🚢 lovely and production ready

PR #1340, opened 2026-06-04
APP-2158: Set design system variables in Tailwind

jesskelsall:
# What's changed

This is a very noisy PR that [finally] removes two generations of old Tailwind variables!  
In particular check out `theme.css`.

- Added all CPR-DS variables in the same format and with almost-identical names to Tailwind, categorised by the same headings. Many of these reference the default Tailwind colour palette.
- Removed old Tailwind variables following a set of principles:
  - App or third party brand specific colours are untouched.
  - Where possible and obvious, the newer equivalent variable is used. This can result in a small amount of difference versus prod but starts to bring our overall design up to standard.
  - All other variables are replaced with Tailwind escapes e.g. `bg-[#bada55]`. These should look exceptional at-a-glance and signal to developers that components/pages are "off brand", providing a drive-by opportunity to bring them in line.

## Why

- Satisfies [APP-2158](https://linear.app/climate-policy-radar/issue/APP-2158/set-design-system-variables-in-tailwind).

## AI attribution

Claude has been essential to analysing Tailwind variable use and mass updating values. This has been backed up by my manual side-by-side comparisons with prod on all 4 custom apps.

### DIFF
--- a/src/components/FaqSection.tsx
+++ b/src/components/FaqSection.tsx
@@ -61,7 +61,7 @@ export const FaqSection = ({ title, faqs, accordionMaxHeight = "464px", sectionI
           </Fragment>
         ))}
         {showMore && (
-          <LinkWithQuery href="/faq" className="!text-text-brand-darker font-semibold text-sm !no-underline">
+          <LinkWithQuery href="/faq" className="!text-[#002ca3] font-semibold text-sm !no-underline">
             See more FAQs →
           </LinkWithQuery>
         )}
--- a/src/styles/theme.css
+++ b/src/styles/theme.css
@@ -1,39 +1,52 @@
 @theme {
   /* CPR Design System variables (June 2026) */
 
-  --color-border-normal: #101a2429;
-  --color-border-light: #101a241f;
-  --color-border-checkbox: #101a2452;
+  /* Color / ID Core */
+  --color-white: #FFFFFF;
+  --color-paper: #FBFBF9;
+  --color-recycled-paper: #E5E3DC;
+  --color-inky-blue: #1A4F8C;
+  --color-inky-navy: #133051;
+  --color-inky-black: #101A24;
 
-  /* NEW SEARCH */
+  /* Color / ID Extended */
+  --color-light-blue: #C6DDE8;
+  --color-mustard: #C8B500;
+  --color-green: #A0AF54;
+  --color-forest: #164F4A;
+  --color-cardboard: #B7AE8E;
 
-  --color-inky-blue: #1a4f8c;
-  --color-inky-black: #101a24;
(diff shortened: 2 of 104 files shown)


src/styles/theme.css
```diff
@@ -151,25 +78,22 @@
...
-  /* MISC */
-
   --shadow-lg: 0px 4px 48px 0px rgba(0, 0, 0, 0.08);
   --shadow-md: 0px 4px 16px 0px rgba(0, 0, 0, 0.08);
   --shadow-xs: 0px 2px 2px 0px rgba(0, 0, 0, 0.08);
   --shadow-oep: 0px 8px 44px 0px rgba(32, 32, 32, 0.16);
 
+  --tracking-slight: 0.015em;
+
   --container-1/2-gap-4: calc(50% - (1 / 2 * 1rem));
```
PatFawbertMills, line 94: Out of interest where do we use these?

themes/ccc/components/Feature.tsx
```diff
@@ -18,7 +18,7 @@ export const Feature = ({ heading, subheading, image, imageAlt, children }: IPro
     <Card
       heading={heading || ""}
       extraClasses="h-full w-full"
-      headingClasses="text-text-brand-darker text-2xl font-bold"
+      headingClasses="text-text-brand text-2xl font-bold"
```
PatFawbertMills, line 21: Flagging in relation to another change.

src/components/FaqSection.tsx
```diff
@@ -61,7 +61,7 @@ export const FaqSection = ({ title, faqs, accordionMaxHeight = "464px", sectionI
           </Fragment>
         ))}
         {showMore && (
-          <LinkWithQuery href="/faq" className="!text-text-brand-darker font-semibold text-sm !no-underline">
+          <LinkWithQuery href="/faq" className="!text-[#002ca3] font-semibold text-sm !no-underline">
```
PatFawbertMills, line 64: Question: Later in this PR we are changing `text-brand-darker` to `text-brand` but here we are using the colour hex - is there a reason for this?

PatFawbertMills: See comment on `Feature.tsx`

jesskelsall: This is because the old and new brand colours are equivalent (they're not the same hex but they look very similar), whereas the new design system doesn't have a darker brand colour, so it can't be mapped onto a new colour in the same way.

src/styles/theme.css
jesskelsall: There are some CCLW and MCF components that use this.

PatFawbertMills (approved):
amazing clean up job 🧹

PR #1342, opened 2026-06-08
APP 2145 :: quick search component

PatFawbertMills:
# What's changed
- Utilise the existing `Input` component for the QuickSearch, add corresponding unit tests and update Storybook
- Drive-by fix bug with the `Select` component and componentise this also

## Why
- Standardising use of components of core interactive elements

## AI attribution
- Writing a v1 of unit tests
- Helping diagnose issue with `Select` box - using State to store the container, not `useRefs`

### DIFF
--- a/src/components/atoms/input/Input.test.tsx
+++ b/src/components/atoms/input/Input.test.tsx
@@ -1,0 +1,64 @@
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { vi } from "vitest";
+
+import { Input } from "./Input";
+
+describe("Input", () => {
+  it("renders an input element", () => {
+    render(<Input />);
+    expect(screen.getByRole("textbox")).toBeInTheDocument();
+  });
+
+  it("passes props through to the input element", () => {
+    render(<Input placeholder="Search..." disabled />);
+    const input = screen.getByRole("textbox");
+    expect(input).toHaveAttribute("placeholder", "Search...");
+    expect(input).toBeDisabled();
+  });
+
+  it("does not render a clear button when clearable is false", () => {
+    render(<Input clearable={false} value="some text" onChange={vi.fn()} />);
+    expect(screen.queryByRole("button")).not.toBeInTheDocument();
+  });
+
+  it("hides the clear button when clearable is true but value is empty", () => {
+    render(<Input clearable value="" onChange={vi.fn()} />);
+    expect(screen.getByRole("button")).toHaveClass("hidden");
+  });
+
+  it("shows the clear button when clearable is true and value is set", () => {
+    render(<Input clearable value="some text" onChange={vi.fn()} />);
+    expect(screen.getByRole("button")).not.toHaveClass("hidden");
+  });
+
+  it("calls onClear when the clear button is clicked", async () => {
+    const onClear = vi.fn();
+    render(<Input clearable value="some text" onChange={vi.fn()} onClear={onClear} />);
+    await userEvent.click(screen.getByRole("button"));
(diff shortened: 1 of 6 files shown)


src/components/atoms/input/Input.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+    await userEvent.click(screen.getByRole("button"));
+    expect(onClear).toHaveBeenCalledTimes(1);
+  });
+
+  it("renders icon after the input by default", () => {
+    const icon = <span data-cy="icon">icon</span>;
+    render(<Input icon={icon} />);
+    const input = screen.getByRole("textbox");
+    const renderedIcon = screen.getByTestId("icon");
+    expect(input.compareDocumentPosition(renderedIcon)).toBe(Node.DOCUMENT_POSITION_FOLLOWING);
```
PatFawbertMills, line 47: Shout out Claude for this little tip

src/components/organisms/ConceptPicker.tsx
```diff
@@ -101,6 +101,7 @@ export const ConceptPicker = ({ containerClasses = "", startingSort = "Grouped",
 
   const [search, setSearch] = useState("");
   const [sort, setSort] = useState<TSort>(startingSort);
+  const [container, setContainer] = useState<HTMLDivElement | null>(null);
```
PatFawbertMills, line 104: Claude suggested I do this, as the ref was always resulting in `null` which is why the select box was not appearing on prod.

PatFawbertMills: Full comment:

The problem is that ref.current is being accessed at render time (line 158), but refs are only populated after the component has mounted to the DOM. During the initial render, ref.current is still null.

React 19 made accessing refs during render an explicit error — previously this would silently return null; now it throws.

The ref is attached to the outer <div> at line 118:


<div ... ref={ref}>
But then on line 158, you read ref.current while that same render is still executing — the <div> hasn't been committed to the DOM yet, so the ref isn't set.

Fix: The standard solution is to use a state variable that gets set via a [callback ref](https://react.dev/learn/manipulating-the-dom-with-refs#how-to-manage-a-list-of-refs-with-a-ref-callback) (or useEffect), so the value is only passed to container after mount:


const [container, setContainer] = useState<HTMLDivElement | null>(null);

// ...

<div ... ref={setContainer}>
  {/* ... */}
  <Select container={container} ... />
With a callback ref (ref={setContainer}), React calls setContainer with the DOM node after mount, triggering a re-render with the real element. The initial render passes null to container, and the second render (post-mount) passes the actual node — no render-time ref access.

src/components/atoms/select/Select.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  it("does not render options when the options list is empty", async () => {
+    render(<Select options={[]} />);
+    await userEvent.click(screen.getByRole("combobox"));
+    expect(screen.queryAllByRole("option")).toHaveLength(0);
+  });
+
+  /*
+    TODO: add test for the onChange callback when an option is selected
```
katybaulch, line 45: have we got a ticket or timeline associated with this TODO? just conscious that we might end up missing it!

PatFawbertMills: I'll stick one in triage.

PatFawbertMills: https://linear.app/climate-policy-radar/issue/APP-2165/select-component-research-and-add-unit-test-for-onchange-handler

katybaulch (approved):
lgtm!!

PR #1343, opened 2026-06-08
Fix/app 2095 fix non default date filter bug not triggering search

katybaulch:
# What's changed

- Fix bug where if the user applies a non-default date filter with no other filters, a search wasn't being triggered
- Driveby update CLAUDE instructions with [Karpathy](https://github.com/multica-ai/andrej-karpathy-skills#) guidelines

## Why

Applying a date filter with no other filters wasn't triggering a search because the old filtersDoesNotContainEmptyRule function returned false whenever the default empty label placeholder was present, making the whole group appear empty.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,5 +1,8 @@
 # Project Overview
 
+For dev setup, commands, testing, theming, and deployment see
+[README.md](README.md).
+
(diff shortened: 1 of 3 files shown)


PatFawbertMills (approved):
Nice catch I didn't spot this.

PR #1345, opened 2026-06-09
Feature/app 2048 topic label in applied filters should include topic name

katybaulch:
# What's changed

- Changed the UI to show the concept display name instead of the Wikibase ID, where an available filter exists
- Fall back to showing the Wikibase ID instead of the display name where it isn't possible to get the display name
- Add regression tests

## Why

We were seeing the Wikibase ID instead of the concept name in the frontend UI which is confusing for users

### DIFF
--- a/src/components/_experiment/appliedLabels/AppliedLabels.tsx
+++ b/src/components/_experiment/appliedLabels/AppliedLabels.tsx
@@ -9,6 +9,11 @@ function getTypeOfLabel(label: string, availableFilters: TLabelResult[]): string
   return found ? found.type : null;
 }
 
+function getLabelDisplayName(label: string, availableFilters: TLabelResult[]): string {
+  const found = availableFilters.find((f) => f.id === label);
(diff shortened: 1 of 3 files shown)


src/components/_experiment/appliedLabels/AppliedLabels.tsx
```diff
@@ -19,14 +24,14 @@ function isFilterComplex(filters: TQueryGroup | null | undefined): boolean {
   return filters.filters.some((f) => "filters" in f && isFilterComplex(f));
 }
 
-function AppliedLabel({ label, type, onSelect, onRemove }: { label: string; type?: string; onSelect: () => void; onRemove: () => void }) {
+function AppliedLabel({ displayName, type, onSelect, onRemove }: { displayName: string; type?: string; onSelect: () => void; onRemove: () => void }) {
```
jamesgorrie, line 27: nit: could call it `labelValue` to be more consistent with the shape of the data

jamesgorrie (approved):

PR #1347, opened 2026-06-11
APP 2141 - drawer component

PatFawbertMills:
# What's changed
- New generic <Drawer> component built base upon our new Component Spec definition
- Replacing all other instances of Drawers, or Slideouts

## Why
- Uses base's new Drawer component with custom styling over the top.

## AI attribution
- Lots of CSS problem solving around the direction styles

## Screenshots
- Search result passage like-for-like (old on right)
Screenshot 2026-06-15 at 16 36 00


src/components/atoms/drawer/Drawer.module.css
```diff
@@ -0,0 +1,152 @@
+@reference "../../../styles/main.css";
```
jesskelsall, line 1: Praise: nice use of a CSS module for a large and complex amount of styling. I expect we'll continue using inline Tailwind by default?

src/components/atoms/drawer/Drawer.tsx
```diff
@@ -1,33 +1,49 @@
...
 import { joinTailwindClasses } from "@/utils/tailwind";
 
-type TProps = VaulDrawerRootProps & {
-  children: ReactNode;
-  direction?: "left" | "right";
-  title: ReactNode;
-  trigger?: ReactNode;
+import styles from "./Drawer.module.css";
+
+type TDirection = "left" | "right" | "top" | "bottom";
```
jesskelsall, line 8: Musing: later we may want a different direction on mobile or desktop.

src/components/atoms/drawer/Drawer.tsx
```diff
@@ -1,33 +1,49 @@
-import { ReactNode } from "react";
-import { Drawer as VaulDrawer, DialogProps as VaulDrawerRootProps } from "vaul";
+import { Drawer as BaseDrawer, DrawerRootProps } from "@base-ui/react/drawer";
+import { LucideX } from "lucide-react";
```
jesskelsall, line 2: Query: Is the a difference between `lucide-react`'s `X` and `LucideX` components? We now have a mix in use.

src/styles/main.css
```diff
@@ -39,3 +39,9 @@
     visibility: hidden;
   }
 }
+
+.highlights {
+  ::global(hi) {
```
jesskelsall, line 44: Praise: this is a very curious but neat CSS selector!

jesskelsall (approved):
Looks good so happy to approve. A few food-for-thought comments but nothing too notable.

src/components/atoms/drawer/Drawer.tsx
PatFawbertMills: Oh, how weird I don't even see `LucideX` on their list of icons on the website 🤷 not sure tbh but prefixing with Lucide is not necessarily now that we are only using their icons.

src/components/atoms/drawer/Drawer.tsx
PatFawbertMills: I had thought I had done this but lets do as part of a fine tune later as there is no degradation from existing to this.

src/components/atoms/drawer/Drawer.module.css
PatFawbertMills: Yeh this is a special case due to the amount of other controls. Don't think it would be possible (let alone desirable) to do this inline!

PR #1350, opened 2026-06-15
Move country name assert to after the existince checks

PatFawbertMills:
# What's changed
- Due to the API returning a 404 for certain geos (Kosovo) the front-end app is crashing out and returning an internal server error
- I've moved the line that was causing the crash to after we check for the existence of data from the API, so when it has no data it will return a 404 instead.

## Why
- Avoid triggering an "Internal server error" to the user


jesskelsall (approved):
Good shout! Approved.

PR #1352, opened 2026-06-15
APP 2155 :: remove targets block

PatFawbertMills:
# What's changed
- Removed the targets block from the Geography pages and the Family pages

## Why
- Data was legacy manually curated targets from pre-KG era
- We will be using the passages classified with "Target" in future

### DIFF
--- a/src/components/pages/familyPage.tsx
+++ b/src/components/pages/familyPage.tsx
@@ -7,7 +7,6 @@ import { CollectionsBlock } from "@/components/blocks/collectionsBlock/Collectio
 import { DocumentsBlock } from "@/components/blocks/documentsBlock/DocumentsBlock";
 import { MetadataBlock } from "@/components/blocks/metadataBlock/MetadataBlock";
 import { NoteBlock } from "@/components/blocks/noteBlock/NoteBlock";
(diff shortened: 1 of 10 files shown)


katybaulch (approved):
simples 🧹 :meerkat: (whaaaa no meerkat?)

src/components/pages/familyPage.tsx
```diff
@@ -7,7 +7,6 @@ import { CollectionsBlock } from "@/components/blocks/collectionsBlock/Collectio
 import { DocumentsBlock } from "@/components/blocks/documentsBlock/DocumentsBlock";
 import { MetadataBlock } from "@/components/blocks/metadataBlock/MetadataBlock";
 import { NoteBlock } from "@/components/blocks/noteBlock/NoteBlock";
-import { TargetsBlock } from "@/components/blocks/targetsBlock/TargetsBlock";
```
jesskelsall, line 10: Suggestion: Go one step further and delete `TargetsBlock` from the codebase.

jesskelsall (approved):
Approved with a comment.

src/components/pages/familyPage.tsx
PatFawbertMills: Wasn't sure as we will 100% want to present targets, in one form or another, in the future

PR #1361, opened 2026-06-17
feat: gate staging-only resources behind environment check in platfor…

odrakes-cpr:
## Summary

Extends the shared ECS cluster IaC to production. Previously this stack was scoped primarily to staging, as review stack deployments were only needed there. This PR brings the `EcsCluster` resource under management for all stacks, and gates staging-only resources (IAM deployment roles, ESC environments, App Runner ECR access, and Pulumi deployment settings) behind a `stack == 'staging'` check.

## Changes

- **ECS cluster**: `EcsCluster` is now provisioned for all stacks, not just staging
- **Staging-only resources**: IAM OIDC deployment role, App Runner ECR access role, shared ECR repositories, ESC environments, and deployment settings are now conditionally created only when `stack == 'staging'`

## Why

The shared ECS cluster is required in production and should be managed as IaC alongside the existing staging setup. Staging-specific deployment infrastructure (review stacks, ESC environments, Pulumi deployment settings) has no place in a production stack and is now explicitly gated.


annaCPR (approved):
nice

PR #1362, opened 2026-06-17
APP-2182: Implement filters designs

jesskelsall:
# What's changed

Updates `CategorySpecificFilters`:
- Makes use of the new `FiltersContext` for setting which filters are selected.
- Replaced `NestedLabel` with various components:
  - `SearchFilterLevel` for deciding how to represent a layer of peer filters using the below components.
  - `SearchFilterCategory` for top-level categories (will be less category-specific next cycle) as accordions with a big title.
  - `SearchFilterGroup` for accordions per label type.
  - `SearchFilterLookup` for a search input when the list of filters is long (>8) and there are no child filters in any.
  - `SearchFilter` for rendering an individual checkbox.
- `CategorySpecificFilters` uses `Drawer` to house the filters and is toggled with a button which it also renders.
- `CategorySpecificFilters` replaces `SearchFilters` on the `/_search` page.

What this PR doesn't do:
- Other top-level filtering buttons such as topics, geo, and date.
- Filter persistence beyond updating the URL (all checkboxes are unchecked when the page loads).
- Indeterminate state for parent filters where not all child filters are checked.

## Why

Satisfies [APP-2182](https://linear.app/climate-policy-radar/issue/APP-2182/implement-filters-designs)

## AI attribution

Only for analysis and help with figuring out Tailwind styles to apply.

## Screenshots

Once the API is working again.


src/components/atoms/input/Input.tsx
PatFawbertMills, line 1: Question: what's the reason of chopping this component down? We use it in a few different places

src/components/organisms/ConceptPicker.tsx
```diff
@@ -137,13 +137,11 @@ export const ConceptPicker = ({ containerClasses = "", startingSort = "Grouped",
         <div className="flex gap-2 items-center justify-between">
           {showSearch && (
             <Input
-              clearable
               name="Topic quick search"
               placeholder="Quick search"
               value={search}
               onChange={(e) => setSearch(e.currentTarget.value)}
               onClear={() => setSearch("")}
-              size="small"
```
PatFawbertMills, line 146: Follow on from Q: here for instance we need it to appear smaller

src/components/atoms/input/Input.tsx
jesskelsall: The extra size options were massively complicating the component and in all but one use we were opting for the small size, and the component provides another way of overriding its styles if needed.

Admittedly I took out the icon customisation overzealously, especially as the geography filter designs will need it again, so I'll put that back in.

As a general rule of thumb though, I think some of our previous generation of UI components are _too_ customisable and make it too easy to get off-design.

src/components/organisms/ConceptPicker.tsx
jesskelsall: What used to be small is now the universal size.

src/components/atoms/input/Input.tsx
```diff
@@ -1,86 +1,41 @@
...
-      <BaseInput className={classes.input} value={value} {...props} />
-      {clearable && (
-        <button type="button" className={`${classes.button} ${value ? "" : "hidden"}`} onClick={handleClear}>
-          <X height="14" width="14" />
-        </button>
-      )}
-      {!iconOnLeft && icon}
+    <div className={allContainerClasses}>
+      <BaseInput className={allInputClasses} value={value} {...props} />
+      <button type="button" className={buttonClasses} onClick={handleClear}>
```
PatFawbertMills: Suggestion: only display clear if we have provided an `onClear` function?

src/components/molecules/searchFilter/SearchFilter.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export const SearchFilter = ({ ancestorPath, label }: IProps) => {
+  const { toggleFilter } = useContext(FiltersContext);
+
+  const pathLabels = [getFilterPathLabel(label), ...ancestorPath];
+
+  return (
+    <li>
+      <Checkbox onCheckedChange={(value) => toggleFilter(pathLabels, value === true)}>{label.value}</Checkbox>
```
PatFawbertMills, line 21: Check: why the need for `value === true` is `value` not already `boolean` here?

src/components/molecules/searchFilter/SearchFilter.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { TFilterPathLabel, TNestedSearchLabel } from "@/types";
+import { getFilterPathLabel } from "@/utils/filters/filterPaths";
+
+interface IProps {
+  ancestorPath: TFilterPathLabel[];
+  label: TNestedSearchLabel;
+}
+
+export const SearchFilter = ({ ancestorPath, label }: IProps) => {
+  const { toggleFilter } = useContext(FiltersContext);
```
PatFawbertMills, line 15: Check my understanding: are we using Context in order to avoid prop drilling down to this level from the ancestor level?

src/components/molecules/searchFilter/SearchFilter.tsx
PatFawbertMills: Ah I see the "indeterminate" state in another file

src/components/atoms/input/Input.tsx
PatFawbertMills: Sounds good, makes sense.

src/components/molecules/searchFilter/SearchFilter.tsx
jesskelsall: That's right, and next cycle it'll be necessary to access all of the checked filter paths at any depth.

src/components/atoms/input/Input.tsx
jesskelsall: I'm going to keep these two props as separate concerns on the principle that one is for appearance and the other is for functionality.

PatFawbertMills (approved):
👏 great work

PR #1364, opened 2026-06-18
Add missing status to Valibot schema and make slug optional based on status

annaCPR:
# What's changed

- Adds a missing new document status `awaiting_source_file` to the `DocumentAttributesSchema`
- Makes the `deprecated_slug` field `DocumentAttributesSchema` optional if the `status` is `awaiting_source_file` as documents with this status will not have a slug (they are mapped from events because there is no file available yet but we have data we still want to surface on the FE in the documents table) [Applies to Sabin documents only]

## Why

Fixes:
Screenshot 2026-06-18 at 11 17 02

## AI attribution

Syntax lookup for conditional Valibot field validation

### DIFF
--- a/src/schemas/attributesSchema.ts
+++ b/src/schemas/attributesSchema.ts
@@ -27,13 +27,22 @@ export const FamilyAttributesSchema = v.object({
 export type TDataInFamilyAttributes = v.InferOutput<typeof FamilyAttributesSchema>;
 export const validateFamilyAttributes = (attributes: TDataInAttributes): TDataInFamilyAttributes => v.parse(FamilyAttributesSchema, attributes);
 
-export const DocumentAttributesSchema = v.object({
-  action_taken: v.optional(v.string()),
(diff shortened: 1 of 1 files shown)


src/schemas/attributesSchema.ts
```diff
@@ -27,13 +27,22 @@ export const FamilyAttributesSchema = v.object({
...
-});
+export const DocumentAttributesSchema = v.pipe(
+  v.object({
+    action_taken: v.optional(v.string()),
+    deprecated_slug: v.optional(v.string()),
+    md5_sum: v.optional(v.string()),
+    variant: v.optional(v.string()),
+    status: toLiteralUnion(["created", "deleted", "published", "awaiting_source_file"]),
+  }),
+  v.forward(
```
jesskelsall, line 38: Praise: Neat, good use of this!

jesskelsall (approved):
Great change, approved.

PR #1366, opened 2026-06-18
Temporarily decrease page_size for demo

katybaulch:
# What's changed

Vespa is not letting us fetch more than [400 hits](https://docs.vespa.ai/en/reference/api/query.html#native-execution-parameters) all of a sudden because it sucks.

Cap page_size for labels at 400 until this is fixed so we at least return _something_


jesskelsall (approved):
Approved.

PR #1368, opened 2026-06-18
Increase page size

katybaulch:
# What's changed

For the demo we reduced the page size to 400 to make sure we returned results, as over 400 results being requested from vespa was causing an error. We need to reincrease it

## Why

Someone added the query profile file in the wrong place - easily done! (and harder to spot - thanks to Fred!)


jesskelsall (approved):
Approved.

PR #1369, opened 2026-06-18
fix: namespaces were the wrong way round

odrakes-cpr:
Wrong namespace for the service, so we end up trying to create it.


annaCPR (approved):
lgtm

PR #1370, opened 2026-06-18
Allow no items when document status is awaiting_source_file

annaCPR:
# What's changed

- Allow document items to be empty when the `document.attributes.status` is `awaiting_source_file`

## Why

- Sabin docs that are `awaiting_source_file` will not have file urls - this is expected but we still want to include them in the document table.
Screenshot 2026-06-18 at 14 52 44

## AI attribution

Syntax help.


jesskelsall (approved):
Nice one, approved!

PR #1382, opened 2026-06-23
POL-3646 :: climate justice topics

PatFawbertMills:
# What's changed
- Adding new topic group for "Climate justic"

## Why
- New classifer that was not being displayed in the app

## AI attribution
- None


katybaulch (approved):
nice one 😄

PR #1383, opened 2026-06-23
Removing @babel/core @babel/preset-env

PatFawbertMills:
# What's changed
- Removing legacy babel packages from the application

## Why
- Being notified we need to update them triggering me checking if we even use them, and we don't.

## AI attribution
- Checking if we are using them


katybaulch (approved):
Taking your word for it, nice tidy!!

PR #1384, opened 2026-06-23
Fix incorrect CSS rule for applying highlight style

PatFawbertMills:
# What's changed
- Fixing a bug with how we have the css for highlights
- Simplify the use-case so that we can just add the class of "highlights" to places we want to have them

## Why
- Triggering an error in the logs due to invalid CSS

## AI attribution
- Advising on a fix


katybaulch (approved):
much nicer!

PR #1385, opened 2026-06-23
Removing Percy

PatFawbertMills:
# What's changed
- Removed reference to Percy

## Why
- We haven't used and do not have immediate intention to use it
- The package included a security issue due to a dependancy

## AI attribution
- Help determine where we have reference to percy


katybaulch (approved):
bin it bin it bin it

PR #1386, opened 2026-06-24
Update node to v24.x

PatFawbertMills:
# What's changed
- Updates to ensure we are using latest Node v24

## Why
- We were missing an `.nvmrc`
- Best to keep our CI and local aligned
- Node 24v is latest and supported by GHA

## AI attribution
- Answering a few general questions, advising to add


.github/workflows/merge_to_main.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v7
       - uses: actions/setup-node@v6
         with:
-          node-version: 22.13.0
+          node-version: 24.x
```
annaCPR: Question: Any chance we can pull this value dynamically from the .nvmrc file? That way we only have to update it in one place, which means there's less of a risk we miss one.

PatFawbertMills: Good point, we can

annaCPR (approved):
very nice

PR #1387, opened 2026-06-24
APP-2213 :: Principle slide-out header

PatFawbertMills:
# What's change
- Add more control for a custom header in the Drawer component, as well as space for extras in the title
- There are default styles for the `Drawer` that covers us for a typical use-case, but for the principle slide-out there was a need for more controls, such as a link and icon

## Why
- The principle slide-out/drawer content has some particular functionalities which the default drawer header did not support

## AI attribution
- None

## Screenshot
- Custom classNames and jsx:
Screenshot 2026-06-24 at 12 11 41

- Default drawer header:
Screenshot 2026-06-24 at 12 18 37


jesskelsall (approved):
Neat, happy with this approach! Approved.

PR #1389, opened 2026-06-25
Fix document table for litigation event documents

annaCPR:
# What's changed

- Allow for source and cdn urls and content type to be null to allow litigation documents without pdfs to be displayed in the document table
- Refactor to define DISPLAY_ALLOWED_STATUSES in the attributes schema and import it when needed for checks so that the list is defined and updated in one place rather than multiple files where they are easy to miss
- Update the FILING_DATE_EVENT_TYPES to include the new `Filed` type
- Update transformFamilyEvents.ts to include the new `awaiting_source_file` status so those events are not skipped
- Exclude litigation from showing '_We do not have this document in our database..._' text in the document table as it is currently not shown on the CCC app (it was showing up when the new-data-model FF was on because data is now coming from a document instead of event)

## Why

To allow for document data such as `action_taken` or `description` to be shown in the document table even if there is no physical document yet.

## AI attribution
None

## Screenshots

Current production (old data model):
Screenshot 2026-06-25 at 14 43 01

Proposed change (new data model) - note the title now included:
Screenshot 2026-06-25 at 14 42 58


src/bff/transformers/partials/transformDocument.ts
```diff
@@ -37,8 +37,8 @@ export const transformDocument = (document: TDataInDocument, events: TFamilyEven
     language: languages[0] ?? "",
     languages,
     md5_sum: documentAttributes.md5_sum || null,
-    slug: documentAttributes.deprecated_slug,
-    source_url: groupedItems.source[0].url,
+    slug: documentAttributes.deprecated_slug || null,
```
PatFawbertMills, line 40: Question: the returned type for this transformer `TFamilyDocumentPublic` has slug set to `string`, just checking setting this to `null` is not going to break somewhere else in the system?

annaCPR: Good catch. Let me double check

jesskelsall: As extra context, we have `strictNullChecks` set to false, so TypeScript will allow a null value against any typed value, which is easy to not account for further downstream. We'd like to turn this on at some point, but it'll be a non-trivial refactor.

jesskelsall (approved):
Beyond Patrick's comment, no further comments, so approved.

src/bff/transformers/partials/transformDocument.ts
annaCPR: That was a nice catch @PatFawbertMills - I looked up everywhere where `document.slug` was being accessed in the code and found a breaking change in the DocumentDrawer which had a dead link to the document page when the slug was missing. This is link is now only created when the slug is present, other wise we just display the title as text.

Screenshot 2026-06-30 at 11 26 08

annaCPR: I also updated the TFamilyDocumentPublic schema to reflect that slugs can be null.

PR #1393, opened 2026-06-30
APP-2214 :: principle summary area

PatFawbertMills:
# What's changed
- Re-using a lot of logic from the existing family page in order to load data into the new search page principle slide-out

## Why
- Decision to integrate the existing design into the new feature so that we can evolve the two views together in tandem

## AI attribution
- Some help refactoring logic into more modular utilities
- Some help making code more performant

## Screenshots
Screenshot 2026-06-30 at 11 37 46


src/bff/methods/getFamilyData.ts
```diff
@@ -22,27 +22,30 @@ import { isCorpusIdAllowed } from "@/utils/checkCorpusAccess";
...
-  try {
-    // As the families API cannot be queried by slugs, we need to get the slugResponse
-    const { data: slugData } = await apiClient.get<TApiItemResponse<TApiSlugResponse>>(`/families/slugs/${slug}`);
-    slugResponse = slugData.data;
-  } catch (error) {
-    errors.push(new Error("Failed to query family slug", error));
-    return { data: null, errors };
+  // If we are passing an importId in - we can skip calling the slugs endpoint to get it
+  let familyImportId = importId;
+  if (slug.length > 0) {
```
jesskelsall: Suggestion: Falsy slugs can only be zero-length, so `.length` is superfluous.

```suggestion
  if (slug) {
```

src/bff/methods/getFamilyData.ts
```diff
@@ -22,27 +22,30 @@ import { isCorpusIdAllowed } from "@/utils/checkCorpusAccess";
...
 
-  let slugResponse: TApiSlugResponse;
-  try {
-    // As the families API cannot be queried by slugs, we need to get the slugResponse
-    const { data: slugData } = await apiClient.get<TApiItemResponse<TApiSlugResponse>>(`/families/slugs/${slug}`);
-    slugResponse = slugData.data;
-  } catch (error) {
-    errors.push(new Error("Failed to query family slug", error));
-    return { data: null, errors };
+  // If we are passing an importId in - we can skip calling the slugs endpoint to get it
```
jesskelsall, line 32: Praise: great shout!

src/components/_experiment/documentDrawer/DocumentDrawer.tsx
```diff
@@ -19,100 +31,99 @@ type TDocumentDrawerProps = {
   onOpenChange: (open: boolean) => void;
 };
 
+type TDrawerContentProps = {
+  familyData: TFamilyPresentationalData;
+  languages: Record<string, string>;
+};
+
+// The content to display within the Drawer - relies on familyData being loaded
```
jesskelsall: Pedantry: I would say this is clear enough from the props definition above.
Question: Is a no-data state covered one component further up?

src/pages/api/family/[id].ts
```diff
@@ -0,0 +1,23 @@
+import { NextApiRequest, NextApiResponse } from "next";
+
+import { getFamilyData } from "@/bff/methods/getFamilyData";
+import { TTheme } from "@/types";
+import { getFeatureFlags } from "@/utils/featureFlags";
+import { getFeatures } from "@/utils/features";
+import { readConfigFile } from "@/utils/readConfigFile";
+
+// Created as a means to call our new transformation layer from within a client-side component
```
Comment (jesskelsall) [inline src/pages/api/family/[id].ts:9]: Praise: Fair enough. I'd been avoiding needing an API route for this sort of work but once it needs to be dynamically loaded client-side it's inevitable. Thanks for putting it in.

jesskelsall (approved):
Approved with a few comments for consideration.

src/components/_experiment/documentDrawer/DocumentDrawer.tsx
PatFawbertMills: Agreed no need for comment.

No data is not currently handled, good to flag and add something as a catch in case the API call fails / somehow the selected Document is not found, etc.

src/bff/methods/getFamilyData.ts
PatFawbertMills: Good shout

src/pages/api/family/[id].ts
Comment (PatFawbertMills) [reply src/pages/api/family/[id].ts:9]: Your approach was the best approach, no use optimising until we need it

PR #1394, opened 2026-06-30
Combine and persist different types of filters

jesskelsall:
# What's changed

This PR closes the loop on complex hierarchy and API driven label filters so that they can be composed in complex ways and persisted in the URL. This means that reloading the page or sharing a link will result in the same applied filters and the same visual state in the filter panels.

High level changes:
- Added `FILTER_GROUPS` to express the different filter buttons and which filters they should include.
  - Labels are then grouped by `groupSearchLabels`.
- Introduced `FiltersAndSort` component to encapsulate all of the logic and rendering surrounding filters, including modifying the URL.
  - Includes `AppliedFilters` which renders the selected filters.
  - `CategorySpecificFilters` is now rendering a drawer but in my next tickets will become more generic and configurable.
- Paired functions for translating internal filters state to URL parameters:
  - `filterPathsToQueryGroup` when making changes to filters and persisting them in the URL.
  - `queryGroupToFilterPaths` when reading from the URL to determine filters' state.
  - Running one component and then the other is idempotent, so the same unit test cases are used when testing both functions, just with inputs and outputs reversed.
  - The search query encoded into the URL includes `checked: true` for labels actually checked, not just laid out in the AND/OR nested structure to ensure a hierarchically-dependent match (same label under multiple parents).
- Each filter component rendering a checkbox now uses `FiltersContext` + `getFilterStatus` to determine whether to render an empty, checked, or indeterminate state.

This PR does not:
- Make the filter drawers look like the designs.
- Make the entire filter controls area look like the designs.
- Handle date filtering (ideally this should be its own ticket).

## Why

Satisfies:
- [APP-2226 Combine different types of filters](https://linear.app/climate-policy-radar/issue/APP-2226/combine-different-types-of-filters)
- [APP-2227 Persist filters via URL parameters](https://linear.app/climate-policy-radar/issue/APP-2227/persist-filters-via-url-parameters)

## AI attribution

Claude was used extensively to implement the `filterPathsToQueryGroup` and `queryGroupToFilterPaths` functions based on two-directional unit test cases. All solutions have been verified and tweaked by hand.

## Screenshots

_Original diagram of the cycle of filters from URL params:_
IMG_2597

_Filters panel showing a mixed selection:_
Screenshot 2026-06-30 at 13 30 03

_Search page with multiple filters applied:_
Screenshot 2026-06-30 at 13 30 34

_Search page with a category and geography filter applied:_
Screenshot 2026-06-30 at 13 32 15


src/constants/filters.ts
```diff
@@ -0,0 +1,24 @@
...
+  {
+    title: "Geography",
+    Icon: Earth,
+    container: "drawer",
+    rootLabelTypes: ["region"],
+  },
+  {
+    title: "Topic",
+    container: "popover",
+    rootLabelTypes: ["topic"],
```
jamesgorrie: question: is this meant to correspond with the API data?

We use `concept` at the API level.

[API example](https://api.climatepolicyradar.org/search/labels?filters=%7B%22op%22%3A%22and%22%2C%22filters%22%3A%5B%7B%22field%22%3A%22type%22%2C%22op%22%3A%22contains%22%2C%22value%22%3A%22concept%22%7D%5D%7D).

jesskelsall: Great spot, thank you!

jamesgorrie: [There are `topics`](https://api.climatepolicyradar.org/search/labels?filters=%7B%22op%22%3A%22and%22%2C%22filters%22%3A%5B%7B%22field%22%3A%22type%22%2C%22op%22%3A%22contains%22%2C%22value%22%3A%22topic%22%7D%5D%7D) - but frustratingly they are actually something else completely. 

quick explainer: We're trying to remain as semantically close to the source as possible.

- `Concepts` are the source data from the [concept store](https://climatepolicyradar.wikibase.cloud/wiki/Main_Page)
- they're transformed into `Labels` of type `concept` 

avoiding the business / UX case for calling them topics somewhere down the line.

src/components/molecules/appliedFilters/AppliedFilters.tsx
```diff
@@ -1,39 +1,38 @@
...
-      ))}
+    <ul
+      className="col-start-1 -col-end-1 cols-5:col-start-2 cols-5:-col-end-2 flex flex-wrap items-center gap-2 list-none"
+      aria-label="Applied filters"
+    >
+      {labels.map((labelPath) => {
+        const label = labelPath[0];
+
+        return (
+          <li key={label.id} className="flex flex-nowrap gap-1 pl-3 pr-2 py-1 bg-[#0050911a] rounded-full">
```
PatFawbertMills: Pedantry: do we have to use this custom colour?

src/hooks/useLabelSearch.ts
```diff
@@ -35,11 +35,6 @@ export const loadLabels = async (query: string): Promise<TSearchLabel[]> => {
         op: "not_contains",
         value: "hazard",
       },
-      {
-        field: "type",
```
PatFawbertMills, line 39: Question: do we want keyword?

PatFawbertMills (approved):
Great work Jess! Code looks solid and bonus points for including tests

src/hooks/useLabelSearch.ts
jesskelsall: This was removing a duplicate - it's 2 items higher up in the list as well.

src/components/molecules/appliedFilters/AppliedFilters.tsx
jesskelsall: Have to? No! I've gone for `bg-flat` instead.

PR #1395, opened 2026-07-01
tags bot in waf and ingests signal into posthog

abby11a:
# What's changed

Enables AWS WAF Bot Control on the frontend WebACL (production only, count-only mode) and wires the signal through to PostHog so bot traffic can be filtered out of analytics.

- `infra/resources/waf.py`: adds an `enable_bot_control` flag to `FrontendWebAcl`. When it's on, WAF runs the standard `AWSManagedRulesBotControlRuleSet` (common inspection, ML off) purely to label traffic, then a second rule stamps an `x-amzn-waf-is-bot: true` header onto anything it flagged. Nothing here blocks, it just watches and tags.
- `infra/__main__.py`: enables Bot Control on only when env == "production", and adds `x-amzn-waf-is-bot` to the CloudFront origin request policy so the header actually makes it through to the app.
- `middleware.ts`: reads the header and writes a client-readable `is_bot` cookie.
- `src/context/PostHogInit.tsx`: registers `is_bot` as a super-property so every event is tagged.

Flow: WAF (count) → x-amzn-waf-is-bot header → CloudFront forwards → middleware sets is_bot cookie → PostHog tags events.

Bot Control runs in count-only mode - it never blocks, challenges, or slows any request; it only observes and labels. 
Gated to production because it's a paid add-on (~$10/mo per WebACL + ~$1/1M requests analysed)

## Costs
Bot Control subscription: 4 production themes (ccc, cclw, cpr, mcf) × $10/mo = $40/mo
Request analysis: ~$1.00 per 1M requests (common level; on top of the existing ~$0.60/1M WAF fee)
Staging and review stacks are unaffected (Bot Control is production-only).

## Why
Bot traffic inflates our PostHog pageview/event metrics and skews the numbers we rely on. Tagging suspected bot requests lets us filter them out in analytics without blocking any real users. (APP-2230)

## AI attribution
Written with help from Claude. Human reviewed.

## Screenshots
N/A - no UI changes

### DIFF
--- a/infra/__main__.py
+++ b/infra/__main__.py
@@ -384,6 +384,7 @@ if not is_review_stack_or_template:
                     "Accept",
                     "Access-Control-Request-Method",
                     "Access-Control-Request-Headers",
+                    "x-amzn-waf-is-bot",
                 ],
             ),
             cookies=CookieConfig(
@@ -553,6 +554,8 @@ if not is_review_stack_or_template:
 
     frontend_web_acl = FrontendWebAcl(
         name=f"{theme}-{env}-frontend-waf",
+        # Bot Control is a paid add-on; only run it in production.
+        enable_bot_control=env == "production",
         tags={"CUSTOM_APP_THEME": theme, "Environment": env},
     )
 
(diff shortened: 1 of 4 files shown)


PatFawbertMills:
Would love a run through of this at some point.

PatFawbertMills (approved):
Great work. Thanks for the run through.

PR #1406, opened 2026-07-15
Build and apply remaining filter components

jesskelsall:
# What's changed

- `SearchFilter`:
  - Now has an accordion for its children.
  - When unchecked, the accordion only closes if none of its children are checked: `filterHasSelectedChildren` util.
- `SearchFilterParent`
  - Added `showAppliedFilters` prop to render `AppliedFilters` if true. This value originates in the `FILTER_GROUPS` const. 
- `AppliedFilters`:
  - Includes a prop to show the "Clear all" button (not needed inside filter drawer).
- `FiltersContext`:
  - Stores a dictionary of label ID → value pairs so we can get a geography's proper name despite not storing it in the URL. (e.g. `country::USA` has a value of `United States of America`).
- `FILTER_DISPLAY`:
  - A new const object for storing alternative names and descriptions for any label.
  - Initially adds descriptions for top-level categories and regions per the designs.

## Why

Satisfies:
- [APP-2198 Build a child-level filter panel checkbox accordion component](https://linear.app/climate-policy-radar/issue/APP-2198/build-a-child-level-filter-panel-checkbox-accordion-component)
- [APP-2225 Build a top-level filter panel checkbox accordion component](https://linear.app/climate-policy-radar/issue/APP-2197/build-a-top-level-filter-panel-checkbox-accordion-component)

## AI attribution

Used for writing tests mostly, but everything has been hand-checked and tweaked.

## Screenshots

Applied geography filter showing the correct name:
Screenshot 2026-07-15 at 11 46 25

Geography regions with labels showing selected children per group:
Screenshot 2026-07-15 at 11 46 36

Filters panel not showing selected children per group by comparison:
Screenshot 2026-07-15 at 11 47 13

### DIFF
--- a/src/components/atoms/checkbox/Checkbox.tsx
+++ b/src/components/atoms/checkbox/Checkbox.tsx
@@ -1,6 +1,6 @@
 import { Checkbox as BaseCheckbox } from "@base-ui/react";
 import { Check, Minus } from "lucide-react";
-import { ReactNode } from "react";
+import { ReactNode, useId } from "react";
 
 import { TCheckboxState } from "@/types";
 import { joinTailwindClasses } from "@/utils/tailwind";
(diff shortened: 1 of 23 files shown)


src/components/atoms/checkbox/Checkbox.tsx
```diff
@@ -30,15 +36,16 @@ export const Checkbox = ({ children: label, className, ...rootProps }: IProps) =
   };
 
   const IndicatorIcon = rootProps.indeterminate ? Minus : Check;
+  const Wrapper = noClickLabel ? "div" : "label";
```
PatFawbertMills, line 39: Praise: neat

src/components/molecules/searchFilter/SearchFilter.tsx
```diff
@@ -15,16 +18,44 @@ interface IProps {
...
+    if (value === false) {
+      const hasCheckedChildren = filterHasSelectedChildren(checkedLabelPaths, ancestorPath, label);
+      if (!hasCheckedChildren) setIsOpen(false);
+    }
+
+    toggleFilter(pathLabels, value);
+  };
+
+  const onToggle = (event: MouseEvent<HTMLButtonElement>) => {
+    if (event.target instanceof HTMLInputElement) return; // Overrides Base UI's Checkbox click dispatcher
```
PatFawbertMills, line 39: Curious: why do we need to override it?

PatFawbertMills (approved):
Nice work.

Couple of non-blocking comments, and failing test.

src/components/molecules/searchFilter/SearchFilter.tsx
jesskelsall: This is to do with making only the checkbox clickable in regards to checking/unchecking the filter. Base UI has control of this in an odd way, so this line gets around what it's doing (I don't remember the specifics). As a result we can have a different click event when clicking the label.

PR #1413, opened 2026-07-23
Adding feature flag and other knock-on updates

PatFawbertMills:
# What's changed
- Adding `"new-search"` feature flag
- Ensure `Header` and `NavBar` components can read feature flags to adjust the style
- Drive-by updates to `Pagination` component
- Fix `[page].tsx` to ensure it can also read feature flags correctly

## Why
- Feature flags need to be able to be read consistently for this body of work

## AI attribution
- Some investigation into an improvement to [page] however it turned out to not be fruitful, so I reverted to the original method - check for any issues with this approach


src/pages/[page].tsx
```diff
@@ -55,6 +58,7 @@ export default function Page({ page }: InferGetStaticPropsType<typeof getStaticP
 
   // TODO: once dynamic imports are no longer needed, both of these are synchronous
```
Comment (katybaulch) [inline src/pages/[page].tsx:59]: question: i thought we got rid of dynamic imports?

katybaulch (approved):
hell yeah! ⭐ question around dynamic imports for my own clarity and certainly isn't blocking

src/pages/[page].tsx
Comment (PatFawbertMills) [reply src/pages/[page].tsx:59]: We did, but in this situation it is the only way I can think of to get the dynamic pages to work as we need them.

I asked Claude to look at this and it went off on a bit of a wild goose chase, plus I think ultimately the better solution means more of a rewrite of this entire approach. E.g. rather than having this kind of dynamic all purpose page, we should just statically generate the various content pages and completely bypass all of this janky logic.

jesskelsall (approved):
Looks good! You've correctly used the feature flags and set up one that's not yet in app-specific config which feels appropriate. Approved.

PatFawbertMills:
Hold off merging I need to fix something with the non-CPR header components

PR #1419, opened 2026-07-25
Add additional checks for unhandled API errors

PatFawbertMills:
# What's changed
- We were seeing errors logged in the our slack #citical-alerts channel which appeared to be originating in the FE code
- With the help of Claude we have isolated the 3 obvious errors where we were not catching API errors

## Why
- Handle API errors without crashing the app - better for users
- Avoid spam in our critical alerts monitoring channel - better for devs

## AI attribution
- Lots of Claude diagnosis and code suggestion


jesskelsall (approved):
Excellent resilience adds, approved.

PR #1421, opened 2026-07-27
Add the negative margen / padding effect to incorporate the overlay without adding more spacing

PatFawbertMills:
# What's changed
- I removed the padding from the NavSearch in the previous changes, I had not checked the search overlay which was then broken because the padding was reserving the space for the overlay that appears when the user interacts with the search box in the header

## Why
- Fix a bug I introduced


jesskelsall (approved):
Approved

PR #1434, opened 2026-07-29
fixes is_waf_bot posthog delivery and renames it from is_bot

abby11a:
# What's changed
- Renamed the WAF bot cookie and PostHog property from `is_bot` to `is_waf_bot` as it was getting confusing with PostHogs `is bot` property.
- Register the property after persistence is configured (and re-register on consent change) so it survives the persistence switch.
- Also set `is_waf_bot` directly on the `$pageview` capture as a safety net.

## Why
The WAF bot signal (from the `x-amzn-waf-is-bot` header, surfaced via a cookie) was only showing up for internal traffic in PostHog. The cause on the client was that `register()` ran at init, before the consent-driven `set_config({ persistence })`, and switching persistence for non-consented (public) sessions cleared the super property. 

Separately, the name `is_bot` collided conceptually with PostHog's own built-in bot detection, which caused a lot of confusion, so it is renamed to make the source explicit.

## Screenshots
n/a

### DIFF
--- a/middleware.ts
+++ b/middleware.ts
@@ -11,8 +11,8 @@ export default function middleware(request: NextRequest) {
   }
 
   // Surface WAF's bot signal as a client-readable cookie for posthog-js.
-  const isBot = request.headers.get("x-amzn-waf-is-bot") === "true";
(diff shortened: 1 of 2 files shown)


PatFawbertMills (approved):
LGTM

PR #1435, opened 2026-07-29
FUS-121 Update frontend ECS scaling

henry-cpr:
# What's changed

Updates the ECS auto-scaling to use request count per target, with scaling triggered by 500 requests to a service in previous 60s. Minimum tasks updated to 3 and max to 8. 

Task counts taken from config so that prod has the updated counts, and ephemeral environments remain on 1 min. 

## Why
Previously 1 task was minimum. Traffic peaks (free load test by the bots!) demonstrated (1) request threads got blocked waiting on I/O for long periods, resulting in massive latency spikes; (2) the autoscaling policy based on 70% CPU didn't trigger, since it wasn't CPU under contention. 

ECS express only allows one autoscaling target rule, so we can't stack CPU, memory, and request counts. I've chosen request counts as a tunable parameter that can proxy CPU/mem usage as well as I/O contention.  

## AI attribution
Claude's my bestie

## NB
I can't recall the processes around deployment etc. I haven't tested this locally, and can't recall if we need to pulumi up locally for staging or if everything is handled by CI now? Plz advise.

### DIFF
--- a/infra/__main__.py
+++ b/infra/__main__.py
@@ -251,6 +251,8 @@ if not is_review_template:
         name=name_prefix,
         config=ExpressGatewayConfig(
             health_check_path="/",
+            min_task_count=3 if env == "production" else 1,
+            max_task_count=8 if env == "production" else 4,
         ),
         image_identifier=cast(str, image_identifier),
         cluster_arn=shared_resources_stack.get_output("frontend_ecs_cluster_arn"),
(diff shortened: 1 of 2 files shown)


PatFawbertMills (approved):
lgtm

infra/__main__.py
```diff
@@ -251,6 +251,8 @@ def _add_review_tags(
         name=name_prefix,
         config=ExpressGatewayConfig(
             health_check_path="/",
+            min_task_count=3 if env == "production" else 1,
+            max_task_count=8 if env == "production" else 4,
```
jamesgorrie, line 255: nit: you could add this the stack config to avoid the fork

jamesgorrie (approved):

jamesgorrie:
As this should/could reduce noise, I have merged this. Please let me know if that was an issue.

PR #1436, opened 2026-07-29
chore: remove frontend services

odrakes-cpr:
# What's changed

Removing the App runner services.

This is part of the ECS migration, traffic has been fully switched over to ecs.

So these lingering services can be removed. 

## Why

Per ECS Migration.

## AI attribution

None here baby.


annaCPR (approved):
lgtm

PR #1437, opened 2026-07-29
fix: fix pull_request pulumi-preview step

jamesgorrie:
- adds correct permissions for `pulumi-preview`
- removes previewing production stacks as there is branch protection on the `production` environment in GitHub
- adds a small change from [here](https://github.com/climatepolicyradar/navigator-frontend/pull/1435) to validate the changes


jamesgorrie:
Merging due to the blocking nature of this break

PR #1439, opened 2026-07-29
FUS 125 document page with new passage search

PatFawbertMills:
# What's changed
- Adding the new passage search to the document page

## Why
- Part of the new search initiative

## AI attribution
- Creating v1 new passages endpoint
- Logic and recommendations on performance optimisation for the new v1 `DocumentPassageViewer` component
- Used heavily for generating unit tests and storybook shells

## Screenshots
Screenshot 2026-07-30 at 16 44 56

### DIFF
--- a/src/api/passages.ts
+++ b/src/api/passages.ts
@@ -59,6 +59,10 @@ interface SearchPassagesParams {
   query: string;
   documentId: string;
   pageSize?: number;
+  // 1-indexed page number. The API's `page` parameter is currently ignored, and its
+  // `next_page` / `total_pages` fields come back null, so callers page by incrementing
+  // this and comparing the running result count against `total_size`.
+  pageToken?: number;
   signal?: AbortSignal;
 }
 
@@ -75,12 +79,13 @@ function configurePassagesFilters(documentId: string): IPassageFilterGroup {
   };
 }
 
-export async function fetchSearchPassages({ query, documentId, pageSize, signal }: SearchPassagesParams): Promise<SearchPassagesResponse> {
+export async function fetchSearchPassages({ query, documentId, pageSize, pageToken, signal }: SearchPassagesParams): Promise<SearchPassagesResponse> {
   const url = new URL(searchPassagesUrl());
 
   url.searchParams.set("query", query);
   url.searchParams.set("filters", JSON.stringify(configurePassagesFilters(documentId)));
   if (pageSize !== undefined) url.searchParams.set("page_size", String(pageSize));
+  if (pageToken !== undefined) url.searchParams.set("page_token", String(pageToken));
 
   const res = await fetch(url, { signal });
   if (!res.ok) throw new Error(`Passage search API error: ${res.status}`);
--- a/src/api/passages.ts
+++ b/src/api/passages.ts
@@ -4,25 +4,25 @@
 
 import { ISearchPassagesParams, ISearchPassagesResponse, TSearchQueryGroup } from "@/types";
 
-function searchPassagesUrl(): string {
+const searchPassagesUrl = (): string => {
   const origin = (process.env.NEXT_PUBLIC_API_URL || "https://api.climatepolicyradar.org").replace(/\/$/, "");
   return `${origin}/search/passages`;
-}
+};
 
-function configurePassagesFilters(documents: string[]): TSearchQueryGroup {
+const configurePassagesFilters = (documents: string[]): TSearchQueryGroup => {
   return {
     op: "or",
     filters: documents.map((documentId) => ({ field: "document_id", op: "contains", value: documentId })),
   };
-}
+};
 
-export async function fetchSearchPassages({
+export const fetchSearchPassages = async ({
   query,
   documents,
   pageSize,
   pageToken,
   signal,
-}: ISearchPassagesParams): Promise<ISearchPassagesResponse> {
+}: ISearchPassagesParams): Promise<ISearchPassagesResponse> => {
   const url = new URL(searchPassagesUrl());
 
   url.searchParams.set("query", query);
@@ -33,4 +33,4 @@ export async function fetchSearchPassages({
   const res = await fetch(url, { signal });
   if (!res.ok) throw new Error(`Passage search API error: ${res.status}`);
   return res.json() as Promise<ISearchPassagesResponse>;
-}
+};
--- a/src/api/passages.ts
+++ b/src/api/passages.ts
@@ -2,57 +2,7 @@
 // native fetch, `NEXT_PUBLIC_API_URL` with a public fallback, and throwing on a
 // non-2xx so react-query can surface the error.
 
-import { TSearchQueryGroup } from "@/types";
-
-export interface IPassageBoundingBox {
-  coordinates: { x: number; y: number }[];
-}
-
-export interface IPassagePageWithBoundingBoxes {
-  number: number;
-  bounding_boxes: IPassageBoundingBox[];
-}
-
-export interface ISearchPassage {
-  id: string;
-  text_block_id: string;
-  idx: number;
-  text: string;
-  language: string | null;
-  type: string;
-  type_confidence: number;
-  page_number: number;
-  pages: number[];
-  pages_with_bounding_boxes: IPassagePageWithBoundingBoxes[];
-  concepts: unknown[];
-  heading_id: string | null;
-  heading_text: string | null;
-  document_id: string;
-  principal_id: string;
-  tokens: string[];
-}
-
-export interface ISearchPassagesResponse {
-  took_ms: number | null;
-  total_size: number | null;
-  page: number;
-  page_size: number;
-  total_pages: number | null;
-  next_page: string | null;
-  previous_page: string | null;
-  results: ISearchPassage[];
-}
-
-interface ISearchPassagesParams {
-  query: string;
-  documents: string[];
-  pageSize?: number;
-  // 1-indexed page number. The API's `page` parameter is currently ignored, and its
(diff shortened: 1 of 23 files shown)


src/api/passages.ts
```diff
@@ -0,0 +1,86 @@
...
+export interface IPassageBoundingBox {
+  coordinates: { x: number; y: number }[];
+}
+
+export interface IPassagePageWithBoundingBoxes {
+  number: number;
+  bounding_boxes: IPassageBoundingBox[];
+}
+
+export interface ISearchPassage {
```
jesskelsall: Suggestion: It would be preferable to have type exports somewhere in `/types` so they can be included in the folder export and not require other files to import this one. This is particularly useful for a future where we have more clearly defined server and client Next components.

src/api/passages.ts
```diff
@@ -0,0 +1,86 @@
...
+  documents: string[];
+  pageSize?: number;
+  // 1-indexed page number. The API's `page` parameter is currently ignored, and its
+  // `next_page` / `total_pages` fields come back null, so callers page by incrementing
+  // this and comparing the running result count against `total_size`.
+  pageToken?: number;
+  signal?: AbortSignal;
+}
+
+function searchPassagesUrl(): string {
```
jesskelsall: Nitpick: I believe we agreed semi-recently to use variable named functions over `function`.

src/components/molecules/passageBlock/PassageBlock.test.tsx
```diff
@@ -28,15 +28,27 @@ describe("PassageBlock", () => {
...
   });
 
-  it("renders the page number when present", () => {
+  it("renders the page number when present, shifted from the 0-indexed model", () => {
     render(<PassageBlock passage={basePassage} />);
-    expect(screen.getByText("Pg. 16")).toBeInTheDocument();
+    expect(screen.getByText("Pg. 17")).toBeInTheDocument();
+  });
+
+  it("lists every page when a passage spans more than one", () => {
```
jesskelsall, line 36: Praise: good test!

src/components/documents/DocumentPassageViewer.tsx
```diff
@@ -0,0 +1,247 @@
...
+// Memoised so the Adobe viewer is not torn down and rebuilt on every search interaction.
+// `startingPageNumber` is deliberately not passed: it sits in EmbeddedPDF's effect
+// dependencies, so feeding the current page back in would re-register the passages on
+// every passage click. The hook already leaves the reader where it is on a refresh.
+const DocumentPreview = memo(({ document, pageNumber, passages }: TDocumentPreviewProps) => (
+  <EmbeddedPDF document={document} documentPassageMatches={passages} pageNumber={pageNumber} />
+));
+DocumentPreview.displayName = "DocumentPreview";
+
+export const DocumentPassageViewer = ({ document, vespaDocumentData }: TProps) => {
```
jesskelsall, line 71: Question: should we add this to the component library as an organism?

For: it is a component which builds upon other component library components.
Against: it is encapsulated in a component for separation of responsibilities, not for reuse.

src/components/documents/DocumentPassageViewer.tsx
```diff
@@ -0,0 +1,247 @@
...
+  content: passage.text,
+  pages: passage.pages?.map((pageNumber) => ({ page_number: pageNumber })),
+  headingText: passage.heading_text ?? undefined,
+  documentTitle,
+});
+
+// concept_counts keys are `<wikibase id>:<label>` and the same concept can appear across
+// several hits, so counts are summed before ranking. Concepts we have no topic record for
+// are dropped, as we would have no label to show.
+const getTopDocumentConcepts = (vespaDocumentData: TSearchResponse, topics: TTopic[], limit: number): TTopic[] => {
```
jesskelsall: Critique: this function on its own is complex enough that it should be in its own `util` file and have unit tests. Consider the other helper functions in this file too.

src/components/documents/DocumentPassageViewer.tsx
```diff
@@ -0,0 +1,247 @@
...
+  const [searchTerm, setSearchTerm] = useState(query);
+  const [pageNumber, setPageNumber] = useState<number | null>(null);
+
+  // Keep the input in step with the URL when the query changes elsewhere, e.g. the
+  // browser back button or a concept being picked from the empty state. Adjusting during
+  // render rather than in an effect avoids a second render pass with a stale input.
+  const [previousQuery, setPreviousQuery] = useState(query);
+  const [hasNavigatedForQuery, setHasNavigatedForQuery] = useState(false);
+  if (query !== previousQuery) {
+    setPreviousQuery(query);
```
jesskelsall, line 83: Caution: I like this pattern but we do need to be careful that intentionally triggering re-renders like this can mess with the React component lifecycle and is not too far in functionality from the dreaded `useEffect` bypass.

src/components/documents/DocumentPassageViewer.tsx
```diff
@@ -0,0 +1,247 @@
...
+    setSearchTerm(query);
+    setHasNavigatedForQuery(false);
+  }
+
+  const canPreview = !!document.cdn_object && document.cdn_object.toLowerCase().endsWith(".pdf");
```
jesskelsall: Suggestion: Make sure that all of the `use*` calls are before any other statements so that computational overhead is only needed when going to render.

src/components/documents/DocumentPassageViewer.tsx
```diff
@@ -0,0 +1,247 @@
...
+    setHasNavigatedForQuery(true);
+    // `page_number` is 0-indexed in the passage model; the PDF viewer is 1-indexed.
+    setPageNumber(firstResultPage + 1);
+  }
+
+  const topConcepts = useMemo(() => getTopDocumentConcepts(vespaDocumentData, topics, TOP_CONCEPTS_LIMIT), [vespaDocumentData, topics]);
+
+  const handleSubmit = (event: React.SubmitEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    // nuqs removes the param when given null, keeping an empty search out of the URL.
```
jesskelsall: Query: could this not also be achieved by setting `""` as the default param value?

src/hooks/usePDFPreview.test.ts
```diff
@@ -0,0 +1,323 @@
+import type { ISearchPassage } from "@/api/passages";
+import { TFamilyDocumentPublic, TPassage } from "@/types";
+
```
jesskelsall, line 2: Praise: getting tests for the PDF preview in any capacity feels like a big win.

src/hooks/usePDFPreview.ts
```diff
@@ -54,9 +130,13 @@ function generateHighlights(document: TFamilyDocumentPublic, documentPassageMatc
   });
 }
 
+// Adobe SDK does not provide a type
 type TAdobeApis = {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
jesskelsall: Suggestion: Instead of `// eslint-disable-next-line`, you can `/* eslint-disable <rule> */` and then `/* eslint-enable <rule> */` at the end of the relevant area of code.

src/hooks/usePDFPreview.ts
```diff
@@ -143,25 +243,54 @@ export default function usePDFPreview(physicalDocument: TFamilyDocumentPublic, a
...
+    if (key === renderedKey) {
+      return;
+    }
+    renderedKey = key;
+    renderedPage = pageNumber;
+
+    const pageHighlights = currentHighlights.filter((highlight) => highlight.pageNumber === pageNumber);
+    try {
+      // Clear annotations before adding provided ones
+      // console.time("Removing annotations");
```
jesskelsall: Tidyup: remove commented console statements.

src/hooks/usePDFPreview.ts
```diff
@@ -171,29 +300,42 @@ export default function usePDFPreview(physicalDocument: TFamilyDocumentPublic, a
...
     }
 
-    // Finally - register a callback on page change
     // Everytime we change page - add the highlights for that page
     // This will catch passage clicks, as well as navigation within the native pdf reader
     await adobeViewer.registerCallback(
       window.AdobeDC.View.Enum.CallbackType.EVENT_LISTENER,
+      // Adobe SDK does not provide a type
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
       async (event: any) => {
```
jesskelsall: Suggestion: prefer `unknown` over `any` as `any` completely turns off the TS type system.

jesskelsall (approved):
I've left a bunch of comments but haven't spotted anything that makes me want to pump the brakes on this PR, so approved so you can crack on.

src/components/documents/DocumentPassageViewer.tsx
PatFawbertMills: I think what I may do in a later pass is make this component more generic so that it can work within the 3 main contexts, rather than just in a document:
- SERP page for principal and document
- Principal page
- Document page (this work)

src/components/documents/DocumentPassageViewer.tsx
PatFawbertMills: This was based on some conversation with Claude in how best to avoid the 'flicker' effect when the passage results would momentarily display the previous search. It referenced the react docs here: https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes (which has a v similar example to ours).

We could use a key of the query on the component, however I believe that would force the PDF reader to re-render.

PR #1442, opened 2026-07-30
FUS-46: UI create: Common search area components / composition

jesskelsall:
# What's changed

- Renamed `FiltersAndSearch` to `SearchControls`.
- Added a straightforward input for modifying the query parameter.
- Moved the sort dropdown into the component.
- Sort options are now configurable per instance of `SearchControls`.

## Why

Satisfies [FUS-46](https://linear.app/climate-policy-radar/issue/FUS-46/ui-create-common-search-area-components-composition) except for # of matching passages which is now covered by [FUS-154](https://linear.app/climate-policy-radar/issue/FUS-154/apply-latest-designs-to-the-filters-controls)

## AI attribution

Lookup and investigation only.

## Screenshots

Screenshot 2026-07-30 at 15 02 15


src/components/organisms/searchControls/SearchControls.tsx
```diff
@@ -0,0 +1,117 @@
...
+    setFilterParam(filterPathsToQueryGroup(updatedCheckedLabelPaths));
+  };
+
+  const clearFilters = () => {
+    setFilterParam(null);
+  };
+
+  const onSort = (sortValue: string) => {
+    setSortParam(sortValue);
+    if (resetPageOnSort) setCurrentPage("1");
```
PatFawbertMills, line 71: Query: what is the purpose of keeping a log of the page number? I can't see it used anywhere within this component. I would imagine if we did need to reset to page 1 we would do it somewhere within the ancestor component tree?

PatFawbertMills (approved):
Looks good. One non-blocking comment regarding storing page number in the `SearchControls` which I might be missing something but seems superfluous.

PR #1444, opened 2026-08-03
fix: use a non-reserved bot label and qualify it for logging

abby11a:
# What's changed

Renames the WAF bot label from `frontend:waf:bot` to `frontend:bot`, and uses the fully qualified label in the logging filter.

## Why

Two separate bugs, both introduced in #1440, both only reachable in production.

**1. `waf` is a reserved WAF label namespace.** AWS reserves `awswaf`, `aws`, `waf`, `rulegroup`, `webacl`, `regexpatternset`, `ipset` and `managed` in label namespaces and names. It doesn't just ignore the label, it rejects the entire `UpdateWebACL` call, which fails the whole stack update:
```
WAFInvalidParameterException: Error reason: The parameter value isn't supported., field: null, parameter: frontend:waf:bot
```
This blocked every production frontend `pulumi up` since #1440 merged, most recently mcf-production update #157 on an unrelated PR (#1436).

**2. The logging filter needs a fully qualified label.** `LabelNameCondition` requires the full `awswaf:<account>:webacl:<name>:...` form. A bare label name doesn't error, it just matches nothing, and with `default_behavior="DROP"` that
means zero logs. So even with bug 1 fixed, WAF bot logging would have appeared to deploy successfully while capturing nothing.

**Why CI didn't catch either.** `enable_bot_control` and `enable_logging` are both gated on `env == "production"`, so the `FlagBotTraffic` rule carrying the label doesn't exist in staging. Staging went green because it never ran the code path.

Docs: https://docs.aws.amazon.com/waf/latest/developerguide/waf-rule-label-requirements.html

## AI attribution

Root-caused and fixed with Claude.

## Screenshots

n/a, infra only.

## NB

Merging this won't deploy it. This only touches `infra/resources/**`, and `merge_to_main.yml`'s paths filter only watches `infra/__main__.py`, so the `pulumi-up` job will be skipped, exactly as happened with #1435.  Will need to run pulumi up after merged (which I don't have permissions to do)

### DIFF
--- a/infra/resources/waf.py
+++ b/infra/resources/waf.py
@@ -3,6 +3,9 @@ import pulumi_aws as aws
 
 from resources.util import tag_name
 
+# Label applied to bot traffic so the logging filter can select it.
+BOT_LABEL = "frontend:bot"
+
 
 class FrontendWebAcl(pulumi.ComponentResource):
     """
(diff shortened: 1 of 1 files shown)


odrakes-cpr (approved):
lgtm

odrakes-cpr:
lol why is claude getting attribution, i found this first.

PR #1446, opened 2026-08-03
fix: own the rollback alarm so 4XX stops failing deployments

abby11a:
# What's changed

Defines the frontend deployment rollback alarm in Pulumi rather than leaving it to ECS Express, and corrects it to alarm on 5XX only.

- new `infra/resources/rollback_alarm.py`
- called from `infra/__main__.py`, production stacks only
- target groups found by their `Name: <theme>-frontend` tag, so no AWS-generated ids are hardcoded

## Why

Production deploys are being rolled back while the app is healthy. ECS says why:

```json
"statusReason": "Service deployment rolled back because of one or more active alarms.",
"triggeredAlarmNames": ["frontend-production/cpr-frontend-production/RollbackAlarm"], "deploymentCircuitBreaker": { "failureCount": 0 }
```

No task failed to start, and there were zero 5XX. The rollbacks were caused entirely by 404s.

The default alarm ECS Express creates has two faults:

1. It counts 4XX as a failed release. Its IF(4xx < 5, 0, 4xx) gate against a 1% threshold means 5 client errors in a minute computes to 10-75%, because per-minute request counts are small. That condition is met in 2-3% of all minutes on a site with crawler traffic.
2. The denominator is wrong. It divides whole-target-group error counts by RequestCountPerTarget, a per-target average, inflating the rate by the task count. The datapoint that killed the 07-30 deploy read 218%.

Deploys also generate their own 404s: each build changes the Next.js build id while assets are served from the container, so clients mid-session request paths the new task lacks. Background odds of a clean bake window are ~3%, yet 5 of 7
cpr deploys rolled back.

Raising the threshold wouldn't fix it. cpr's true 4XX rate is 1.07%, already over 1%. The alarm has to stop counting 4XX.

Neither the alarm nor the deployment configuration is exposed by create-express-gateway-service, so ExpressGatewayServiceArgs can't reach them. We take the alarm over under the same name instead: PutMetricAlarm is an upsert, so this adopts the AWS-created alarm, and any future rewrite by ECS Express gets reverted on the next pulumi up.

All four production themes run 3 tasks, so all four alarms are affected. cclw is the most exposed at 6.14% 4XX, roughly six times cpr's, and has been passing on timing alone.

## AI attribution
Claude helped a lot both with diagnosing and implementing

## Screenshots
n/a, infra only.

Follow-ups, not in scope:
serve _next/static from a CDN so deploys stop generating 404s at source
enable ALB access logs, currently off, the only way to confirm which URLs 404

### DIFF
--- a/infra/__main__.py
+++ b/infra/__main__.py
@@ -24,6 +24,7 @@ from resources.ecs_express_service import (
     ExpressGatewayServiceComponent,
 )
 from resources.github_actions_role import GitHubActionsRole
+from resources.rollback_alarm import RollbackAlarm, RollbackAlarmConfig
 from resources.util import (
     BehaviourOptions,
     CookieConfig,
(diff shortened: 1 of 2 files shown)


katybaulch (approved):
This looks good Abby! Smart choice to make the rollback alarm explicit. Let's get this in! Do you need a hand deploying?

PR #1453, opened 2026-08-05
Enable static assets for CPR prod

abby11a:
# What's changed
Enables static asset flag


katybaulch (approved):
lgtm, let's give it a whirl!

PR #1458, opened 2026-08-05
removes the manual alarm that was created

abby11a:
# What's changed
Reverts the previous commit https://github.com/climatepolicyradar/navigator-frontend/commit/8c7f788a98ca5b90dec1dc1bbd78f14db88f6152 as the alarm did not override the express alarm and the code is not needed


katybaulch (approved):
good call, nice tidy!

PR #1460, opened 2026-08-05
Sorting on document passage search

PatFawbertMills:
# What's changed
- Adding new `<Sort>` component / migrating old one from experimental into this
- Adding sort capability to the passage search

## Why
- Requirement for Passage Search initiative

## AI attribution
- Initial port of existing component to more generic component
- Generate v1 storybook and unit tests


src/types/passages.ts
```diff
@@ -37,6 +37,10 @@ export interface ISearchPassagesResponse {
   results: ISearchPassage[];
 }
 
+export const SEARCH_PASSAGES_SORT_KEYS = ["relevance desc", "idx asc"] as const;
```
PatFawbertMills, line 40: I considered making this more like search with field and order, but actually we only need to support two use cases which is page number (always ascending) and relevance (always descending). The alternatives make no usable sense.

PR #1461, opened 2026-08-05
Fix broken document links and unhandled-error 500s across family/document/collection pages

katybaulch:
## Root cause - broken links

The API's document attributes schema (attributesSchema.ts:32-47) deliberately makes deprecated_slug optional for documents with status: "awaiting_source_file" -- that status is also in DISPLAY_ALLOWED_STATUSES, i.e. it's allowed to render on the page. So a document can legitimately be displayable and have no slug yet, specifically while it's waiting on its source file. transformDocument.ts:40 passes this through honestly as slug: deprecated_slug || null.

eventTable.tsx's getDocumentLink didn't know about this contract — it decided whether to render a preview link purely from hasMatches/PDF cdn_object, not slug presence. An awaiting_source_file document with a PDF preview but no slug still got rendered as <PageLink href="/documents/null"> (same for the topic-link block further down). Following that link hit SSR for /documents/[id], which correctly returned notFound: true — so it never crashed, but it generated dead links and noisy GET /families/slugs/null -> 404 log lines on every affected family page.

## Root cause - actual 500s (getFamilyData.ts, getCollectionData.ts, getDocumentData.ts)
ApiClient.get() (http-common.ts:57-69) never rejects on HTTP error status -- it catches AxiosError and resolves with error.response instead. Only network-level failures (DNS, timeout) actually reject. That means the try { await apiClient.get(...) } catch blocks around the /families/slugs/{slug} lookups in all three files were dead code for 4xx/5xx responses: the catch never fired, and the code proceeded to use an error-shaped response body as if it were the real payload.

Three concrete crash sites from this:

1. getFamilyData.ts (family.geographies.filter(...), family.collections.map(...)): a families-api 200 response with family present but geographies/collections missing/malformed threw Cannot read properties of undefined uncaught.
2. getCollectionData.ts (getChildDocuments(dataInCollection.documents)): if the preceding data-in fetch failed, dataInCollection stayed null, and this line dereferenced it anyway, uncaught by any try/catch.
3. The slug-lookup blocks in all three files silently proceeded on a non-200/malformed body instead of treating it as a failure.

## Fix

- eventTable.tsx: gate both link-rendering paths on document.slug being present, so a slug-less document falls through to the existing "we don't have this document, contact us" copy instead of a dead link.
- getFamilyData.ts / getCollectionData.ts / getDocumentData.ts: explicit .status !== 200 checks on the /families/slugs/{slug} response, matching the pattern already used for the Vespa/subdivisions fetches elsewhere in these files, so a non-200 or malformed slug response is treated as a failure and returns { data: null, errors } instead of silently continuing.
- getFamilyData.ts: guard family.geographies/family.collections with Array.isArray before calling array methods, falling back to [] and logging an error.
- getCollectionData.ts: only look up child families when dataInCollection is actually populated.

NOTE: the try/catch-doesn't-catch-HTTP-errors pattern exists at every apiClient.get() call site across these three files, not just the ones above -- this PR fixes the specific crash sites identified, not every instance of the pattern.

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -29,10 +29,12 @@ jobs:
         with:
           ref: ${{ github.event.workflow_run.head_sha || github.sha }}
           fetch-depth: 2
+      - id: base-sha
+        run: echo "sha=$(git rev-parse HEAD~1)" >> "$GITHUB_OUTPUT"
       - uses: dorny/paths-filter@v4
         id: filter
         with:
-          base: HEAD~1
+          base: ${{ steps.base-sha.outputs.sha }}
           filters: |
             staging-frontend-cpr:
               - 'infra/Pulumi.cpr-staging.yaml'
--- a/src/bff/methods/getCollectionData.test.ts
+++ b/src/bff/methods/getCollectionData.test.ts
@@ -1,7 +1,14 @@
 import { describe, expect, it } from "vitest";
 
 import { DEFAULT_FEATURES } from "@/constants/features";
-import { collectionByIdHandler, collectionSlugHandler, dataInCollectionHandler, testCollectionSlug } from "@/tests/mocks/api/collectionDataHandlers";
+import {
+  collectionByIdHandler,
+  collectionSlugHandler,
+  dataInChildFamilyHandler,
+  dataInCollectionHandler,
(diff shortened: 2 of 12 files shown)


abby11a (approved):
Nice, looks good!

